PR #5, opened 2020-11-05
Move filter classes into separate file

mberz:
Separate the filter objects form the coefficient calculating generating functions.


tests/test_filter.py
```diff
@@ -2,7 +2,7 @@
 from unittest import mock
 import numpy as np
 import numpy.testing as npt
-import pyfar.dsp.filter as hpfilter
+import pyfar.dsp.classes as hpfilter
```
f-brinkmann, line 5: hpfilter sounds like haiopy - how about pfilt or something like that?

f-brinkmann (approved):
Only one naming suggestion - ready to merge from my point of view...

tests/test_filter.py
mberz: true, it's only in the tests though.

mberz:
> Only one naming suggestion - ready to merge from my point of view...

👍

PR #18, opened 2020-11-14
Enhacement signal

### LINKED ISSUE #9: Signal attribute
- add `comment` attribute
- rename `Signal.shape to Signal.cshape` to be consistent with `Coordinates`

f-brinkmann:

Closes #9 

- added comment attribute
- changes shape attribute to cshape for consistency with Coordinates
- improved __repr__


pyfar/signal.py
```diff
@@ -343,10 +343,18 @@ def __repr__(self):
...
-            "--------------------\n"
-            "{} channels with {} samples @ {} Hz sampling rate".format(
-                self.cshape, self.n_samples, self._sampling_rate))
+            f"{self.domain} domain audio Signal\n"
+            "--------------------------------\n"
+            f"{self.cshape} channels with {self.n_samples} samples @ "
+            f"{self._sampling_rate} Hz sampling rate\n"
+            f"Sinal type: {self.signal_type}\n"
+            f"FFT normalisation: {self.fft_norm}\n"
+            f"Comment: ")
```
mberz, line 352: I think I would rather keep the `__repr__`  compact.
The comment might increase terminal spam too much, probably resulting in users not using it.

mberz: I think I would also be in favor of using as few lines as possible.

pyfar/signal.py
```diff
@@ -301,10 +304,26 @@ def signal_length(self):
...
 
+    @property
+    def comment(self):
+        """Get comment."""
+        return self._comment
+
+    @comment.setter
+    def comment(self, value):
+        """Set comment."""
+        self._comment = value
```
mberz, line 325:
```suggestion
        self._comment = str(value)
```

As in filter?

PR #19, opened 2020-11-14
Filter comment

### LINKED ISSUE #10: Filter comments
Add a `comment` attribute to the Filter class and write comment with all pyfar filter functions. E.g. 'Butterworth highpass of order 2'

f-brinkmann:

Closes #10 

- added comment property to Filter class
- all filters now write comments upon creation

### DIFF
--- a/pyfar/dsp/classes.py
+++ b/pyfar/dsp/classes.py
@@ -195,7 +195,7 @@ class Filter(object):
     @comment.setter
     def comment(self, value):
         """Set comment."""
-        self._comment = value
+        self._comment = str(value)
 
 
 class FilterFIR(Filter):
--- a/pyfar/dsp/classes.py
+++ b/pyfar/dsp/classes.py
@@ -55,7 +55,8 @@ class Filter(object):
             self,
             coefficients=None,
             filter_func=None,
-            state=None):
+            state=None,
+            comment=None):
         """
         Initialize a general Filter object.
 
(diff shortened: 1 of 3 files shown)


pyfar/dsp/classes.py
```diff
@@ -184,6 +187,16 @@ def reset(self):
...
 
+    @property
+    def comment(self):
+        """Get comment."""
+        return self._comment
+
+    @comment.setter
+    def comment(self, value):
+        """Set comment."""
+        self._comment = value
```
mberz, line 198: I think it makes sense here to store the input as a string in the setter in case of numerical input.

tests/test_filter_classes.py
```diff
@@ -38,6 +39,13 @@ def test_filter_init_with_state():
     npt.assert_array_equal(filt._state, state)
 
 
+def test_filter_comment():
+    filt = fo.Filter(coefficients=None, state=None, comment='Bla')
+    assert filt.comment == 'Bla'
+    filt.comment = 'Blub'
+    assert filt.comment == 'Blub'
+
```
mberz, line 47: See previous comment on string comment

pyfar/dsp/classes.py
f-brinkmann: you mean `self._comment = str(value)` ?

mberz: Yeah, in case the user inputs something like `self._comment = str(500)`

mberz (approved):

Issue #16, opened 2020-11-13
Hold not working in plots


mberz:
## Description

Holding plots as standard in matplotlib is not working currently and needs fixing.

f-brinkmann:
closed by pull request #22

PR #22, opened 2020-11-15
Hold plots as in matplotlib default

f-brinkmann:

Adresses the hold fcuntionality requested in issue #16  

- added hold functionality


sikersten (changes requested):
Solves the hold bug for figures with one axis (time, freq...). Unfortunately, replotting into figures with 2 subplots (e.g. freq_phase) fails. Should this issue be solved separately?

f-brinkmann:
> 
> 
> Solves the hold bug for figures with one axis (time, freq...). Unfortunately, replotting into figures with 2 subplots (e.g. freq_phase) fails. Should this issue be solved separately?

hm, it works for me. This should yield a 2x1 subplot with two lines in the lower plot. Does this work for you guys?
~~~
import pyfar.plot as pp
from pyfar import Signal

X = Signal(np.sin(np.linspace(0, 20*np.pi, 512)), 44100)
Y = Signal(np.random.random(512), 44100)

plt.subplot(2, 1, 1)
pp.line.time(X)
plt.subplot(2, 1, 2)
pp.line.time(Y)
pp.line.time(X)
~~~

For me it works with and without passing axes objects to `pp.line.time`

sikersten:
> 
> 
> > Solves the hold bug for figures with one axis (time, freq...). Unfortunately, replotting into figures with 2 subplots (e.g. freq_phase) fails. Should this issue be solved separately?
> 
> hm, it works for me. This should yield a 2x1 subplot with two lines in the lower plot. Does this work for you guys?
> 
> ```
> import pyfar.plot as pp
> from pyfar import Signal
> 
> X = Signal(np.sin(np.linspace(0, 20*np.pi, 512)), 44100)
> Y = Signal(np.random.random(512), 44100)
> 
> plt.subplot(2, 1, 1)
> pp.line.time(X)
> plt.subplot(2, 1, 2)
> pp.line.time(Y)
> pp.line.time(X)
> ```
> 
> For me it works with and without passing axes objects to `pp.line.time`

Sorry, I think my comment was not clear. What I meant is:
```
import pyfar.plot as pp
from pyfar import Signal

X = Signal(np.sin(np.linspace(0, 20*np.pi, 512)), 44100)
Y = Signal(np.random.random(512), 44100)

ax = pp.line.freq_phase(X)
pp.line.freq_phase(Y)
```
This does not work for me, does it work for you?

pyfar/plot/line.py
```diff
@@ -379,21 +380,26 @@ def summary(signal, ax=None, style='light', **kwargs):
...
+    >>> # generate a signal
+    >>> s = Signal([0, 0, 1, 0, 0, 0, 0, 0, 0, 0], 1e3)
+    >>>
+    >>> # two by two plot of time and frequency signals
+    >>> ppl.multi(s, [[ppl.time, ppl.time_dB], [ppl.freq, ppl.group_delay]])
 
     """
-
     if not isinstance(signal, Signal):
         raise TypeError('Input data has to be of type: Signal.')
```
sikersten, line 399: Would a check is plots are pyfar handles make sense here?

sikersten (commented):
A more general remark regarding most of the plot functions: the y-limits are set according to the 
last data plotted into the figure. This is different to the standard matplotlib behavior. Do we want it like that?

f-brinkmann:
> A more general remark regarding most of the plot functions: the y-limits are set according to the
> last data plotted into the figure. This is different to the standard matplotlib behavior. Do we want it like that?

It's working nicely now :)

sikersten (approved):

pyfar/plot/line.py
```diff
@@ -351,23 +351,24 @@ def freq_group_delay(signal, log_prefix=20, log_reference=1, ax=None,
...
+                                     ax, **kwargs)
 
     plt.tight_layout()
     Interaction('LineXLog', ax[0], signal, style, **kwargs)
 
     return ax
 
 
-def summary(signal, ax=None, style='light', **kwargs):
+def multi(signal, plots, ax=None, style='light', **kwargs):
```
mberz, line 362: Multi is a very arbitrary function name. Could we come up with something more explicit?

pyfar/plot/line.py
```diff
@@ -351,23 +351,24 @@ def freq_group_delay(signal, log_prefix=20, log_reference=1, ax=None,
...
     """
-    Plot the time domain, the time domain in dB, the magnitude spectrum,
-    the frequency domain, the phase and group delay.
+    Generate multiple plots of a Signal object.
```
mberz, line 364: The docstring is also not very informational ;)

pyfar/plot/line.py
f-brinkmann: I agree, but do not have the best ideas. I came up with:
- lines
- multi_line

pyfar/plot/line.py
f-brinkmann: How about `Generate custom subplot layouts of a Signal object.`

Still not the best, but the example at the end can also help to clarify.

pyfar/plot/line.py
sikersten: Could this function be extended to non-line plots? What about a name with something like subplots?

PR #27, opened 2020-11-18
update requirements in txt file and setup.py

### LINKED ISSUE #7: Update requirements.txt and maintain setup
## Description
_requirements_dev.txt_ needs to be checked if it is up-to-date with recently added and merged features. A distinction between the current release and development requirements needs to be prepared for the first release. This can be done together with maintaining the setup, README etc.

mberz:

Closes #7 

- List dependencies for installing in the setup file
- Update dependencies for development, including testing, and generating the docs in requirements.txt


f-brinkmann (changes requested):
remove python 3.5. and 3.6 from setup.py.
add python 3.8 to setup.py

setup.py
```diff
@@ -11,11 +11,28 @@
 with open('HISTORY.rst') as history_file:
     history = history_file.read()
 
-requirements = []
+requirements = [
+    'numpy>=1.14.0',
+    'scipy>=1.5.0',
+    'pyfftw',
```
sikersten, line 17: Do we handle pyfftw as a strict requirement or is it optional for improved performance?
from fft.py, line 95:
```try:
    import pyfftw
    pyfftw.config.NUM_THREADS = multiprocessing.cpu_count()
    from pyfftw.interfaces import numpy_fft as fft_lib
except ImportError:
    warnings.warn(
        "Using numpy FFT implementation.\
        Install pyfftw for improved performance.")
    from numpy import fft as fft_lib```

File: requirements_dev.txt [hunk 2]
```diff
@@ -14,5 +14,5 @@ scipy>=1.5.0
 pyfftw
 PyQt5
```
sikersten, line 15: What's the reason PyQt5 is not listed in setup.py?

sikersten (commented):
Looks good to me. Should we update the README.rst also in this pull request or use another one for that?

setup.py
mberz: Good question. I left it in there, since installing with pip is straight forward. There's a pypi package. It is not distributed via conda though, only conda-forge, that's why I left it as an "optional" dependency.

requirements_dev.txt
mberz: There's currently no active code importing it. I'll just remove this.

mberz: removed

f-brinkmann (commented):
Should we add jupyterlab for the notebook that we have?

mberz:
jupyter: not as a requirement :)

f-brinkmann (approved):

mberz:
I've also updated the README file

PR #65, opened 2020-12-19
add channel concatenation function

mberz:

Closes #None

- Add a concatenation function, concatenating Signals along specified channels.


pyfar/signal.py
```diff
@@ -726,3 +726,43 @@ def _divide(a, b):
...
+    data = np.concatenate(
+        (first._data, second._data),
+        axis=axis)
+
+    if first.comment is not None:
+        if second.comment is None:
+            comment = str(first.comment)
+        else:
+            comment = str(first.comment) + '\n' + str(second.comment)
+    else:
```
f-brinkmann, line 760: this case might not be catched:
`elif second.comment is not None: comment = str(second.comment)`

f-brinkmann (changes requested):
see comment on `elif` case

mberz:
May be meaningful to implement this as a class method rather than a function to avoid confusion with merges of other classes.

twennemann:
We had the idea, to add an `ignore_axis` parameter in `utils.broadcast_scape`, so certain caxis will retain for concatenation.
Example:
We want to concatenate `signal1.cshape = (1, 2, 3)`  and `signal2.cshape = (1, 1, 2)`.                                                                                   
So `utils.broadcast_cshape(signal2, ignore_axis=-1)` will return `signal2.cshape = (1, 2, 2)`.
The concatenated signal will subsequently returned with `signal_conc.cshape = (1, 2, 5)`.

mberz:
Closing, replaced by more recent PR #452.

PR #67, opened 2020-12-19
added flatten and reshape method for signal

### LINKED ISSUE #63: creshape and cflatten for Signal class
It would be nice if Signal hat a channel_reshape `creshape()` and channel_flatten function that return copies of the signal with another channel layout. Implementation will be analogous to np.flatten and np.reshape, but the last dimension (n_samples, n_bins) can not be changed.

A use case are the private plot functions (_line.*) and Interaction that require 2D arrays

f-brinkmann:

Closes #63 

- added flatten and reshape methods for signal
- added test for new methods

### DIFF
--- a/tests/test_signal.py
+++ b/tests/test_signal.py
@@ -346,12 +346,16 @@ def test_find_nearest_frequency():
 
 def test_reshape():
 
-    # test reshape
-    x = np.random.rand(6, 256)
-    signal_in = Signal(x, 44100)
+    # test reshape with tuple
+    signal_in = Signal(np.random.rand(6, 256), 44100)
     signal_out = signal_in.reshape((3, 2))
-
     npt.assert_allclose(signal_in._data.reshape(3, 2, -1), signal_out._data)
+    assert id(signal_in) != id(signal_out)
+
+    # test reshape with int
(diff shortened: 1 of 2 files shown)


tests/test_signal.py
```diff
@@ -344,6 +344,44 @@ def test_find_nearest_frequency():
...
+    # test reshape
+    x = np.random.rand(6, 256)
+    signal_in = Signal(x, 44100)
+    signal_out = signal_in.reshape((3, 2))
+
+    npt.assert_allclose(signal_in._data.reshape(3, 2, -1), signal_out._data)
+    assert id(signal_in) != id(signal_out)
+
+    # test assertion for non-tuple input
+    with pytest.raises(ValueError):
```
sikersten, line 358: Additional check for new shape of type int?

f-brinkmann: yes, good idea - I've done that :)

mberz (approved):
+1 on the changes. 
I've implemented a suggestion to also allow for using auto matching of a single dimension as supported by numpy reshaping.

mberz:
Potential conflicts may occur after #69 is merged

sikersten (approved):

PR #68, opened 2020-12-20
add time units for time plots

f-brinkmann:

Closes nothing

- added unit for time plots. Time data can now be plotted on time axis in seconds, mili seconds, micro seconds, and samples
- default behavior detects the best unit (as already implemented in group delay plot)


pyfar/plot/_line.py
```diff
@@ -152,19 +152,34 @@ def _time(signal, dB=False, log_prefix=20, log_reference=1,
...
+    if unit is None:
+        unit = _time_auto_unit(signal.times[..., -1])
+    # set the unit
+    if unit == 's':
+        times = signal.times.T
+    elif unit == 'ms':
+        times = signal.times.T * 1e3
+    elif unit == 'mus':
+        times = signal.times.T * 1e6
+        unit = 'micro s'
```
sikersten, line 165: I would prefer to use the unicode micro sign µ here, or do you expect errors with the label fonts?

tests/test_plot.py
```diff
@@ -10,7 +10,7 @@
 from pyfar import Signal
 
 # flag for creating new baseline plots (required if the plot look changed)
-create_baseline = False
+create_baseline = True
```
sikersten, line 13: Reset to false?

pyfar/plot/_line.py
f-brinkmann: I did not find a simple uniode solution and used Matplotlibs build in mathsupport. To make everything look nice I changed the default fonts and font size in our plot styles (the new fonts are slightly larger than the old one, thus the decreased font size).

tests/test_plot.py
f-brinkmann: Yes, thx for noting - it's time to use enviornment variables for this :)

pyfar/plot/plotstyles/light.mplstyle
```diff
@@ -20,8 +20,9 @@ font.style          : normal
 font.variant        : normal
 font.weight         : normal
 font.stretch        : normal
-font.size           : 10.0
```
mberz, line 23: I would not specify font sizes, they are resolution/display specific

pyfar/plot/plotstyles/dark.mplstyle
```diff
@@ -20,9 +20,9 @@ font.style          : normal
 font.variant        : normal
 font.weight         : normal
 font.stretch        : normal
-font.size           : 10.0
```
mberz, line 23: I would not specify font sizes, they are resolution/display specific

pyfar/plot/_line.py
```diff
@@ -152,19 +152,34 @@ def _time(signal, dB=False, log_prefix=20, log_reference=1,
         ymin = ymax - 90
         ymax = ymax + 10
 
+    # auto detect the time unit
+    if unit is None:
+        unit = _time_auto_unit(signal.times[..., -1])
```
mberz, line 157: Suggestion: I would implement a function which handles the setting of the plot times array, as well as the unit string. The current implementation results in too much redundant code.

pyfar/plot/_line.py
```diff
@@ -288,7 +303,7 @@ def _group_delay(signal, unit=None, xscale='log', ax=None, **kwargs):
     data = dsp.group_delay(signal)
     # auto detect the unit
     if unit is None:
```
mberz, line 305: See comment above

pyfar/plot/_line.py
f-brinkmann: The reason for this is that `_time_auto_unit()` is also used by the `_group_delay()` plot. If we remove redundancy here we would add it there. Leave it like it is?

pyfar/plot/plotstyles/light.mplstyle
f-brinkmann: Agreed and removed all font size specs from both plot styles...

pyfar/plot/plotstyles/dark.mplstyle
f-brinkmann: see above

pyfar/plot/_line.py
f-brinkmann: In this case we would need a second function for handling units in the `_group_delay()` plot, which would add redundancy at a different spot. In the current form, we can separately test the auto uniting, which is reused and have separate code for handling the underlying data, which differs across the plot functions.

Leave it like this, or do I miss what you ment?

PR #69, opened 2020-12-21
Feature/non equidistant audio data

### LINKED ISSUE #42: Class for non-equidistant audio data
## Description

Signal is great, but its data has to be sampled at equidistant times/frequencies, which is not always the case. Should we have a rather simple class that can store time/freq data at user specified times/frequencies?

f-brinkmann:

Closes #42 

- New classes `DataTime` and `DataFrequency` to store non-equidistant or incomplete audio data.
- New classes were realized by redistributing and inheriting the code from Signal
- There is only one change in `Signal` behavior: If passing `comment=None`, `self._comment` is set to `'none'` instead of `None`. This is more consistent, because the setter uses `str(comment)` anyhow.

### Next steps

- make arithmetic and plot functions work for new classes (I will open issues, when this gets merged)


pyfar/signal.py
```diff
@@ -118,60 +47,126 @@ def domain(self, new_domain):
...
-                self.time = fft.irfft(
-                    self._data, self.n_samples, self._sampling_rate,
-                    self._fft_norm)
-            elif new_domain == 'freq':
-                # If the new domain should be freq, we had sampled time data
-                # and need to do a Fourier Transform
-                self.freq = fft.rfft(
-                    self._data, self.n_samples, self._sampling_rate,
-                    self._fft_norm)
+        self._domain = new_domain
```
mberz, line 50: I would remove the domain setter here. The domain setter is only meaningful for Signal. All other classes either do not have a domain, or are limited to a single domain. For example changing the domain on the time domain class would require a cast to a different container class (and a respective ifft)

pyfar/signal.py
```diff
@@ -287,41 +278,162 @@ def fft_norm(self, value):
...
+        indices = np.zeros_like(freqs).astype(np.int)
+        for idx, freq in enumerate(freqs):
+            indices[idx] = np.argmin(np.abs(self.frequencies - freq))
+        return np.squeeze(indices)
 
-        The channel shape gives the shape of the signal data excluding the last
-        dimension, which is `self.n_samples` for time domain signals and
-        `self.n_bins` for frequency domain signals.
+
+class Signal(DataFreq, DataTime):
```
mberz, line 302: Good use of the multi-inheritance feature :)

mberz (commented):
I like the inheritance of both data classes. Good draft, let's continue in that direction.

pyfar/signal.py
```diff
@@ -192,25 +187,46 @@ def find_nearest_time(self, value):
...
-        self.domain = 'time'
-        return self._data
 
-    @time.setter
-    def time(self, value):
-        data = np.atleast_2d(value)
-        self._domain = 'time'
-        self._data = data
-        self._n_samples = data.shape[-1]
+class DataFreq(Audio):
```
mberz, line 191:
```suggestion
class DataFrequency(Audio):
```

pyfar/signal.py
f-brinkmann: Good point - removed the setter from `Audio`. It is now only available for `Signal`

pyfar/signal.py
f-brinkmann: Yes, this is better - done.

tests/test_audio.py
```diff
@@ -0,0 +1,22 @@
+import pytest
+from pyfar.signal import Audio as Audio
```
mberz, line 2:
```suggestion
from pyfar.signal import Audio
```

tests/test_data_frequency.py
```diff
@@ -0,0 +1,95 @@
+import numpy as np
+import numpy.testing as npt
+import pytest
+from pyfar.signal import DataFrequency as DataFrequency
```
mberz, line 4:
```suggestion
from pyfar.signal import DataFrequency
```

tests/test_data_time.py
```diff
@@ -0,0 +1,84 @@
+import numpy as np
+import numpy.testing as npt
+import pytest
+from pyfar.signal import DataTime as DataTime
```
mberz, line 4:
```suggestion
from pyfar.signal import DataTime
```

tests/test_signal.py
```diff
@@ -16,13 +16,16 @@ def test_signal_init_list(impulse_list):
...
+    assert isinstance(signal, Signal)
+
 
 def test_signal_init_default_parameter(impulse_list):
     # using all defaults
     signal = Signal(impulse_list, 44100)
     assert signal.domain == 'time'
     assert signal.fft_norm == 'none'
-    assert signal.comment is None
+    assert signal.comment == 'None'
```
mberz, line 28:
```suggestion
    assert signal.comment == 'none'
```
Let's keep this consistent and better distinguishable from None

mberz (commented):
Looking good so far. 
One thing that comes to mind now that I read it a couple of times though are the names. 
I think I would prefer TimeData and FrequencyData, just out of grammatical/language reasons/correctness

tests/test_data_frequency.py
```diff
@@ -0,0 +1,154 @@
+import numpy as np
```
sikersten, line 1: Rename file to test_frequency_data?

tests/test_data_frequency.py
```diff
@@ -0,0 +1,154 @@
...
+    idx = freq.find_nearest_frequency(.15)
+    assert idx == 1
+
+    # test for a list
+    idx = freq.find_nearest_frequency([.15, .4])
+    npt.assert_allclose(idx, np.asarray([1, 2]))
+
+
+def test_separation_from_data_frequency():
+    """Check if attributes from FrequencyData are really not available."""
```
sikersten, line 126: Separation from time_data

pyfar/signal.py
```diff
@@ -444,6 +193,416 @@ def __setitem__(self, key, value):
...
+        Audio.__init__(self, 'freq', comment, dtype)
+
+        # init data
+        self._data = np.atleast_2d(np.asarray(data, dtype=np.complex))
+
+        # init frequencies
+        self._frequencies = np.atleast_1d(np.asarray(frequencies).flatten())
+        if self._frequencies.size != self.n_bins:
+            raise ValueError(
+                "The number of freqencies must be data.shape[-1]")
```
sikersten, line 339: The monotonous increase check from initializing times in TimeData would also make sense for frequencies

tests/test_data_time.py
```diff
@@ -0,0 +1,144 @@
+import numpy as np
```
sikersten, line 1: Rename file to test_time_data?

pyfar/signal.py
```diff
@@ -444,6 +193,416 @@ def __setitem__(self, key, value):
...
+        """Number of samples."""
+        return self._n_samples
+
+    @property
+    def signal_length(self):
+        """The length of the signal in seconds."""
+        return self.times[-1]
+
+    @property
+    def times(self):
```
sikersten, line 266: Would a setter for `times` make sense, e.g. to shift in time?

tests/test_signal.py
f-brinkmann: Done as suggested :)

pyfar/signal.py
```diff
@@ -444,6 +193,416 @@ def __setitem__(self, key, value):
...
+            self._n_samples = n_samples
+
+    @property
+    def fft_norm(self):
+        """
+        The normalization for the Fourier Transform.
+
+        See pyfar.fft.normalization for more information.
+        """
+        return self._fft_norm
```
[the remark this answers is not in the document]
f-brinkmann: Yes - the values of the frequency data may change with the FFT norm (e.g. a factor of two for all frequencies except 0 Hz and Nyquist). This will be important, if we later have a function that interpolates `FrequencyData` to a complete half-sided spectrum that can be entered into a `Signal`

pyfar/signal.py
```diff
@@ -17,225 +22,35 @@
...
-    @property
-    def sampling_rate(self):
-        """The sampling rate of the signal."""
-        return self._sampling_rate
-
-    @sampling_rate.setter
-    def sampling_rate(self, value):
-        self._sampling_rate = value
-
     @property
```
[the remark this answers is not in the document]
f-brinkmann: Good point. The getter and setter for `signal_type` were moved to `Signal`

pyfar/signal.py
```diff
@@ -17,225 +22,35 @@
...
+        #       in DataFrequqncy and Signal otherwise.
+        self._VALID_TYPES = ["power", "energy"]
+        self._VALID_DOMAINS = ["time", "freq"]
+        self._VALID_FFT_NORMS = [
+            "none", "unitary", "amplitude", "rms", "power", "psd"]
 
-        Audio.__init__(self)
-        self._sampling_rate = sampling_rate
+        # initialize global parameters
+        self.comment = comment
```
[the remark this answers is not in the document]
f-brinkmann: I would prefer not to set `self._data` to make clear that this is a purely abstract class that exists only for inheritance. Is that ok with you?

tests/test_data_frequency.py
f-brinkmann: Yes - done.

tests/test_data_frequency.py
f-brinkmann: yes, of course :)

tests/test_data_time.py
f-brinkmann: Yes - done...

pyfar/signal.py
f-brinkmann: Good point. Added this to the init and made an additional test.

pyfar/signal.py
f-brinkmann: We might want it here (and also a setter for frequencies in `FrequencyData`). The Problem is that we do not want it in `Signal`, which inherits all methods of `TimeData` (https://medium.com/@george.shuklin/why-you-cant-remove-parent-methods-in-python-d4935e6d54a4).

I thus tend not to implement the setter - we should discuss this during the weekly.

pyfar/signal.py
```diff
@@ -444,6 +193,416 @@ def __setitem__(self, key, value):
             raise TypeError(
```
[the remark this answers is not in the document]
f-brinkmann: I added `_assert_matching_meta_data` to `TimeData` and `FrequencyData`. Setitem now works for all three classes.

pyfar/signal.py
```diff
@@ -379,38 +154,12 @@ def comment(self):
...
-        """
-        repr_string = (
-            f"{self.domain} domain {self.signal_type} Signal:\n"
-            f"{self.cshape} channels with {self.n_samples} samples @ "
-            f"{self._sampling_rate} Hz sampling rate and {self.fft_norm} "
-            "FFT normalization\n")
-
-        return repr_string
-
     def __getitem__(self, key):
```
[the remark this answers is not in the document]
f-brinkmann: I added private functions `_return_item` to all classes that return an instance of the correct Class.

pyfar/signal.py
```diff
@@ -444,6 +193,416 @@ def __setitem__(self, key, value):
...
+
+    @freq.setter
+    def freq(self, value):
+        spec = np.atleast_2d(np.atleast_2d(value))
+        new_num_bins = spec.shape[-1]
+        self._data = spec
+        self._domain = 'freq'
+
+        # also set n_samples in case of a Signal object
+        if isinstance(self, Signal):
```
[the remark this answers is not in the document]
f-brinkmann: Yes, it would be possible. I chose this, because it is less repeated code.

pyfar/signal.py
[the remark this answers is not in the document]
sikersten: Nice! I think it makes sense to also define a default of `_assert_matching_meta_data` for _Audio and redefine it in the subclasses

pyfar/signal.py
```diff
@@ -379,57 +143,25 @@ def comment(self):
...
-                    "Index must be int, not {}".format(type(key).__name__))
-        items = Signal(
-            data,
-            sampling_rate=self.sampling_rate,
-            domain=self.domain,
-            fft_norm=self.fft_norm,
-            dtype=self.dtype)
-
-        return items
+        return self._return_item(data)
```
sikersten, line 164: `_return_item` is not a function of _Audio, so would it make sense to implement a default of it for _Audio and keep the redefinitions in the subclasses?

f-brinkmann: Good point. I added empty methods for this and `assert_matching_meta_data` that raise `NotImplementedErrors` if called from the `Audio` base class. There are also tests for checking this :)

pyfar/signal.py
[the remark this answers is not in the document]
f-brinkmann: Good point. I added an empty method that raises a `NotImplementedError` if called from the base class. There is also a test for this :)

pyfar/signal.py
```diff
@@ -16,288 +21,47 @@
...
+        #       in DataFrequqncy and Signal otherwise.
+        self._VALID_TYPES = ["power", "energy"]
+        self._VALID_DOMAINS = ["time", "freq"]
+        self._VALID_FFT_NORMS = [
+            "none", "unitary", "amplitude", "rms", "power", "psd"]
 
-        Audio.__init__(self)
-        self._sampling_rate = sampling_rate
+        # initialize global parameters
+        self.comment = comment
```
mberz, line 42: 👍

mberz (approved):

mberz:
I fixed some typos and spelling issues. Ready to merge.

Issue #30, opened 2020-11-22
Duplicate mocks


f-brinkmann:
there are identical mocks in different test files - should we move all mocks to a single file for a better overview and reusability?

mberz:
Yes, I have thought about that as well. I think this makes sense. This would also make testing the mocks simpler and more compact.

Agent49:
Definitely, don't repeat yourself! Everything else makes maintenance harder.

mberz:
I like the idea of having a separate file for the mocks, that way one can simply check in there for existing mocks and not search through a lot of separate file.

Agent49:
Actually I also think it's a good idea to have all of them in a seperate file, not at least for consistency. We still can  `from mock import my_mock` to only import whats really used in the test. 

Btw. do we whant to gitignore the generated SOFA files or wouldn't be better to hardwire the data?

mberz:
> Actually I also think it's a good idea to have all of them in a seperate file, not at least for consistency. We still can `from mock import my_mock` to only import whats really used in the test.

Yes, that's what I was thinking of. The file should also have it's respective tests of course.

> Btw. do we whant to gitignore the generated SOFA files or wouldn't be better to hardwire the data?

Please stick to the issue at hand. This is a separate issue; I have openend one in #41

sikersten:
An alternative would be to implement the mocks in a `conftest.py`: https://docs.pytest.org/en/stable/fixture.html#conftest-py-sharing-fixture-functions

sikersten:
In our current test implementations we often declare the mocks as pytest fixtures. I think it is unnecessary to do so, as our mocks do not serve the purpose of _setting up services, state, or other operating environments_ (see https://docs.pytest.org/en/stable/fixture.html). So implementing them in conftest.py would be the wrong way. Please comment, if I'm missing something.

f-brinkmann:
It says that fixtures '_may_  setup services, state, or other operating environments'  - so I assume you can also use them for other stuff...

PR #77, opened 2021-01-19
Harmonizing tests

sikersten:

Refers to #30 

- Stubs are implemented in stub_utils.py in order to be able to share them between tests in different files.
- Pytest fixtures are only implemented in conftest.py in order to be able to share them between tests in different files.
- All tests are refactored according to the two facts above.

### Next steps

- Based on these changes testing guidelines will be formulated in a follow-up pull request.
- Probably a revision of all tests will be necessary based on the new guidelines.


tests/stub_utils.py
```diff
@@ -0,0 +1,225 @@
...
+    Returns
+    -------
+    signal
+        stub of pyfar Signal class
+    """
+
+    # Use MagicMock and side_effect to mock __getitem__
+    # See "Mocking a dictionary with MagicMock",
+    # https://het.as.utexas.edu/HET/Software/mock/examples.html
+    def getitem(slice):
```
Agent49, line 32: Sweet!

tests/test_fft.py
```diff
@@ -20,294 +19,178 @@ def test_n_bins_odd():
...
-    samplingrate = 40e3
-    num_periods = np.floor(num_samples / samplingrate * frequency)
-    # round to the nearest frequency resulting in a fully periodic sine signal
-    # in the given time interval
-    frequency = num_periods * samplingrate / num_samples
-    times = np.arange(0, num_samples) / samplingrate
-
-    signal_time = 1 * np.cos(2 * np.pi * frequency * times)
-    signal_spec = fft.rfft(signal_time, num_samples, samplingrate, 'rms')
+def test_fft_orthogonality_sine_even_np(sine, fft_lib_np):
```
Agent49, line 22: `fft_lib_np` never used, should be deleted.
... this occurs in the majority of the following test functions in this file.

tests/test_fft.py
```diff
@@ -20,294 +19,178 @@ def test_n_bins_odd():
...
+        sine.time, sine.n_samples, sine.sampling_rate, sine.fft_norm)
     transformed_signal_time = fft.irfft(
-        signal_spec, num_samples, samplingrate, 'rms')
-
-    npt.assert_allclose(signal_time, transformed_signal_time, rtol=1e-10)
-
+        signal_spec, sine.n_samples, sine.sampling_rate, sine.fft_norm)
+    npt.assert_allclose(
+        transformed_signal_time, sine.time,
+        rtol=1e-10, atol=10*np.finfo(float).eps)
```
Agent49, line 29: missing whitespaces (flake8: E226)
... this occurs in the majority of the following test functions in this file.

tests/test_io.py
```diff
@@ -1,266 +1,84 @@
...
-            signal.time,
-            signal_ref,
-            rtol=1e-10)
+    signal = io.read_sofa(generate_sofa_GeneralFIR)[0]
+    npt.assert_allclose(signal.time, noise_two_by_three_channel.time)
 
 
-def test_read_sofa_GeneralTF(tmpdir):
+def test_read_sofa_GeneralTF(
+        generate_sofa_GeneralTF, noise_two_by_three_channel):
```
Agent49, line 59: `noise_two_by_three_channel` is never used

tests/test_plot.py
```diff
@@ -219,45 +217,49 @@ def test_line_xscale_option(signal_mocks):
             compare_images(baseline, output, tol=10)
 
 
-def test_line_group_delay_unit(signal_mocks):
-    """Test plottin the group delay with different units."""
+def test_time_unit(sine, impulse_group_delay):
```
Agent49, line 220: sine never used.

tests/test_stub_utils.py
```diff
@@ -0,0 +1,329 @@
...
+def test_impulse_func_rms():
+    """Test generation of delta impulse,
+    RMS FFT normalization.
+    """
+    n_samples = 6
+    cshape = (1,)
+    delay = 0
+    fft_norm = 'rms'
+    freq_truth = 1 / n_samples * np.array(
+                        [[1, np.sqrt(2), np.sqrt(2), 1]], dtype=complex)
```
Agent49, line 95: E126... also in a few more lines

tests/test_signal.py
```diff
@@ -236,88 +208,84 @@ def test_setter_fft_norm(sine):
...
     with pytest.raises(ValueError, match='FFT norms do not match'):
         signal[0] = set_signal
 
 
 def test_magic_setitem_wrong_n_samples(sine, impulse):
     """Test the magic function __setitem__."""
-    sr = 44100
-    signal = Signal(sine, sr)
-    set_signal = Signal(sine[..., :-10]*2, sr)
+    signal = Signal(sine.time, sine.sampling_rate)
```
Agent49, line 279: impulse is never used.

Agent49 (changes requested):
This is a big chunk of great code. Really nice work which looks more than good to me. There are only minor changes to make, like deleting unused variables and same flake8 violations. I am not sure how we'd like to handle them. 

Great work. I like how conftests reuses stub_utils and the way its organized a lot. Tests are looking much cleaner and easier to maintain now!

tests/stub_utils.py
sikersten: thx to @f-brinkmann

tests/test_fft.py
sikersten: @mberz Can you explain the reason?

tests/test_stub_utils.py
sikersten: I think I missed something here, how would this be written correctly?

tests/test_stub_utils.py
```diff
@@ -0,0 +1,329 @@
+import pytest
+import numpy as np
+import numpy.testing as npt
+
+from pyfar import Signal
+
+import stub_utils
```
f-brinkmann, line 7:
```suggestion
from . import stub_utils
```

sikersten: I get the warning `Attempted relative import beyond top-level package` thrown for doing it that way, also VSCode does not detects the tests anymore... Any propose on whats better?

mberz (approved):
General re-structuring looks good to me :+1:

tests/test_fft.py
Agent49: Mmmm still missing the whitespace around math operators. In the end this is not important at all. On the other hand why don't we share all the same flake-ignore-list with the CI-server to avoid this?

sikersten: I did not change it here as it appears in `test_fft.py`, where already other flake8 warnings appear in the code I did not touch... I would suggest to open a different issues on the flake8 problems appearing in various files?

Agent49: From https://flake8.pycqa.org/en/2.5.5/config.html:
"If the ignore option is not in the configuration and not in the arguments, only the error codes E123/E133, E226 and E241/E242 are ignored (see the warning and error codes)."

I'll approve that and leave this point to the next meeting.

tests/test_stub_utils.py
Agent49: This was over-indented. The correct indention would have been 4 spaces.

Agent49 (approved):

PR #84, opened 2021-02-03
Enable plotting TimeData and FrequencyData objects

### LINKED ISSUE #82: Provide plot functionality for `TimeData` and `FrequencyData`
## Description

Currently, the plot functions only work for objects of the `Signal` class. Extending them for the new audio classes should be easy. In the best case, extending the check for the instance at the beginning of each plot function might already do the trick.

f-brinkmann:

Closes #82

- Enable Plots for new audio classes
- Testing for new plots
- Added stubs for new audio classes
- Testing for stubs


sikersten (commented):
Nice! The good programming style of the functionalities for the signal class now allows this comparatively easy implementation! 👍 
Only minor questions:
* What about `custom_subplots`?
* Are the two class checks (e.g. in `time` and `_time`) mandatory or redundant?

f-brinkmann:
> 
> 
> Nice! The good programming style of the functionalities for the signal class now allows this comparatively easy implementation! 👍
> Only minor questions:
> 
>     * What about `custom_subplots`?
> 
>     * Are the two class checks (e.g. in `time` and `_time`) mandatory or redundant?

**custom_subplots:** The only plot that works for `TimeData` is time. In this case custom subplots make no sense. `FrequencyData` supports `freq` and `phase` plots but there already is the `freq_phase` that is also supported. The use of custom subplots would be very limited here, so I saved some work and did not support it.

**Checks in `_time`:** Interaction calls the private functions. With the checks, we get reasonable errors if for example using interactions to switch from a time-plot to a frequency-plot of a `TimeData`object (You don't see the errors if you are not actively debugging interaction, but if you are they might be easier to understand).

pyfar/plot/line.py
```diff
@@ -46,8 +46,8 @@ def time(signal, dB=False, log_prefix=20, log_reference=1, unit=None, ax=None,
     matplotlib.pyplot.plot() for possible **kwargs.
     """
 
-    if not isinstance(signal, Signal):
-        raise TypeError('Input data has to be of type: Signal.')
+    if not isinstance(signal, Signal) and not isinstance(signal, TimeData):
+        raise TypeError('Input data has to be of type: Signal or TimeData.')
```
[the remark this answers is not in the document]
f-brinkmann: see reply below :)

tests/test_stub_utils.py
```diff
@@ -67,6 +67,85 @@ def test_signal_stub_frequencies_odd():
...
+    npt.assert_allclose(time_data_stub.times, np.atleast_1d(times))
+
+
+def test_time_data_stub_slice():
+    """Test slicing the TimeData stub."""
+
+    time = [[1, 0, -1], [2, 0, -2]]
+    times = [0, .1, .4]
+
+    frequency_data_stub = stub_utils.time_data_stub(time, times)
```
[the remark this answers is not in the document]
f-brinkmann: THX - corrected!

pyfar/plot/line.py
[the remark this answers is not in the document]
sikersten: Thanks for the explanation! Would it work, if the checks are only done in the private functions?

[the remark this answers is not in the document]
f-brinkmann: Good point! I removed the checks from the public functions. This might have been a left over from the initial dev...

pyfar/plot/line.py
```diff
@@ -104,11 +100,6 @@ def freq(signal, dB=True, log_prefix=20, log_reference=1, xscale='log',
     matplotlib.pyplot.plot() for possible **kwargs.
     """
 
-    if not isinstance(signal, Signal):
-        raise TypeError('Input data has to be of type: Signal.')
-    if xscale not in ['linear', 'log']:
-        raise ValueError(f"xscale must be 'linear' or 'log' but is '{xscale}'")
```
sikersten, line 110: Is this check missing in _freq?

f-brinkmann: The checks were inconsistent before. `freq` was the only functions that checked it, also `xscale` is used elsewhere. We also do not check for the `unit` parameter. Should I add those?

sikersten: I think that would be an improvement! Maybe there is an elegant way, the code snippets in `_freq`, `_phase` and `_group_delay` related to `xscale` look pretty similar

pyfar/plot/_line.py
```diff
@@ -110,6 +110,19 @@ def _set_axlim(ax, setter, low, high, limits):
         setter((min(limits[0], low), max(limits[1], high)))
 
 
+def _lower_frequency_limit(signal):
+    """Return the lower frequency limit for plotting."""
+    if isinstance(signal, Signal):
+        lower_frequency_limit = max(20, signal.frequencies[1])
```
mberz, line 116: Why is there a difference for FrequencyData and Signals here?

mberz:
The pull requests introduces a bug in the toggling of the y-axis:
``` python
import pyfar
import numpy as np
sig = pyfar.Signal(np.random.normal(0, 1, 2**16), 48e3)
pyfar.plot.spectrogram(sig)
```
Then press `shift+y`
results in:
```
  File "/home/marco/miniconda/envs/pyfar/lib/python3.8/site-packages/matplotlib/cbook/__init__.py", line 224, in process
    func(*args, **kwargs)
  File "/home/marco/projects/python/pyfar/pyfar/pyfar/plot/_interaction.py", line 432, in select_action
    self.toggle_plot(EventEmu(self.plot[self.params.plot]))
  File "/home/marco/projects/python/pyfar/pyfar/pyfar/plot/_interaction.py", line 492, in toggle_plot
    ax = _line._spectrogram_cb(
  File "/home/marco/projects/python/pyfar/pyfar/pyfar/plot/_line.py", line 519, in _spectrogram_cb
    _check_time_unit(unit)
  File "/home/marco/projects/python/pyfar/pyfar/pyfar/plot/_line.py", line 153, in _check_time_unit
    raise ValueError(
ValueError: Unit is hann but must be s, ms, mus, samples, or None.
```

mberz:
Issue persists after merging the current develop branch

pyfar/plot/_line.py
f-brinkmann: This is a design choice. The idea is this:
- We do not want to plot 0 Hz, because we most often plot on a log freq scale that does not have 0 Hz
- We want to plot starting at 20 Hz, if we have data down to this
- If we do not have data down to 20 Hz, we plot the lowest available frequency.

For signals, `frequencies[0]=0` in all cases. If we used that for checking, `_lower_frequency_limit()` will always return 20 Hz for signals, although the lowest available non-zero frequency might be higher than that.

f-brinkmann:
> 
> 
> Issue persists after merging the current develop branch

Thanks for finding this - I fixed it :)

PR #86, opened 2021-02-03
Arithmetic operations for all audio classes

### LINKED ISSUE #83: Provide arithmetic operations for `TimeData` and `FrequencyData`?
## Description

Arithmetic operations currently only work for objects of the `Signal` class. Do we want to provide them for the remaining audio classes? It might be fairly easy to do by extending `signal._assert_match_for_arithmetic`.

f-brinkmann:

Closes #83

- Added support for arithmetic operations for `TimeData` and `FrequencyData`
- Added tests for new functionality


sikersten (approved):
It's a little tricky to get through, but looks good to me :)

mberz:
I stumbled upon a bug when I started to review: Plus and minus operators only work on the first sample in the time domain. Check the following:

``` python
import numpy as np
import pyfar

sig = pyfar.Signal(np.sin(np.arange(2**14)/48e3*2*np.pi*1e3), 48e3, fft_norm='rms')
(sig-2).time

```

f-brinkmann:
> 
> 
> I stumbled upon a bug when I started to review: Plus and minus operators only work on the first sample in the time domain. 

Hm, I think this is correct, although not intuitive at first sight. Note that you are doing a frequency domain subtraction with 2, i.e., a spectrum that has 2 everywhere. Here's an example with a shorter signal:

```
# spectrum of pure sine signal
s = pyfar.Signal([0, 0, 0, 6, 0, 0, 0], 48e3, domain='freq', fft_norm='none')
# time signal is [ 1  0 -1  0  1  0 -1  0  1  0 -1  0]
print(s.time)

s2 = s - 2

# spectrum after subtraction is as expected [-2 -2 -2 4 -2 -2 -2]
print(s2.freq)

# time signal after subtraction is [-1 0 -1 0 1 0 -1 0 1 0 -1 0] (only the first bin changed)
print(s2.time)

# this looks funky but is correct. The FFT again gives the correct spectrum [-2 -2 -2 4 -2 -2 -2]
print(np.fft.rfft(s2.time))


# I think you had this kind of subtraction in mind
s3 = s - np.atleast_2d([0, 0, 0, 3, 0, 0, 0])
# which decreases the amplitude of the time signal everywhere [0.5 0 -0.5 0 0.5 0 -0.5 0 0.5  0 -0.5 0]
print(s3.time)
```

mberz:
Ah, I somehow had in mind that `+, -` were supposed to be in the time domain by default, and `*,/, **` in the frequency domain. The frequency domain subtraction indeed seems very counter-intuitive to me. Any other opinions on this?

f-brinkmann:
I have a slight tendency for keeping all defaults 'freq':

I think using `+` and `-` with scalars has little applications. The only that came to my mind is removing a DC-offset which requrires time domain operators to work with scalars. However, it might be confusing and harder to remember different defaults for different operators.

It might be more common to use two signals with  `+` and `-`. In this case the results are identical regardless of the domain.

mberz:
True, I was also thinking in a similar direction. Maybe it might even be worth removing `+,-` operations with scalars for signals, since it's very specific and not very intuitive.

mberz (approved):
Looks good to me, but please add some examples for the arithmetic operations in the time and frequency domains to the docstring of the module. The + & - operators when dealing with scalar values is still something that I expect to be counter-intuitive to most users.

PR #90, opened 2021-02-05
sphericalvoronoi.copy() implemented.

Agent49:
I guess we don't need copy-Methods in all classes. Just those belonging to the public interface, not?

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,7 +4,9 @@ import os.path
 import sofa
 import scipy.io.wavfile as wavfile
 
+from pyfar.spatial.spatial import SphericalVoronoi
 from pyfar.orientations import Orientations
+from pyfar.coordinates import Coordinates
 
 import stub_utils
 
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -555,3 +557,41 @@ def positions():
...
+        np.tile(theta4, 3)))
+    phi = np.tile(np.array(
+            [phi1, phi2, phi3, phi1 + np.pi / 3,
+             phi2 + np.pi / 3, phi3 + np.pi / 3]), 2)
+    rad = np.ones(np.size(theta))
+
+    s = Coordinates(
+        phi, theta, rad,
+        domain='sph', convention='top_colat')
+    return SphericalVoronoi(s)
```
sikersten, line 597: What's the reason for these properties? (Compared to most of the other fixtures in this file, it's a little confusing that this is not a stub, but totally makes sense)

Agent49: This and the other new fixtures are no stubs because they are used to ensure that the encoding and decoding operates properly on Pyfar objects. Also, when class implementation changes in a way that would break the encoding/decoding the respective test should fail.

Agent49: I didn't see we are in the _feature/add-missing-copy_. Actually this belongs rather to feature/object-equality-check.

sikersten: It's called in test_copy, so it makes sense here... Maybe better readable implementation would be more suitable in this case, but the meaning of its properties will probably be more clear in another pull request

Agent49: Yes you are right. It is used by test_copy so nothing to fix here.

Agent49:
Merged pull request #90 from feature/add-missing-copy.

PR #91, opened 2021-02-05
Implemented equality checks with deepdiff

Agent49:
Every public class has now the `__eq__(self)`-method. This enables comparison with `actual == other`. Deepdiff has been chosen because it compares recursivly, shows differences if present and has some nice ignore -features. Moreover it seems to be very stable and is very well supported.

### Test Not Passing
Two tests in test_signal are deliberately not passing. `SignalIterator` seems to have an infinite recursion problem.

### DIFF
--- a/pyfar/spatial/spatial.py
+++ b/pyfar/spatial/spatial.py
@@ -1,3 +1,4 @@
+import deepdiff
 import numpy as np
 from scipy import spatial as spat
 
@@ -31,6 +32,12 @@ class SphericalVoronoi(spat.SphericalVoronoi):
             raise ValueError("All sampling points need to be on the \
                     same radius.")
         super().__init__(points, radius_round, center)
+
+    def __eq__(self, other):
+        """Check for equality of two objects."""
+        return not deepdiff.DeepDiff(
+            self, other, ignore_type_in_groups=[
+                (np.int32, np.intc), (np.int64, np.intc)])
 
 
 def calculate_sph_voronoi_weights(
--- a/pyfar/spatial/spatial.py
+++ b/pyfar/spatial/spatial.py
@@ -36,9 +36,7 @@ class SphericalVoronoi(spat.SphericalVoronoi):
 
     def __eq__(self, other):
         """Check for equality of two objects."""
-        return not deepdiff.DeepDiff(
-            self, other, ignore_type_in_groups=[
-                (np.int32, np.intc), (np.int64, np.intc)])
+        return not deepdiff.DeepDiff(self, other)
 
     def copy(self):
         """Return a deep copy of the Coordinates object."""
--- a/pyfar/spatial/spatial.py
+++ b/pyfar/spatial/spatial.py
@@ -1,5 +1,6 @@
 import deepdiff
 import numpy as np
+from pyfar import utils
 from scipy import spatial as spat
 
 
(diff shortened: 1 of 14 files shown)


tests/test_signal.py
```diff
@@ -371,3 +371,16 @@ def test_flatten():
 
     npt.assert_allclose(signal_in._data.reshape((6, -1)), signal_out._data)
     assert id(signal_in) != id(signal_out)
+
+
+def test___eq___equal(signal):
+    sine = np.sin(2 * np.pi * 440 * np.arange(0, 1, 1 / 44100))
+    actual = Signal(sine, 44100, len(sine), domain='time')
```
sikersten, line 378: This would fail, if something in conftest changes, maybe the approach with signal.copy() would be better?

pyfar/spatial/spatial.py
```diff
@@ -32,6 +34,16 @@ def __init__(self, sampling, round_decimals=12, center=0.0):
                     same radius.")
         super().__init__(points, radius_round, center)
 
+    def __eq__(self, other):
+        """Check for equality of two objects."""
+        return not deepdiff.DeepDiff(
+            self, other, ignore_type_in_groups=[
```
sikersten, line 40: A short comment on the ignores would be nice :)

tests/test_filter_classes.py
```diff
@@ -171,3 +171,15 @@ def test_atleast_3d_first_dim():
...
+
+def test___eq___equal(filter):
+    actual = filter.copy()
+    assert filter == actual
+
+
+def test___eq___notEqual(filter):
+    coeff = np.array([[[1.0, 0.0, 0.0], [1.0, 0.0, 0.0]]])
+    state = np.array([[[1.0, 0.0]]])
+    actual = fo.Filter(coefficients=coeff, state=state, comment='my comment')
```
sikersten, line 184: It's no directly clear, what's the actual difference to `filter` (In an extreme case, `filter` could be changed in conftest.py so that it matches `actual`)

tests/test_signal.py
```diff
@@ -371,3 +371,16 @@ def test_flatten():
...
+    assert signal == actual
+
+import deepdiff
+
+def test___eq___notEqual(signal):
+    sine = np.sin(2 * np.pi * 220 * np.arange(0, 1, 1 / 44100))
+    actual = Signal(sine, 44100, len(sine), domain='time')
```
sikersten, line 385: See comments above, difference between `filter` and `actual` not well defined

sikersten (changes requested):
Implementation looks good! Only minor changes requested on tests using fixtures :)

pyfar/spatial/spatial.py
Agent49: I deleted this, because the `ignore_type_in_groups` will be necessary for comparing with an object read from disk and thus be a requirement of  `feature-save-object`.

tests/conftest.py
```diff
@@ -539,19 +543,131 @@ def generate_sofa_postype_error(
...
+
+@pytest.fixture
+def coordinates():
+    """ Coordinates object.
+    """
+    return Coordinates([0, 1], [2, 3], [4, 5])
+
+
+@pytest.fixture
+def signal():
```
sikersten, line 583: I would suggest rename it to something like `sine_signal` to indicate its properties

PR #100, opened 2021-02-19
update documentation for arithmetic operations

f-brinkmann:
As discussed in the weekly. You can already review. I will merge this after #86

### DIFF
--- a/pyfar/signal.py
+++ b/pyfar/signal.py
@@ -22,6 +22,9 @@ scalar values, e.g.,
 
 > result = pyfar.signal.add((signal_1, 1), 'time')
 
+In this case the scalar `1` is broadcasted, i.e., 1 is added to every sample of
+the Signal (or every bin in case of a frequency domain operation).
+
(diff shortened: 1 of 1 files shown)


pyfar/signal.py
```diff
@@ -6,11 +6,39 @@
...
+
+> result = pyfar.signal.add((signal1, signal2), 'freq')
+
+Time domain operations are applied for `TimeData` objects, i.e.,
+
+> result = time_data_1 + time_data_2
+
+is equivalent to
+
+> result = pyfar.signal.add((time_data_1, time_data_2), 'time')
```
sikersten, line 41: Maybe it would be nice to shortly mention why one needs to be careful using the operators with array likes and scalars for signals, or would this be to much text?

f-brinkmann: Yes, good point. I added a sentence that explains how operations with scalars work.

sikersten (approved):

pyfar/signal.py
mberz: 👍

mberz (approved):

PR #107, opened 2021-02-26
Added tests and fixtures for TimeData and FrequencyData.

Agent49:
This is a **really small pull request** that implements things taht are used by the following feature/save-timedata-frequencydata, but actually not part of the feature itself.


f-brinkmann (approved):
Looks good to me!

PR #108, opened 2021-02-26
Feature/save timedata frequencydata

Agent49:
Please review #107 first.


f-brinkmann (approved):
Thanks for the effort - it all looked good to me!

PR #109, opened 2021-03-02
Enhancement/write wav checks

sikersten:
`write_wav`:
- Changed suffix check to enable compatibility with pathlib objects
- Added if case for flattening warning

### DIFF
--- a/pyfar/io.py
+++ b/pyfar/io.py
@@ -91,10 +91,6 @@ def write_wav(signal, filename, overwrite=True):
     data = data.reshape(-1, data.shape[-1])
     warnings.warn("Signal flattened to {data.shape[0]} channels.")
 
-    # Check for .wav file extension
-    if filename.split('.')[-1] != 'wav':
-        warnings.warn("Extending filename by .wav.")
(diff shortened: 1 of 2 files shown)


pyfar/io.py
```diff
@@ -89,12 +89,8 @@ def write_wav(signal, filename, overwrite=True):
 
     # Reshape to 2D
     data = data.reshape(-1, data.shape[-1])
-    warnings.warn("Signal flattened to {data.shape[0]} channels.")
-
-    # Check for .wav file extension
-    if filename.split('.')[-1] != 'wav':
-        warnings.warn("Extending filename by .wav.")
-        filename += '.wav'
```
f-brinkmann, line 97: is it possible to support pathlib objects and checking the extension? Otherwise users might pass invalid filenames without the wav ending.

sikersten: Found a solution, .wav suffix is now added withouth warning

f-brinkmann: One line for three with more functionality - nice!

PR #115, opened 2021-03-03
Feature/save numpy arrays

Agent49:
Added support for saving numpy arrays in the collection.


f-brinkmann (approved):
Thanks for the effort. The changes look good to me.

From what I see, it would now be relatively easy to allow saving python internals such as strings and lists - is that right?

Agent49:
> 
> 
> Thanks for the effort. The changes look good to me.
> 
> From what I see, it would now be relatively easy to allow saving python internals such as strings and lists - is that right?

Thanks for the review! No, this will be the next feature-branch. I split it up for smaller pull requests/reviews.

PR #123, opened 2021-03-16
Coordinates __getitem__ bugfix

### LINKED ISSUE #122: Coordinates __getitem__, slicing bug
## Description

Slicing of coordinates changes the original object instead of returning a new object.

## What I Did

```
import pyfar
coords = pyfar.Coordinates([0,1], [0,1], [0,1])
coords_sliced = coords[0]
coords.cshape == (2,)
```

sikersten:

Closes #122 

- New Coordinates object is returned by calling self.copy()
- Added test

### DIFF
--- a/pyfar/coordinates.py
+++ b/pyfar/coordinates.py
@@ -1634,7 +1634,7 @@ class Coordinates(object):
     def __getitem__(self, index):
         """Return mutable slice of Coordinates object at index."""
 
-        new = self
+        new = self.copy()
         # slice points
         new._points = new._points[index]
         # slice weights
--- a/pyfar/coordinates.py
+++ b/pyfar/coordinates.py
@@ -1632,7 +1632,7 @@ class Coordinates(object):
         return kdtree
 
     def __getitem__(self, index):
-        """Return mutable slice of Coordinates object at index."""
+        """Return copied slice of Coordinates object at index."""
 
         new = self.copy()
         # slice points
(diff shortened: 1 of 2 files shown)


pyfar/coordinates.py
```diff
@@ -1634,7 +1634,7 @@ def _make_kdtree(self):
     def __getitem__(self, index):
         """Return mutable slice of Coordinates object at index."""
```
f-brinkmann, line 1635:
```suggestion
        """Return copied slice of Coordinates object at index."""
```

Mutable might suggest that the original changes, if you change the slice. This is not the case and it should not be the case. If it was the following could happen:
- create Coordinates object `c` in Cartesian coordinates
- make a slice of c and change it to spherical coordinates
- if the slice was mutable `c` would be a mixture of Cartesian and spherical coordinates...

tests/test_coordinates.py
```diff
@@ -413,6 +413,11 @@ def test_getitem():
     assert isinstance(new, Coordinates)
     assert new.cshape == (1, 5)
 
+    # test if sliced object stays untouched
+    coords = Coordinates([0, 1], [0, 1], [0, 1])
+    new = coords[0]
+    assert coords.cshape == (2,)
```
f-brinkmann, line 419: Additional check if coords do not change

```suggestion
    new = coords[0]
    new.set_cart(2, 2, 2)
    assert coords.cshape == (2,)
    npt.assert_allclose(coords.get_cart()[0], np.array([0, 0, 0]))
```

f-brinkmann (commented):
Thx for fixing that - I made two additional suggestions for improving the old code :)

PR #130, opened 2021-03-26
Run CI against wheel binaries

### LINKED ISSUE #117: Source and binary distribution testing via CI
## Description

As a first step towards a release we should test source and binary distributions.

mberz:

Closes #117 

- Build wheel binaries using CI
- Run all tests against wheels
- (Source distributions are already tested via tox)

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -17,3 +17,4 @@ matplotlib
 python-sofa>=0.2.0
 urllib3
 deepdiff
+tox-wheel
--- a/tox.ini
+++ b/tox.ini
@@ -3,8 +3,8 @@ envlist = py37, py38, flake8, examples, build, wheel
 
 [travis]
 python =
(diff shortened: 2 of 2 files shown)


f-brinkmann (commented):
We once discussed testing flake8 in a separate environment, but I guess this is the better solution with the new Travis credit system.

One thing I did not figure out: If I build the wheel using `python setup.py bdist_wheel` I get a wheel that tells you it works with Python 2 *pyfar-0.1.0-py2.py3-none-any.whl*. Should we worry about this or does #131 fix this anyhow?

mberz:
The wheel will in theory be compatible with python 2 and 3, but running the install will ultimately fail if you're using versions not marked as not supported in the `setup.py`. 
In the near future we'll probably need to move some of the CI configuration from the tox config to the travis config anyways, so I'd not bother with those things now.

f-brinkmann (approved):

PR #139, opened 2021-04-02
Documentation for pyfar.coordinates

f-brinkmann:

I only edited the docstrings. Everything else is untouched.


pyfar/coordinates.py
```diff
@@ -154,39 +151,31 @@ def __init__(self, points_1=None, points_2=None, points_3=None,
     def set_cart(self, points_1, points_2, points_3,
                  convention='right', unit='met'):
```
sikersten, line 153: For cartesian coordinates, the only possible convention parameter is `'right'` and only possible unit is `'met'`, right?

pyfar/coordinates.py
```diff
@@ -460,39 +423,31 @@ def get_sph(self, convention='top_colat', unit='rad', convert=False):
     def set_cyl(self, points_1, points_2, points_3,
                 convention='top', unit='rad'):
```
sikersten, line 429: For cylindrical coordinates, the only possible convention parameter is `'top'`, isn't it?

pyfar/coordinates.py
```diff
@@ -665,27 +611,16 @@ def systems(self, show='all', brief=False):
...
-        brief : boolean
+            all coordinate systems. The default is 'all'.
+        brief : boolean, optional
             Will only list the domains, conventions and units if True. The
             default is False.
 
         Returns
         -------
         Prints to console.
-
```
sikersten, line 678: Put `Notes` here instead `.. note::` of above?

pyfar/coordinates.py
```diff
@@ -1698,26 +1684,27 @@ def cart2sph(x, y, z):
         0 < colatitude < \\pi
 
 
-    Notes
-    -----
-    To ensure proper handling of the azimuth angle, the arctan2 implementation
-    from numpy is used.
+    .. note::
+        To ensure proper handling of the azimuth angle, the arctan2
+        implementation from numpy is used.
```
sikersten, line 1696: `Notes` instead? Not consistent with `cart2cyl`

pyfar/coordinates.py
```diff
@@ -1856,26 +1847,27 @@ def cyl2cart(azimuth, height, radius):
         0 < azimuth < 2 \\pi
 
 
-    Notes
-    -----
-    To ensure proper handling of the azimuth angle, the arctan2 implementation
-    from numpy is used.
+    .. note::
+        To ensure proper handling of the azimuth angle, the arctan2
+        implementation from numpy is used.
```
sikersten, line 1859: Notes instead? Not consistent with `cart2cyl`

sikersten (changes requested):
- It is not clear when `Notes` or `.. note::` is used.. let's keep it consistent within all docstrings
- committed some formatting fixes

pyfar/coordinates.py
```diff
@@ -783,36 +718,35 @@ def get_nearest_k(self, points_1, points_2, points_3, k=1,
...
+            If the ``points`` have the shape ``tuple``, then the ``distance``
+            has the shape ``tuple+(k,)``. When k == 1, the last dimension is
             squeezed. Missing neighbors are indicated with infinite distances.
-        index : ndarray of ints
-            The locations of the neighbors in ``self.data``.
-            If ``points`` have shape ``tuple``, then ``index`` has shape
-            ``tuple+(k,)``. When k == 1, the last dimension of the output is
-            squeezed. Missing neighbors are indicated with ``self.csize``.
+        index : numpy array of ints
+            The locations of the neighbors in ``self.data``. Dimension
```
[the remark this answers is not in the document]
f-brinkmann: good point - they should not. I've changed it to the getter methods

pyfar/coordinates.py
```diff
@@ -824,6 +758,20 @@ def get_nearest_k(self, points_1, points_2, points_3, k=1,
         (self.csize, 3) before they are passed to cKDTree. The index that
```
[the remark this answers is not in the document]
f-brinkmann: I added an example for clarification

pyfar/coordinates.py
```diff
@@ -851,25 +799,25 @@ def get_nearest_cart(self, points_1, points_2, points_3, distance,
...
-            search for everything within ``distance + atol``. The default is
-            1e-15.
+        atol : float, optional
+            a tolerance that is added to `distance`. The default is 1e-15.
 
         Returns
         -------
-        index : ndarray of ints
+        index : numpy array of ints
             The locations of the neighbors in ``self.data``.
```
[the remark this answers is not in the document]
f-brinkmann: added the same example below :)

pyfar/coordinates.py
```diff
@@ -907,25 +870,25 @@ def get_nearest_sph(self, points_1, points_2, points_3, distance,
...
```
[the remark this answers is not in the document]
f-brinkmann: added the same example...

pyfar/coordinates.py
```diff
@@ -70,13 +64,13 @@ def __init__(self, points_1=None, points_2=None, points_3=None,
...
-        Init coordinates container with a set of three points and meta data.
+        Create Coordinates object with or without coordinate points.
 
         The points that enter the Coordinates object are defined by the
-        coordinate system, i.e., the `domain` and `convention`:
+        `domain`, `convention`, and `unit`:
 
         +--------------------+----------+------------+----------+----------+
-        | domain, convention | points_1 | points_2   | points_3 | units    |
+        | domain, convention | points_1 | points_2   | points_3 | unit     |
```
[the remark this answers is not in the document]
f-brinkmann: It would be nice, but I'd have to generate it, which would be to much work for now. I will ask David. He did similar sketches before :)

pyfar/coordinates.py
f-brinkmann: for now yes - but we might want to add something in the future and it is more consistent with the remaining domains :)

pyfar/coordinates.py
f-brinkmann: see above - keeping it is more flexible for future extensions

pyfar/coordinates.py
```diff
@@ -881,6 +829,21 @@ def get_nearest_cart(self, points_1, points_2, points_3, distance,
         (self.csize, 3) before they are passed to cKDTree. The index that
```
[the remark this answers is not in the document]
f-brinkmann: added the same example

pyfar/coordinates.py
f-brinkmann: `Notes` is a section from the numpy docstring guide that I used it things that are nice to know or for things that are specific only for certain parameters or returns . `.. note::` can be used to highlight things that are important for general use. I would suggest to keep both.

pyfar/coordinates.py
f-brinkmann: Yes - did that.

pyfar/coordinates.py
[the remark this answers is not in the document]
sikersten: I thinks that's ok for now, let's put in on to the to do list to implement a reshape of the indexes :)

pyfar/coordinates.py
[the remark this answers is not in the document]
sikersten: Sorry, comment was unclear, this refers to `self.data`, needs to be adjusted also for `get_nearest_cart` and `get_nearest_sph`

pyfar/coordinates.py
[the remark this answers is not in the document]
sikersten: Sorry, comment was unclear, this refers to self.data, needs to be adjusted also for get_nearest_cart and get_nearest_sph

pyfar/coordinates.py
sikersten: Thanks for clarifying, removing `.. note::` at other positions made it more consitent

sikersten (commented):
- Commited some changes to be consistent with @f-brinkmann recent commits
- One more general question: at many positions in explanations and examples the expression `self` is used. I'm not totally happy about that, as `self` only is used internally but not by the user. Though the meaning often is clear (if you've ever implemented a class), any idea for a more elegant way to handle that? How is that done in other projects? On possibility is to leave out the `self` (e.g., `self.get_cart()` -> `get_cart()`, `self.cshape` -> `cshape`).
- The formatting of methods and attributes is not always consistent (e.g. cshape vs. `cshape`). How should we handle this in general?

PR #141, opened 2021-04-04
Documentation for audio classes

f-brinkmann:

I included inherited methods in the documentation and added documentation for magic methods `__iter__`, `__getitem__`, and `__setitem__`.

I touched one thing not related to the documentation
- made SignalIterator a private class

### DIFF
--- a/pyfar/signal.py
+++ b/pyfar/signal.py
@@ -381,7 +381,7 @@ class FrequencyData(_Audio):
         fft_norm : str, optional
             The normalization of the Discrete Fourier Transform (DFT). Can be
             'none', 'unitary', 'amplitude', 'rms', 'power', or 'psd'. See
-            :py:func:`pyfar.dsp.fft.normalization` and _[#] for more
+            :py:func:`pyfar.dsp.fft.normalization` and [#]_ for more
             information. The default is 'none', which is typically used for
             energy signals, such as impulse responses.
         comment : str, optional
@@ -558,7 +558,7 @@ class Signal(FrequencyData, TimeData):
         fft_norm : str, optional
             The normalization of the Discrete Fourier Transform (DFT). Can be
             'none', 'unitary', 'amplitude', 'rms', 'power', or 'psd'. See
-            :py:func:`pyfar.dsp.fft.normalization` and _[#] for more
+            :py:func:`pyfar.dsp.fft.normalization` and [#]_ for more
             information. The default is 'none', which is typically used for
             energy signals, such as impulse responses.
         comment : str
--- a/pyfar/signal.py
+++ b/pyfar/signal.py
@@ -1,5 +1,5 @@
 """
-Provide Container classes and arithmethic operations for audio data.
+Container classes and arithmethic operations for audio data.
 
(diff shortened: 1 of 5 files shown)


pyfar/signal.py
```diff
@@ -376,15 +401,16 @@ def __init__(self, data, frequencies, fft_norm=None, comment=None,
...
-            pyfar.fft.normalization and _[#] for more information. The default
-            is 'none', which is typically used for energy signals, such as
-            impulse responses.
-        comment : str
+        fft_norm : str, optional
+            The normalization of the Discrete Fourier Transform (DFT). Can be
+            'none', 'unitary', 'amplitude', 'rms', 'power', or 'psd'. See
+            :py:func:`pyfar.dsp.fft.normalization` and [#]_ for more
+            information. The default is 'none', which is typically used for
+            energy signals, such as impulse responses.
```
sikersten, line 409: Comment on why `fft_norm` it is useful for FrequencyData would be nice

pyfar/signal.py
```diff
@@ -769,13 +802,26 @@ def __len__(self):
         return self.n_samples
 
     def __iter__(self):
-        """Iterator for signals. The actual iteration is handled through
-        numpy's array iteration.
+        """Iterator for audio objects.
+
+        Iterate across the first dimension of an audio object. The actual
+        iteration is handled through numpy's array iteration.
```
sikersten, line 807: Only used by Signal, adjust docstring?

pyfar/signal.py
```diff
@@ -124,9 +129,9 @@ def reshape(self, newshape):
         reshaped : Signal
```
sikersten: Could also be FrequencyData or TimeData, same holds for flatten()

pyfar/signal.py
```diff
@@ -800,135 +846,141 @@ def __next__(self):
...
-        Data to be added. Can contain Signal objects, numbers, and, array
-        likes.
-    domain : 'time', 'freq'
+        Data to be added. Can contain pyfar audio objects, array likes, and
+        scalars. Pyfar audio objects can not be mixed, e.g., TimeData and
+        FrequencyData objects do not work together.
+    domain : 'time', 'freq', optional
         Flag to indicate if the operation should be performed in the time or
         frequency domain. If working in the frequency domain, the FFT
         normalization is removed before the operation (See
```
sikersten, line 861: removed = set to 'none'? Is this information important for the user? same holds for other arithmetic operations

sikersten (commented):
- Commited some minor changes, mainly replaced "signal" with "audio object" at some positions
- Formatting of references is not consistent with e.g. pyfar.signals. Let's agree on a format for all references?
- I don't quite get when to use ``attribute`` with `` and `attribute` with `, what's the rule?
- self is used in docstrings, see comments in review on #139

pyfar/signal.py
f-brinkmann: No, I also it is better this way

pyfar/signal.py
f-brinkmann: yes, did that.

PR #142, opened 2021-04-08
flake8 rules added to CONTRIBUTING.rst, closes #93

### LINKED ISSUE #93: Flake8 Guideline
## How do we want to handle flake8 checks?

In a review or generally when I pulled from origin I often run into the situation of eliminating flake8 warnings. I propose that we all use the same flake8-configuration as the build server. 

## Recommendation for a flake8 Configuration
The default ignores `E123/E133, E226 ` and `E241/E242`. Maybe we should keep this configuration at our workstations on a system level (~/flake8). However, we could have a stricter configuration in our IDEs ("python.linting.flake8Args").

We could, for example, ignore E226 (missing whitespace around arithmetic operator) on the build server and when hitting flake8 on command line to enable things like `pi/2`. However, by enabling the squiggly line in the editor we avoid a violation of E226 that was not intended, like `sine = 0.5 *sin(2*pi *f*t)`.

Agent49:
Micro-change to CONTRIBUTING.rst, see #93

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -83,7 +83,7 @@ Ready to contribute? Here's how to set up `pyfar` for local development.
     $ python setup.py test or py.test
     $ tox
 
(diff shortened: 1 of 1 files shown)


f-brinkmann (commented):
Thanks. This looks fine to me. One question not related to this pull. Should we make line 83 simpler and write only `$ pytest`?

PR #143, opened 2021-04-09
Read the docs/io

sikersten:

Touched one thing not related to the documentation: replaced to .far suffix check in read() and write with `pathlib.Path(filename).with_suffix('.far')`, this also allows pathlib.Path objects as input.

### DIFF
--- a/pyfar/io.py
+++ b/pyfar/io.py
@@ -1,3 +1,14 @@
+"""
+Read and write objects to disk, read and write WAV files, read SOFA files.
+
+The methods ``read(...)`` and ``write(...)`` allow to save or load several
+pyfar objects and other variables. So, e.g., workspaces in notebooks can be
+stored. Signal objects can be imported and exported as WAV files using
+``read_wav()`` and ``write_wav()``. ``read_sofa()`` provides functionality
+to read the data stored in a SOFA file. The impulse response is
+imported as Signal object, the source and receiver positions are imported as
+Coordinates objects.
+"""
 import scipy.io.wavfile as wavfile
 import os.path
 import pathlib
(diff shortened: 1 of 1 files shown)


pyfar/io.py
```diff
@@ -107,32 +113,29 @@ def write_wav(signal, filename, overwrite=True):
...
+    filename : string, Path
         Input SOFA file (cf. [#]_, [#]_).
 
     Returns
     -------
-    signal : signal instance
-        An audio signal object from the pyfar Signal class
-        containing the IR data from the SOFA file with cshape being
+    signal : Signal
+        Signal object containing the IR data. cshape is
```
f-brinkmann, line 126: Maybe more precise:
```suggestion
        Signal object containing the data stored in SOFA_Object.Data.IR. cshape is
```

pyfar/io.py
```diff
@@ -107,32 +113,29 @@ def write_wav(signal, filename, overwrite=True):
...
-        An object from the pyfar Coordinates class containing
-        the source coordinates from the SOFA file
-        with matching domain, convention and unit.
-    receiver_coordinates: coordinates instance
-        An object from the pyfar Coordinates class containing
-        the receiver coordinates from the SOFA file
-        with matching domain, convention and unit.
+    source_coordinates : Coordinates
+        Coordinates object containing the source coordinates.
+        The domain, convention and unit are automatically matched.
```
f-brinkmann, line 130: More explicit:
```suggestion
        Coordinates object containing the data stored in SOFA_object.SourcePosition.
        The domain, convention and unit are automatically matched.
```

pyfar/io.py
```diff
@@ -107,32 +113,29 @@ def write_wav(signal, filename, overwrite=True):
...
+    receiver_coordinates : Coordinates
+        Coordinates object containing the receiver coordinates.
+        The domain, convention and unit are automatically matched.
```
f-brinkmann, line 133: More explicit:
```suggestion
        Coordinates object containing the data stored in SOFA_object.RecevierPosition.
        The domain, convention and unit are automatically matched.
```

pyfar/io.py
```diff
@@ -107,32 +113,29 @@ def write_wav(signal, filename, overwrite=True):
...
 
     Notes
     -----
-    * This function is based on the python-sofa package.
-    * Only SOFA files of DataType 'FIR' are supported.
+    * This function is based on the python-sofa.
```
f-brinkmann, line 137: Add a reference to https://github.com/spatialaudio/python-sofa ?

f-brinkmann (changes requested):
Thanks for the effort. I made a small commit and only have minor suggestions on top of that. One general thing: We should discuss the tables in `write/read_wav` it is not correct that float32 is bound to +/-1. wav files with values >1 can be written and read.

PR #151, opened 2021-04-09
Move orientations, signal, and coordinates classes into separate folder

mberz:
Move the modules
- `audio` (used to be`signal`)
- `coordinates`
- `orientations`
- `filter`

into the classes module.

Top level imports for
- FilterFIR
- FilterIIR
- FilterSOS

The modules implementing respective class functionality can still be imported as `pyfar.classes.<module>`


mberz:
Travis seems to be having some issues. Test pass locally on my machine.

f-brinkmann (approved):
Thanks for the effort this looks all good for me. Two general question came to my mind:
- Should we import the arithmetic functions from ``classes.signal`` (e.g. ``add``) in a way that they can be used as ``pyfar.add``?
- should we rename ``classes.signal`` to ``classes.audio``? Maybe not today, but in the future...

mberz:
1. I've been thinking of moving the import for arithmetic operations, but decided against it to avoid confusion with arithmetic operations for other classes. They do not exist as of now, but I think they would be useful (at least for coordinates).
2. I would either rename now, or keep as is. Renaming in the future potentially changes a lot of user code down the line.

PR #155, opened 2021-04-10
Use Travis to deploy to PyPI

mberz:

Use Travis-CI to deploy to PyPI on
- Tags with version numbers
- After all tests and build of source and binary dists passed


f-brinkmann (approved):
We could think about re-adding some comments to make it easier for others to understand what is going on.

Issue #105, opened 2021-02-26
Plot defaults for 3D axes


f-brinkmann:
Add sensible default Values for 3D axes to pyfar plotstyles

PR #169, opened 2021-04-15
Enhanceme spatial plots

f-brinkmann:

Fixing #105 and more: The plots `scatter` and `quiver` used in `Coordinates.show()` and `Orientations.show()` did not use the pyfar plotstyle.

- added default marker `.` to pyfar plotstyles
- use pyfar default plotstyle for spatial plots (scatter and quiver)
- use pyfar colors for plotting (scatter and quiver)
- added default marker size depending on the number (scatter)


mberz:
I'm getting very dark backgrounds and grid lines compared to the mpl default plot behavior. Is this intended?
scatter

f-brinkmann:
That annoyed me as well, but I did not find a quick solution. I don't find a property in the mplstyle that sets the background color and the 3D axes seem to ignore the grid color of the mplstyle. We could have a private function that does the settings, e.g.

```python
ax.w_xaxis.set_pane_color((1, 1, 1, 1))
ax.w_yaxis.set_pane_color((1, 1, 1, 1))
ax.w_zaxis.set_pane_color((1, 1, 1, 1))
ax.xaxis._axinfo["grid"].update({'color':(0, 0, 0, .3)})
ax.yaxis._axinfo["grid"].update({'color':(0, 0, 0, .3)})
ax.zaxis._axinfo["grid"].update({'color':(0, 0, 0, .3)})
```

We could also use the theme colors and won`t need the hard coded colors of the example.

sikersten (changes requested):
I get an error for calling `plot.scatter` with `projection='ortho'` as well as an deprecation warning: 
```
spatial.py:123: MatplotlibDeprecationWarning: Calling gca() with keyword arguments was deprecated in Matplotlib 3.4. Starting two minor releases later, gca() will take no keyword arguments. The gca() function should only be used to get the current axes, or if no axes exist, create new axes with default keyword arguments. To create a new axes with non-default arguments, use plt.axes() or plt.subplot().
  ax = plt.gca(projection=projection)
```

Am I doing something wrong? If not, should we fix this as part of this pull request?

f-brinkmann:
Hm, I tought we fixed that a while ago. Can you check if the new code works for you? I need to find some time to update my environment later :)

PR #171, opened 2021-04-17
Revise testing for coordinates

f-brinkmann:

- Unified use of numpy testing
- Removed tests that were not needed
- Added parameterization for one case
- Minor bugdixes:
    - `Coordinates.get_cyl()` did not convert to degrees
    - ValueError was not raised in `Coordinates.get_slice()` if a non existing coordinate was passed
    - `Coordinates.__getitem__` did not maintain the correct shape of the coordinate points if only one coordinate was left

### DIFF
--- a/pyfar/classes/coordinates.py
+++ b/pyfar/classes/coordinates.py
@@ -544,7 +544,7 @@ class Coordinates():
                 f"Conversion for {convention} is not implemented.")
 
         # convert to degrees
-        if self._system['unit'] == 'deg':
+        if new_system['unit'] == 'deg':
             pts_1 = pts_1 / np.pi * 180
 
         # return points and convert internal state if desired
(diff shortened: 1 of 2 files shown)


pyfar/classes/coordinates.py
```diff
@@ -273,19 +273,25 @@ def get_cart(self, convention='right', unit='met', convert=False):
                 y, z, x = sph2cart(pts[..., 0], pts[..., 1], pts[..., 2])
 
             else:
+                # Can not be tested. Will only be raised if a coordinate system
```
sikersten, line 276: If it can not be tested, what would be a situation, when this error is raised? Maybe leave it out?

f-brinkmann: This would be raised if we add a new coordinate system later, but forget to implement the conversion for it. I would thus leave it in.

PR #178, opened 2021-05-07
Windowing function

sikersten:

- added windowing function to pyfar.dsp
- function is based seminar work by @udowa on https://github.com/pyfar-seminar/pyfar/tree/feature/windowing (adjusted naming of variables, minor changes in structure, added tests)
- windows a taken from scipy.signal.windows
- I tried to define descriptive parameter names, but I'm open for better solutions


pyfar/dsp/dsp.py
```diff
@@ -227,6 +227,169 @@ def spectrogram(signal, dB=True, log_prefix=20, log_reference=1,
...
+        if length != sorted(length):
+            raise ValueError("Values in length need to be in ascending order.")
+        # Convert to samples
+        if unit == 's':
+            length = [round(li*signal.sampling_rate) for li in length]
+        elif unit == 'ms':
+            length = [round(li*signal.sampling_rate/1e3) for li in length]
+        elif unit != 'samples':
+            raise ValueError(f"unit is {unit} but has to be"
+                             f" 'samples', 's' or 'ms'.")
```
f-brinkmann, line 333: This would make it more flexible and we also do this in `pyfar.signals`:
```suggestion
        elif unit == 'samples':
            length = [int(li) for li in length]
        else:
            raise ValueError(f"unit is {unit} but has to be"
                             f" 'samples', 's' or 'ms'.")
```

f-brinkmann (commented):
Looks good to me. I only had minor points and a few little things that are up for discussion.

pyfar/dsp/dsp.py
```diff
@@ -227,6 +227,169 @@ def spectrogram(signal, dB=True, log_prefix=20, log_reference=1,
...
+    For the left sight of a symmetric window and for ``shape='left'``,
+    the indexes of the samples given in `length` denote the first sample of
+    the windowing which is non-zero and the first being one. For the right
+    side of a symmetric window and for ``shape='right'``, the samples denote
+    the last sample being one and the last being non-zero.
+
+    This function calls `~scipy.signal.windows.get_window` to create the
+    window.
+    Window types:
+    - ``scipy.signal.windows.boxcar``
```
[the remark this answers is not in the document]
sikersten: This is how I copied it from scipy.signal.windows.get_window, but we can simplify it to be more readable

pyfar/dsp/dsp.py
sikersten: thx

pyfar/dsp/dsp.py
```diff
@@ -227,6 +227,169 @@ def spectrogram(signal, dB=True, log_prefix=20, log_reference=1,
...
+
+    # Apply window
+    signal_win = signal.copy()
+    if truncate:
+        signal_win.time = signal_win[..., win_start:win_stop+1].time*win
+    else:
+        # create zeropadded window with shape of signal
+        window_zeropadded = np.zeros(signal.n_samples)
+        window_zeropadded[win_start:win_stop+1] = win
+        signal_win.time = signal_win.time*window_zeropadded
```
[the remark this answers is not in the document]
f-brinkmann: Hm, also true. I don't have a strong preference. Maybe leave this as it is...

pyfar/dsp/dsp.py
```diff
@@ -227,6 +227,202 @@ def spectrogram(signal, dB=True, log_prefix=20, log_reference=1,
...
+        win = np.ones(length[-1]-length[0]+1)
+        win[0:length[1]-length[0]] = fade_in
+        win[length[2]-length[0]+1:length[3]-length[0]+1] = fade_out
+        win_start = length[0]
+        win_stop = length[3]
+
+    # Apply window
+    signal_win = signal.copy()
+    if truncate:
+        signal_win.time = signal_win[..., win_start:win_stop+1].time*win
```
mberz, line 383: Shouldn't this be:
```suggestion
        signal_win.time = signal_win.time[..., win_start:win_stop+1]*win
```

pyfar/dsp/dsp.py
```diff
@@ -227,6 +227,202 @@ def spectrogram(signal, dB=True, log_prefix=20, log_reference=1,
...
+        fade_in = fade_in[:int(fade_in_samples/2)]
+        fade_out_samples = int(2*(length[3]-length[2]))
+        fade_out = sgn.windows.get_window(
+            window, fade_out_samples, fftbins=False)
+        fade_out = fade_out[int(fade_out_samples/2):]
+        win = np.ones(length[-1]-length[0]+1)
+        win[0:length[1]-length[0]] = fade_in
+        win[length[2]-length[0]+1:length[3]-length[0]+1] = fade_out
+        win_start = length[0]
+        win_stop = length[3]
```
mberz, line 378: This is bit of redundant code here: Couldn't this be implemented as a recursive function using the first two cases, simply multiplying left and right sided windows in the end for the last case?

pyfar/dsp/dsp.py
```diff
@@ -227,6 +227,202 @@ def spectrogram(signal, dB=True, log_prefix=20, log_reference=1,
...
+            raise ValueError(
+                "Values in length require window to be longer than signal.")
+
+    # Create window
+    # win_start and win_stop define the first and last sample of the window
+    if length is None:
+        win_samples = signal.n_samples
+        win = sgn.windows.get_window(window, win_samples, fftbins=False)
+        win_start = 0
+        win_stop = signal.n_samples-1
```
mberz, line 345: I don't see so many use-cases for this one, how about you others?

pyfar/dsp/dsp.py
```diff
@@ -227,6 +227,202 @@ def spectrogram(signal, dB=True, log_prefix=20, log_reference=1,
...
+    elif len(length) == 2:
+        if shape == 'symmetric':
+            win_samples = length[1]-length[0]+1
+            win = sgn.windows.get_window(window, win_samples, fftbins=False)
+            win_start = length[0]
+            win_stop = length[1]
```
mberz, line 351: Could we also have a window that is symmetric with respect to t=0? This is usually very useful for non minimum-phase IRs without a linear phase component.

pyfar/dsp/dsp.py
```diff
@@ -227,6 +227,202 @@ def spectrogram(signal, dB=True, log_prefix=20, log_reference=1,
     return frequencies, times, spectrogram
 
 
+def windowing(signal, window='hann', length=None, shape='symmetric',
```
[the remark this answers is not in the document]
sikersten: Good suggestion, changed it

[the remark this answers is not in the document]
sikersten: changed it to second argument, for default see comment below on length=None

pyfar/dsp/dsp.py
```diff
@@ -227,6 +227,202 @@ def spectrogram(signal, dB=True, log_prefix=20, log_reference=1,
...
+        If ``left`` or ``right``, the beginning and the end of the fade is
+        defined by the two values in `length`. The default is ``symmetric``.
+    unit : string
+        Unit of the parameter `length`. Can be set to ``s`` (seconds), ``ms``
+        (milliseconds) or ``samples``. If ``samples``, the values in length
+        denote the first and last sample being included. Time values are
+        rounded to the nearest sample. The default is ``samples``.
+    truncate : bool
+        If ``True``, the signal is truncated to the length of the window.
+        The default is ``False``.
```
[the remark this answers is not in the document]
sikersten: changed it to crop

pyfar/dsp/dsp.py
sikersten: Agree

pyfar/dsp/dsp.py
sikersten: Good suggestion! Suggestion: single value of length (len(length)==1 and shape=='symmetric')

pyfar/dsp/dsp.py
sikersten: The problem is that the right slope depends on length[0] and length[1] for shape='right', but on length[2] and length[3] for the symmetric case. The only alternative I see would be to create a right slope with length[-2] and length[-1] in all cases, but in the case without fade out this would cause obsolete additional computational effort.

pyfar/dsp/dsp.py
sikersten: Yes, changed it

pyfar/dsp/dsp.py
sikersten: Good Point! Added option shape='symmetric_zero', also as default

mberz: 👍

pyfar/dsp/dsp.py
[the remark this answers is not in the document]
mberz: Mh, how about interval/intervals?

pyfar/dsp/dsp.py
mberz: I think computational effort is not really something we need to worry about in this function. I would go for the most readable solution. I was thinking about something like this (my own very simple window function I've been using for my personal projects, just as an example):

``` python
def _half_sided_window(interval, n_samples):
    window_length = np.int(np.diff(interval))
    window = sps.hann(2*np.abs(window_length))
    if np.sign(window_length) == -1:
        window_left = window[:np.abs(window_length)]
        window_full = np.concatenate((
            np.zeros(interval[1]),
            window_left,
            np.ones(n_samples-interval[0])))

    else:
        window_right = window[np.int(np.abs(window_length))::]
        window_full = np.concatenate((
            np.ones(interval[0]),
            window_right,
            np.zeros(n_samples-interval[1])))

    return window_full


def time_window(data, intervals, window='hanning', samples=True):
    n_samples = data.shape[-1]

    intervals = np.asarray(intervals)

    if np.size(intervals) == 2:
        window_full = _half_sided_window(intervals, n_samples)
    elif np.size(intervals) == 4:
        window_left = _half_sided_window(intervals[:2], n_samples)
        window_right = _half_sided_window(intervals[2:], n_samples)
        window_full = window_left * window_right

    data = data.copy() * np.broadcast_to(window_full, data.shape)

    return data
```

pyfar/dsp/dsp.py
```diff
@@ -227,6 +227,219 @@ def spectrogram(signal, dB=True, log_prefix=20, log_reference=1,
     return frequencies, times, spectrogram
 
 
+def time_window(signal, length=None, window='hann', shape='symmetric',
+                unit='samples', crop=True):
```
mberz, line 231: I would go for `crop=False` as default, since a crop will change the length, and in case of a fade in also the linear-phase part of the signal/ir
```suggestion
                unit='samples', crop=False):
```

pyfar/dsp/dsp.py
```diff
@@ -227,6 +227,219 @@ def spectrogram(signal, dB=True, log_prefix=20, log_reference=1,
...
+        win_start = length[0]
+        win_stop = length[3]
+    else:
+        raise ValueError(
+            "Length needs to contain two or four values.")
+
+    # Apply window
+    signal_win = signal.copy()
+    if crop:
+        signal_win.time = signal_win.time[..., win_start:win_stop+1]*win
```
mberz, line 400: One more thing I just realized: If a crop is applied in the beginning of the signal, we also apply a linear-phase part with a negative delay of `np.exp(-1j*win_start)`. This may be potentially unwanted (for example for HRTF measurments, or any other measurments where the original phase needs to be preserved). I would suggest to have more options for the crop operation, for example `'full'` which is the current version, `'none'` which doesn't crop at all, and an additional option which crops to the end of the fade out window. I don't have a good one word description though ;)

pyfar/dsp/dsp.py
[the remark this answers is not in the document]
mberz: I don't think performance is a critical issue here :)

pyfar/dsp/dsp.py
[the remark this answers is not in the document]
f-brinkmann: Interval sounds good to me. I would also vote for keeping the default.

pyfar/dsp/dsp.py
f-brinkmann: Would be fine for me, but I don't have a strong preference for either.

pyfar/dsp/dsp.py
f-brinkmann: I see at least some use cases, and the current syntax would not allow to create a **pure** symmetric window (because the second and first value for ``length``/``interval`` denote the sample with a value of ``1``.  How about this:

pyfar/dsp/dsp.py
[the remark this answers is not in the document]
sikersten: Changed it

pyfar/dsp/dsp.py
sikersten: The problem with an interval of one value and a center parameter is that that the center would need o be between two samples for an even lengh of a window (a pure symmetric). Therefore I suggest to remove interval=None, and instead use the shape='symmetric' opion with start and end, which automatically creates a window with even(pure symmetric) or odd (value at center is 1) window depending on the input.

pyfar/dsp/dsp.py
sikersten: Wrote private functions to increase readability

pyfar/dsp/dsp.py
sikersten: Changed default to crop='none'

pyfar/dsp/dsp.py
sikersten: Good point! Added three options ``'none'``, ``'window'`` and ``'end'``. Any ideas for better descriptions?

mberz: I like the three you chose

pyfar/dsp/dsp.py
```diff
@@ -227,6 +227,407 @@ def spectrogram(signal, dB=True, log_prefix=20, log_reference=1,
...
+    arguments the needed parameters.
+
+    Examples
+    --------
+
+    Options for parameter `shape`.
+
+    >>> import pyfar
+    >>> import numpy as np
+    >>>
```
mberz, line 332:
```suggestion

```

pyfar/dsp/dsp.py
```diff
@@ -227,6 +227,407 @@ def spectrogram(signal, dB=True, log_prefix=20, log_reference=1,
...
+            signal_windowed = pyfar.dsp.time_window(
+                signal, interval=[25,45], shape=shape)
+            ax = pyfar.plot.time(signal_windowed, label=shape)
+        ax.legend(loc='right')
+
+    Window with fade-in and fade-out defined by four values in `interval`.
+
+    >>> import pyfar
+    >>> import numpy as np
+    >>>
```
mberz, line 356:
```suggestion

```

mberz (approved):
2 small formatting things in examples ;)

f-brinkmann (approved):

PR #180, opened 2021-05-08
Feature/reconstructing fractional octave bands

f-brinkmann:

None

- add perfectly reconstructing filterbank
- add testing


sikersten (commented):
Looks good to me (but I have only little experience in this field).
Maybe we find a better solution to relate the 2 fractional_octave_bands functions than have to different ones?

pyfar/dsp/filter.py
```diff
@@ -976,3 +976,187 @@ def _coefficients_fractional_octave_bands(
...
+    # number of frequency bins
+    n_bins = int(n_samples // 2 + 1)
+
+    # fractional octave frequencies
+    _, f_m, f_cut_off = pf.dsp.filter.fractional_octave_frequencies(
+        num_fractions, frequency_range, return_cutoff=True)
+
+    # discard fractional octaves, if the center frequency exceeds
+    # half the sampling rate
+    f_id = f_m < sampling_rate / 2
```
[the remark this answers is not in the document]
f-brinkmann: Good point - I made this a warning, which is consistent with the behavior of `fractional_octave_bands`

f-brinkmann:
> 
> 
> Looks good to me (but I have only little experience in this field).
> Maybe we find a better solution to relate the 2 fractional_octave_bands functions than have to different ones?

I also thought about this, but I think the parameters differ too much across the two functions. I thus made the names very similar, so they both pop up in auto complete. I hope this makes the users aware of the different choices depending on the application.

tests/test_filter.py
```diff
@@ -227,3 +229,59 @@ def test_crossover(impulse):
...
+        y, _ = pfilt.reconstructing_fractional_octave_bands(
+            x, frequency_range=(8e3, 16e3), overlap=overlap, slope=slope,
+            n_samples=2**10)
+        reference = np.loadtxt(os.path.join(
+            os.path.dirname(__file__), "references",
+            f"filter.reconstructing_octaves_{overlap}_{slope}.csv"))
+        # restricting rtol was not needed locally. It was added for tests to
+        # pass on travis ci
+        npt.assert_allclose(
+            y.time, np.atleast_2d(reference), rtol=.01, atol=1e-10)
```
mberz, line 280: That's a very small tolerance, is this maybe related to how the data is read? I'm interested where the tolerance is coming from, since it would be a potentially large error in the filter.

pyfar/dsp/filter.py
```diff
@@ -976,3 +976,189 @@ def _coefficients_fractional_octave_bands(
...
+           J. Acous. Soc. Am., 127(2), 884–895, doi: 10.1121/1.3273888
+
+    Examples
+    --------
+
+    Filter and re-synthesize impulse signal
+
+    >>> import pyfar as pf
+    >>> import numpy as np
+    >>>
```
mberz, line 1045:
```suggestion
```

pyfar/dsp/filter.py
```diff
@@ -976,3 +976,189 @@ def _coefficients_fractional_octave_bands(
...
+    >>> x = pf.signals.impulse(2**12)
+    >>> y, f = pf.dsp.filter.reconstructing_fractional_octave_bands(x)
+    >>> y_sum = pf.Signal(np.sum(y.time, 0), y.sampling_rate)
+    >>>
+    >>> # plot
```
mberz, line 1050:
```suggestion
```
Inline comments shouldn't really be required in such a short snippet. Documentation can be added as text above.

pyfar/dsp/filter.py
```diff
@@ -976,3 +976,189 @@ def _coefficients_fractional_octave_bands(
...
+
+    References
+    ----------
+    .. [#] Antoni, J. (2010). Orthogonal-like fractional-octave-band filters.
+           J. Acous. Soc. Am., 127(2), 884–895, doi: 10.1121/1.3273888
+
+    Examples
+    --------
+
+    Filter and re-synthesize impulse signal
```
mberz, line 1041: - I'd add a legend
- How about a plot showing the original and re-synthesized impulse? Would be quite nice here

pyfar/dsp/filter.py
```diff
@@ -976,3 +976,189 @@ def _coefficients_fractional_octave_bands(
                 order, Wn, btype=btype, output='sos')
         sos[idx, :, :] = sos_coeff
     return sos
+
+
+def reconstructing_fractional_octave_bands(
+        signal, num_fractions=1, frequency_range=(63, 16000),
+        overlap=1, slope=0, n_samples=2**12, sampling_rate=None):
+    """
+    Perfectly reconstructing fractional octave filter bank.
```
mberz, line 985: 1. I wouldn't call this perfectly, since the reconstruction error potentially depends on the length of the FIR filter. Even the paper calls it Orthogonal-like and not orthogonal. Maybe there's a better name for this.
2. Should we add some information on how this is different from the IIR filter bank?

pyfar/dsp/filter.py
```diff
@@ -976,3 +976,189 @@ def _coefficients_fractional_octave_bands(
...
+    # get impulse responses
+    h = pf.dsp.fft.irfft(g, n_samples, sampling_rate, 'none')
+
+    # window
+    h *= spsignal.windows.hann(h.shape[-1])
+
+    # create filter object
+    filt = pf.FilterFIR(h, sampling_rate)
+    filt.comment = (
+        "Perfect reconstructing linear phase fractional octave filter bank."
```
mberz, line 1154:
```suggestion
        "Reconstructing linear phase fractional octave filter bank."
```

mberz (changes requested):
This function uses quite a number of one letter variables, which is not always very well readable. For at least some of these replacing with meaningful variable names might be beneficial. (p, g, P, f,  etc.)

pyfar/dsp/filter.py
f-brinkmann: Changed as suggested :)

pyfar/dsp/filter.py
f-brinkmann: Removed the inline comments but left blank lines between blocks for better readability

pyfar/dsp/filter.py
f-brinkmann: Good idea. I tried a legend but used titles in the end (the legend for the filtered frequency plot had to many entries.)

f-brinkmann (commented):
> 
> 
> This function uses quite a number of one letter variables, which is not always very well readable. For at least some of these replacing with meaningful variable names might be beneficial. (p, g, P, f, etc.)

My idea was to use the same letters as in the paper so it will be easier to compare the implementation and paper. I left these variables as they were but changed others to more meaningful names.

tests/test_filter.py
f-brinkmann: I puzzled here - saving references in csv files worked witout restrictions in other cases. I thought of different number formats across systems, but that's only a guess.

pyfar/dsp/filter.py
mberz: Yes, it was merely the "perfectly" I disagreed with :)

mberz:
> > This function uses quite a number of one letter variables, which is not always very well readable. For at least some of these replacing with meaningful variable names might be beneficial. (p, g, P, f, etc.)
> 
> My idea was to use the same letters as in the paper so it will be easier to compare the implementation and paper. I left these variables as they were but changed others to more meaningful names.

Alright, in that case I can agree :)

mberz (approved):

PR #185, opened 2021-05-14
Minimum phase 

mberz:

Add minimum-phase version of signals


pyfar/dsp/dsp.py
```diff
@@ -355,3 +355,134 @@ def _cross_fade(first, second, indices):
...
+
+    if (pad is True) or (return_magnitude_error is True):
+        sig_minphase_pad = pad_zeros(
+            signal_minphase, signal.n_samples - signal_minphase.n_samples)
+
+        if return_magnitude_error is False:
+            return sig_minphase_pad
+
+        error_mag = (np.abs(signal.freq) - np.abs(sig_minphase_pad.freq))
+        error = pyfar.FrequencyData(error_mag, signal.frequencies)
```
f-brinkmann, line 418: While this is already helpful, the most important question for me when using minimum phase generation always was: What is the largest error in the region where the original signal has relatively much energy. I used the first 60 dB of the dynamic range in AKtools and returned the error in dB. Something like this, maybe:
```
error_range = np.max(np.abs(signal.freq)) * 10**(-60/20)
error_dB = np.abs(20 * np.log10(signal.freq / sig_minphase.freq))
error_dB = np.max(error_dB[np.abs(signal.freq) >= error_range])
```

pyfar/dsp/dsp.py
```diff
@@ -355,3 +355,134 @@ def _cross_fade(first, second, indices):
...
+        (signal_minphase.cshape[0], int(np.floor((signal.n_samples + 1)/2))),
+        dtype=signal.dtype)
+
+    for ch in range(signal_minphase.cshape[0]):
+        signal_minphase.time[ch] = sgn.minimum_phase(
+            signal_flat.time[ch],
+            method=method,
+            n_fft=n_fft)
+
+    if (pad is True) or (return_magnitude_error is True):
```
f-brinkmann, line 410: shouldn't this be two separate if cases? Now zeros are padded `if return_magnitude_error is True`

f-brinkmann (changes requested):

pyfar/dsp/dsp.py
mberz: Yes, this is required for the error calculation. If pad is False, the unpadded version of the result is returned. See below ;)

pyfar/dsp/dsp.py
```diff
@@ -355,3 +355,134 @@ def _cross_fade(first, second, indices):
...
+        sig_minphase_pad = pad_zeros(
+            signal_minphase, signal.n_samples - signal_minphase.n_samples)
```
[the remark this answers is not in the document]
mberz: See comment above

pyfar/dsp/dsp.py
mberz: Mh, the thing is: This is a rather arbitrary measure which is likely to explode in the stop band region of band-pass filters.

mberz:
RTDs link for this branch is found [here](https://pyfar.readthedocs.io/en/feature-minimumphase/pyfar.dsp.html#pyfar.dsp.minimum_phase)

mberz:
This PR requires zero-padding from #183

mberz:
keeping this on the bookshelf: https://www.dsprelated.com/freebooks/filters/Minimum_Phase_Filters.html

f-brinkmann (approved):
Thanks for the effort! Only two minor comments that I would leave up to you

pyfar/dsp/dsp.py
```diff
@@ -850,3 +850,261 @@ def _cross_fade(first, second, indices):
...
+        The ratio between the magnitude of the linear phase version and the
+        minimum phase versions of the filter.
+
+
+    Examples
+    --------
+
+    Minmum-phase version of an ideal impulse with a group delay of 64 samples
+
+    >>> import pyfar
```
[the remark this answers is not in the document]
mberz: Good point, using the import as pf is something I'd prefer

pyfar/dsp/dsp.py
```diff
@@ -850,3 +850,261 @@ def _cross_fade(first, second, indices):
...
+        for h, style in zip(
+                (h_linear, h_min_hom, h_min_hil),
+                ('-', '-.', '--')):
+            pf.plot.time(h, linestyle=style, ax=axs[0])
+            axs[0].grid(True)
+            pf.plot.freq(h, linestyle=style, ax=axs[1])
+            pf.plot.group_delay(h, linestyle=style, ax=axs[2])
+        axs[1].legend(['Linear', 'Homomorphic', 'Hilbert'])
+
+    Return the magnitude ratios between the mimimum and linear phase filters
```
sikersten, line 970:
```suggestion
    Return the magnitude ratios between the minimum and linear phase filters
```

PR #186, opened 2021-05-14
Cyclic time shift

mberz:

Perform cyclic time shifts of signals


pyfar/dsp/dsp.py
```diff
@@ -850,3 +851,56 @@ def _cross_fade(first, second, indices):
     result = first * window_first + second * window_second
 
     return result
+
+
+def time_shift(signal, shift, unit='samples'):
+    """Apply a time-shift to a signal.
```
f-brinkmann, line 857: Add empty line :)

pyfar/dsp/dsp.py
```diff
@@ -850,3 +851,56 @@ def _cross_fade(first, second, indices):
...
+    if shift.size == 1:
+        shift = np.ones(signal.cshape) * shift
+
+    if unit == 'seconds':
+        shift_samples = np.round(shift*signal.sampling_rate).astype(int)
+    elif unit == 'samples':
+        shift_samples = shift.astype(int)
+    else:
+        raise ValueError(
+            f"Unknown unit: {unit}, has to be 'samples' or 'seconds'.")
```
f-brinkmann, line 892:
```suggestion
            f"unit is {unit} but has to be 'samples' or 'seconds'.")
```

tests/test_dsp.py
```diff
@@ -212,6 +213,66 @@ def test_regu_inversion(impulse):
...
+    n_samples = 10
+    test_signal = impulse(n_samples, delay=delay, sampling_rate=sampling_rate)
+
+    # shift to the right
+    shift_samples = 2
+    shifted = dsp.time_shift(test_signal, shift_samples, unit='samples')
+    ref = impulse(
+        n_samples, delay=delay+shift_samples, sampling_rate=sampling_rate)
+
+    npt.assert_allclose(shifted.time, ref.time)
```
f-brinkmann, line 228: Can you add a line for testing if the input remained unchanged for explicity? It's implicitly tested already :)

f-brinkmann (approved):
Looks good to me - I only had minor comments that I would leave up to you.

tests/test_dsp.py
mberz: :+1:

pyfar/dsp/dsp.py
mberz: :+1:

pyfar/dsp/dsp.py
mberz: :+1:

tests/test_dsp.py
```diff
@@ -212,6 +213,66 @@ def test_regu_inversion(impulse):
     npt.assert_allclose(res.freq[:, -1], [0.25])
 
 
+def test_time_shift():
```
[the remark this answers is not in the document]
mberz: I'm leaving this as is, makes spotting issues easier

pyfar/dsp/dsp.py
```diff
@@ -850,3 +851,57 @@ def _cross_fade(first, second, indices):
...
+    ----------
+    signal : Signal
+        The signal to be shifted
+    shift : int, float
+        The time-shift value. A positive value will result in right shift on
+        the time axis (delaying of the signal), whereas a negative value
+        yields a left shift on the time axis (non-causal shift to a earlier
+        time).
+    unit : str, optional
+        Unit of the shift variable, this can be either samples or seconds.
```
sikersten, line 872: To be consistent with units in other functions, I'd suggest to have `'samples'` and `'s` instead of `'seconds'`, also `'ms'` would be nice :)
```suggestion
        Unit of the shift variable, this can be either ``'samples'``,  ``'s'`` for seconds or ``'ms'`` for milliseconds.
```

pyfar/dsp/dsp.py
```diff
@@ -850,3 +851,57 @@ def _cross_fade(first, second, indices):
...
+        By default 'samples' is used. Note that in the case of specifying
```
sikersten, line 873:
```suggestion
        By default ``'samples'`` is used. Note that in the case of specifying
```

tests/test_dsp.py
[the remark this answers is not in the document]
sikersten: I think it would help to split it in different test for the purpose of spotting issues (or use `@pytest.mark.parametrize()`, which does that automatically)

pyfar/dsp/dsp.py
```diff
@@ -850,3 +851,57 @@ def _cross_fade(first, second, indices):
...
+
+    Parameters
+    ----------
+    signal : Signal
+        The signal to be shifted
+    shift : int, float
+        The time-shift value. A positive value will result in right shift on
+        the time axis (delaying of the signal), whereas a negative value
+        yields a left shift on the time axis (non-causal shift to a earlier
+        time).
```
sikersten, line 870: It is not clear, that individual shifts for the channels can be used. Is the resulting error for a mismatch of `cshape` and `shift.shape` easy to understand?

sikersten (changes requested):
Looks good, only minor changes suggested :)

pyfar/dsp/dsp.py
mberz: Frankly I don't really see the need to specify milliseconds, since it's not really more convenient than adding a `e-3` to the specified value. And I don't like the inconsistency in spelling samples, but abbreviating seconds to s either. But I'd let myself get overruled here.

pyfar/dsp/dsp.py
mberz: Indeed a good point, I've added documentation on this and added an example

tests/test_dsp.py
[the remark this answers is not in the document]
mberz: 👍

sikersten (approved):

PR #187, opened 2021-05-17
Feature: Interpolate spectrum

f-brinkmann:

Closes None

- added function and testing for interpolating incomplete spectra

**Note:** testing and processing for method "magnitude_minimum" is at dummy stage. It will be added when minimum phase generation is available. Please review now to make sure we can release this on Friday.


pyfar/dsp/dsp.py
```diff
@@ -351,6 +351,70 @@ def regularized_spectrum_inversion(
...
+            Separate interpolation if the absolute values and unwrapped phase
+        ``'abs_linear'``
+            Interpolation of the absolute values and generation of a linear
+            phase response with a group delay of ``n_samples / 2``
+        ``'abs_minumum'``
+            Interpolation of the absolute values and generation of a minimum
+            phase response
+        ''`abs'``
+            Interpolate the absolute values only. Results in a zero phase
+            signal, which is symmetric around the first sample
```
sikersten, line 384: In the context of spectra, I think "magnitude" is a better term than "absolute value". What's your opinion?

pyfar/dsp/dsp.py
```diff
@@ -351,6 +351,70 @@ def regularized_spectrum_inversion(
...
+        Specifies the input data for the interpolation
+
+        ``'complex'``
+            Separate interpolation of the real and imaginary part
+        ``'abs_unwrap'``
+            Separate interpolation if the absolute values and unwrapped phase
+        ``'abs_linear'``
+            Interpolation of the absolute values and generation of a linear
+            phase response with a group delay of ``n_samples / 2``
+        ``'abs_minumum'``
```
sikersten, line 379:
```suggestion
        ``'abs_minimum'``
```

pyfar/dsp/dsp.py
```diff
@@ -351,6 +351,70 @@ def regularized_spectrum_inversion(
...
+    method : string
+        Specifies the input data for the interpolation
+
+        ``'complex'``
+            Separate interpolation of the real and imaginary part
+        ``'abs_unwrap'``
+            Separate interpolation if the absolute values and unwrapped phase
+        ``'abs_linear'``
+            Interpolation of the absolute values and generation of a linear
+            phase response with a group delay of ``n_samples / 2``
```
sikersten, line 378: I think it would be nice to have a parameter for the group_delay, either as separate parameter or as tuple `method=('abs_linear', group_delay)`

pyfar/dsp/dsp.py
```diff
@@ -351,6 +351,70 @@ def regularized_spectrum_inversion(
...
+        ``'next'`` (see ``scipy.interpolate.interp1d``).
+
+        ``('first', 'second', 'third')``
+            Three element tuple specifies the kind of inter/extrapolation below
+            the lowest frequency (first), between the lowest and highest
+            frequency (second), and above the highest frequency (third)
+        ``'first'``
+            String that specifies the kind of interpolation between the lowest
+            and highest frequency. Values outside the frequency range are
+            taken from the next/previous available value.
```
sikersten, line 399: What's the reason to set nearest neighbor as default?

pyfar/dsp/dsp.py
```diff
@@ -351,6 +351,70 @@ def regularized_spectrum_inversion(
...
+
+    fscale : string, optional
+
+        ``'linear'``
+            Interpolate on a linear frequency axis
+        ``'log'``
```
sikersten, line 405: How would this be connected to `n_samples`?

pyfar/dsp/dsp.py
```diff
@@ -351,6 +351,70 @@ def regularized_spectrum_inversion(
...
+            Interpolate on a logarithmic frequency axis
+
+        The default is ``'linear'``.
+    clip : bool, tuple
+        The interpolated frequency data is clipped to the range specified by
+        this two element tuple. E.g., ``clip=(0, 1)`` will assure that no
+        values smaller than 0 and larger than 1 occur in the interpolated
+        frequency data. The default is ``False`` which does not clip the data.
+    """
```
sikersten, line 414: Return Signal or FrequencyData?

pyfar/dsp/dsp.py
```diff
@@ -351,6 +351,70 @@ def regularized_spectrum_inversion(
...
+            Interpolate the absolute values only. Results in a zero phase
+            signal, which is symmetric around the first sample
+
+    kind : tuple, string
+        Specifies the kind of interpolation as a string. The string has to be
+        ``'linear'``, ``'nearest'``, ``'nearest-up'``, ``'zero'``,
+        ``'slinear'``, ``'quadratic'``, ``'cubic'``, ``'previous'``, or
+        ``'next'`` (see ``scipy.interpolate.interp1d``).
+
+        ``('first', 'second', 'third')``
```
sikersten, line 392: Elegant option!

pyfar/dsp/dsp.py
f-brinkmann: agreed and changed.

pyfar/dsp/dsp.py
f-brinkmann: I thought this would be the most common and safe case, since extrapoation is always risky. We could change this to use 'first' everywhere or remove the option to specify only one method entirely.

pyfar/dsp/dsp.py
f-brinkmann: No connection to `n_samples`. This would interpolate on `np.log(frequency_data.frequencies)`. Not sure if this option is needed often. I included it to improve flexibility.

pyfar/dsp/dsp.py
f-brinkmann: Signal!

pyfar/dsp/dsp.py
sikersten: I'm struggling with the constant extrapolation to low frequencies, for example in the context of room transfer functions this would lead to physically wrong results.
I like the `('first', 'second', 'third')` option. Alternatively we could define a parameter referring to the interpolation and one or two defining the extrapolation.

f-brinkmann: That's true. I think extrapolation will only yield (approximately) correct results in special cases anyhow. Considering this, I would suggest to only keep the long version `('first', 'second', 'third')` and discard the short version `'first'`

mberz (commented):
I had the idea to make this class based with an interface very similar to the scipy interpolators, see https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp1d.html
This way one would keep the interpolator and use the `__call__` dunder function to perform the actual interpolation step. The `__call__` function then returns a Signal, or FrequencyData, depending on the requested input.
We could simply inherit from interp1d for example.

pyfar/dsp/dsp.py
```diff
@@ -811,6 +813,338 @@ def regularized_spectrum_inversion(
...
+    ----------
+    data : FrequencyData
+        Input data to be interpolated. `data.fft_norm` must be `'none'`.
+    method : string
+        Specifies the input data for the interpolation
+
+        ``'complex'``
+            Separate interpolation of the real and imaginary part
+        ``'magnitude_unwrap'``
+            Separate interpolation if the magnitude and unwrapped phase values
```
mberz, line 832: I would call this `magnitude_phase`

pyfar/dsp/dsp.py
```diff
@@ -811,6 +813,338 @@ def regularized_spectrum_inversion(
...
+        ``'magnitude_linear'``
+            Interpolation of the magnitude values and generation of a linear
+            phase response with a specified group delay. (See parameters
+            `group_delay` and `unit` below and
+            :py:func:`~pyfar.dsp.linear_phase`).
+        ``'magnitude_minimum'``
+            Interpolation of the magnitude values and generation of a minimum
+            phase response
```
mberz, line 840: I have thought about this a bit:
These options are pretty much independent from the actual interpolation step, but apply further processing on the spectrum, that is they do phase manipulation, independent of the interpolation of the spectrum. For this reason I would suggest removing them from the interpolation. This would make this class much slimmer, since it already has quite a number of options already. We can add examples on how to generate a lin phase variant of the interpolated signal for example.

pyfar/dsp/dsp.py
```diff
@@ -811,6 +813,338 @@ def regularized_spectrum_inversion(
...
+        ``False`` which does not clip the
+        data.
+    group_delay : float, optional
+        The group delay of the linear phase response. Must be specified if
+        `method` is ``'magnitude_linear'``. The unit of the group delay is
+        given by `unit` (see below).
+    unit : string, optional
+        Unit of the group delay. Can be ``'s'`` for seconds, ``'ms'`` for
+        milliseconds, ``'mus'`` for microseconds, or ``'samples'``. The
+        default is ``'samples'``.
```
mberz, line 888: These could be removed following my comment above

pyfar/dsp/dsp.py
```diff
@@ -811,6 +813,338 @@ def regularized_spectrum_inversion(
...
+    >>>     group_delay=32)
+    >>> signal = interpolator(64, 44100)
+
+    Inspect the data. Note that this plot can be also created by the
+    interpolator object by ``signal = interpolator(64, 44100, show=True)``
+
+    >>> with pf.plot.context():
+    >>>     _, ax = plt.subplots(2, 2)
+    >>>
+    >>>     # time signal (linear amplitude)
```
mberz, line 922:
```suggestion
```
I'd remove these in-line comments and replace with text explaining what's going on above

pyfar/dsp/dsp.py
```diff
@@ -811,6 +813,338 @@ def regularized_spectrum_inversion(
...
+    >>>     ax[0][1].set_xlim(0, signal.sampling_rate/2)
+    >>>
+    >>>     # frequency plot (log x-axis)
+    >>>     pf.plot.freq(signal, dB=False, ax=ax[1][1], label='intput')
+    >>>     pf.plot.freq(data, dB=False, ax=ax[1][1],
+    >>>                  c='r', ls='', marker='.', label='output')
+    >>>     min_freq = np.min([signal.sampling_rate / signal.n_samples,
+                               data.frequencies[0]])
+    >>>     ax[1][1].set_xlim(min_freq, signal.sampling_rate/2)
+    >>>     ax[1][1].legend(loc='best')
```
mberz, line 941:
```suggestion
    >>>     pf.plot.time(signal, ax=ax[0, 0])
    >>>     pf.plot.time(signal, ax=ax[1][0], dB=True)
    >>>     pf.plot.freq(signal, dB=False, xscale="linear", ax=ax[0, 1])
    >>>     pf.plot.freq(data, dB=False, xscale="linear",
    ...                  ax=ax[0, 1], c='r', ls='', marker='.')
    >>>     ax[0, 1].set_xlim(0, signal.sampling_rate/2)
    >>>     pf.plot.freq(signal, dB=False, ax=ax[1, 1], label='input')
    >>>     pf.plot.freq(data, dB=False, ax=ax[1, 1],
    ...                  c='r', ls='', marker='.', label='output')
    >>>     min_freq = np.min([signal.sampling_rate / signal.n_samples,
    ...                        data.frequencies[0]])
    >>>     ax[1, 1].set_xlim(min_freq, signal.sampling_rate/2)
    >>>     ax[1, 1].legend(loc='best')
```

pyfar/dsp/dsp.py
```diff
@@ -811,6 +813,338 @@ def regularized_spectrum_inversion(
     return inverse
 
 
+class interpolate_spectrum():
+    """
+    Interpolate incomplete spectrum to complete single sided spectrum.
```
mberz, line 818:
```suggestion
    Interpolate an incomplete spectrum to a complete single sided spectrum.
```

pyfar/dsp/dsp.py
```diff
@@ -811,6 +813,338 @@ def regularized_spectrum_inversion(
...
+
+    TODO: Add reference to minimum phase functions
+
+    Parameters
+    ----------
+    data : FrequencyData
+        Input data to be interpolated. `data.fft_norm` must be `'none'`.
```
mberz, line 825: Maybe add some remarks on this being supposed to interpolate TFs

pyfar/dsp/dsp.py
```diff
@@ -811,6 +813,338 @@ def regularized_spectrum_inversion(
...
+                                ax=ax[0][1], c='r', ls='', marker='.')
+                ax[0][1].set_xlim(0, sampling_rate/2)
+                # frequency plot (log x-axis)
+                pyfar.plot.freq(signal, dB=False, ax=ax[1][1], label='intput')
+                pyfar.plot.freq(self._input, dB=False, ax=ax[1][1],
+                                c='r', ls='', marker='.', label='output')
+                min_freq = np.min([sampling_rate / n_samples,
+                                   self._input.frequencies[0]])
+                ax[1][1].set_xlim(min_freq, sampling_rate/2)
+                ax[1][1].legend(loc='best')
```
mberz, line 1129: I'm getting more and more fond of the idea of dropping show functions, but keeping the code here in the Examples section. This way anyone interested in visualizing the data can do so and very easily modify the resulting plots, making this much more flexible.
I'd remove the show option, but keep the example.
```suggestion
```

pyfar/dsp/dsp.py
```diff
@@ -811,6 +813,338 @@ def regularized_spectrum_inversion(
...
+        if self._clip:
+            signal.freq = \
+                np.clip(np.abs(signal.freq), self._clip[0], self._clip[1]) \
+                * np.exp(-1j * phase(signal))
+
+        # generate linear or minimum phase
+        if self._method == "magnitude_minimum":
+            pass
+        elif self._method == "magnitude_linear":
+            signal = linear_phase(signal, self._group_delay, self._unit)
```
mberz, line 1107: See above comment on removing these

pyfar/dsp/dsp.py
```diff
@@ -811,6 +813,338 @@ def regularized_spectrum_inversion(
...
+            freq = interpolated[0]
+
+        # get initial signal
+        signal = pyfar.Signal(freq, sampling_rate, n_samples, "freq")
+
+        # clip the magnitude
+        if self._clip:
+            signal.freq = \
+                np.clip(np.abs(signal.freq), self._clip[0], self._clip[1]) \
+                * np.exp(-1j * phase(signal))
```
mberz, line 1101:
```suggestion
            signal.freq = np.clip(
                np.abs(signal.freq),
                self._clip[0],
                self._clip[1]) * np.exp(-1j * phase(signal))
```

pyfar/dsp/dsp.py
```diff
@@ -811,6 +813,338 @@ def regularized_spectrum_inversion(
...
+        """
+        Return frequencies for creating or quering interpolation objects.
+
+        In case logfrequencies are requested, 0 Hz entries are replaced by
+        the next highest frequency, because the logarithm of 0 does not exist.
+        """
+        if self._fscale == "log":
+            if frequencies[0] == 0:
+                frequencies[0] = frequencies[1]
+            frequencies = np.log(frequencies)
```
mberz, line 1143: I'd change the behavior of this to set the magnitude at 0 Hz to 0, since this would be a static sound pressure, or dc voltage.
But this is personal preference, since I almost always deal with signals which are DC free

mberz (changes requested):

pyfar/dsp/dsp.py
f-brinkmann: Changed as suggested

pyfar/dsp/dsp.py
f-brinkmann: Yes, good idea. I added a remark below the one line doc string.

pyfar/dsp/dsp.py
f-brinkmann: changed as suggested

pyfar/dsp/dsp.py
f-brinkmann: I added information to the text above, removed some comments and kept them where I felt they aid a quicker understanding of the code.

pyfar/dsp/dsp.py
f-brinkmann: agreed :)

pyfar/dsp/dsp.py
f-brinkmann: Agreed - this was the old me trying to put too much into one function...

pyfar/dsp/dsp.py
f-brinkmann: yep :)

pyfar/dsp/dsp.py
f-brinkmann: yes :)

pyfar/dsp/dsp.py
f-brinkmann: Let's discuss this on Friday...

pyfar/dsp/dsp.py
f-brinkmann: I thought about this too. There are examples where the current version works better, e.g., interpolating reflexion coefficients of walls, and where your suggestion works better, e.g., interpolating HRTFs of room transfer functions. I thought about making this an optional input parameter but did not due it to keep things simlpe and because changing the value afterwards is easy (`signal.freq[:::, 0]=0`).

mberz: Makes sense

pyfar/dsp/dsp.py
```diff
@@ -811,6 +813,315 @@ def regularized_spectrum_inversion(
     return inverse
 
 
+class interpolate_spectrum():
```
mberz, line 816: I missed this the first time: Since this is a class, I'd use the class naming conventions here
```suggestion
class InterpolateSpectrum():
```

mberz (changes requested):

pyfar/dsp/dsp.py
f-brinkmann: thought about it as well and than used small letters in analogy to the scipy interpolation classes to indicate the function like use. But I don't have a strong opinion - maybe discuss with the others on Friday?

mberz: Sure, I think the name `interp1d` is more a legacy kinda thing, to me it seems that the majority of scipy interpolators implemented as classes now follow the class naming convention (see [here](https://docs.scipy.org/doc/scipy/reference/interpolate.html)). There's some implemented as functions, not to be confused ;)

f-brinkmann: True - I did not take a look at all interpolation classes before. Changed to `InterpolateSpectrum` :)

sikersten (commented):
Minor Comments, approve it for the release, as it is urgent

pyfar/dsp/dsp.py
```diff
@@ -811,6 +813,318 @@ def regularized_spectrum_inversion(
...
+    >>>     # time plot (linear and logarithmic amplitude)
+    >>>     pf.plot.time(signal, ax=ax[0, 0])
+    >>>     pf.plot.time(signal, ax=ax[1, 0], dB=True)
+    >>>     # frequency plot (linear x-axis)
+    >>>     pf.plot.freq(signal, dB=False, xscale="linear", ax=ax[0, 1])
+    >>>     pf.plot.freq(data, dB=False, xscale="linear",
+    ...                  ax=ax[0, 1], c='r', ls='', marker='.')
+    >>>     ax[0, 1].set_xlim(0, signal.sampling_rate/2)
+    >>>     # frequency plot (log x-axis)
+    >>>     pf.plot.freq(signal, dB=False, ax=ax[1, 1], label='intput')
```
mberz, line 926:
```suggestion
    >>>     pf.plot.freq(signal, dB=False, ax=ax[1, 1], label='input')
```

pyfar/dsp/dsp.py
```diff
@@ -811,6 +813,318 @@ def regularized_spectrum_inversion(
...
+            pf.plot.time(signal, ax=ax[0, 0])
+            # time signal (log amplitude)
+            pf.plot.time(signal, ax=ax[1, 0], dB=True)
+            # frequency plot (linear x-axis)
+            pf.plot.freq(signal, dB=False, xscale="linear", ax=ax[0, 1])
+            pf.plot.freq(data, dB=False, xscale="linear",
+                         ax=ax[0, 1], c='r', ls='', marker='.')
+            ax[0, 1].set_xlim(0, signal.sampling_rate/2)
+            # frequency plot (log x-axis)
+            pf.plot.freq(signal, dB=False, ax=ax[1, 1], label='intput')
```
mberz, line 958:
```suggestion
            pf.plot.freq(signal, dB=False, ax=ax[1, 1], label='input')
```

pyfar/dsp/dsp.py
```diff
@@ -811,6 +813,318 @@ def regularized_spectrum_inversion(
...
+                pyfar.plot.time(signal, ax=ax[0, 0])
+                # time signal (log amplitude)
+                pyfar.plot.time(signal, ax=ax[1, 0], dB=True)
+                # frequency plot (linear x-axis)
+                pyfar.plot.freq(signal, dB=False, xscale="linear", ax=ax[0, 1])
+                pyfar.plot.freq(self._input, dB=False, xscale="linear",
+                                ax=ax[0, 1], c='r', ls='', marker='.')
+                ax[0, 1].set_xlim(0, sampling_rate/2)
+                # frequency plot (log x-axis)
+                pyfar.plot.freq(signal, dB=False, ax=ax[1, 1], label='intput')
```
mberz, line 1103:
```suggestion
                pyfar.plot.freq(signal, dB=False, ax=ax[1, 1], label='input')
```

mberz (commented):
Mr Intput :)

mberz (approved):

PR #190, opened 2021-05-27
Improved filter processing

### LINKED ISSUE #88: Filter objects can not be saved
## Description

Filter objects can currently not be save because they contain filter functions (e.g. `lfilter`) that would require pickling to be saved.

## Suggested Solution

Instead of the filter functions, the filter objects could contain strings (e.g., `'lfilter'`). The filter functions could than be returned by a new function, e.g., `filter_func = Filter.get_filter_func('lfilter')`.

mberz:

Related to #88 

The function applying a filter to input data is no longer stored as handle, but handled through a static class method.
This should result in filter objects that can be saved using the pyfar i/o routines

### DIFF
--- a/pyfar/classes/filter.py
+++ b/pyfar/classes/filter.py
@@ -84,7 +84,6 @@ class Filter(object):
     def __init__(
             self,
             coefficients=None,
-            filter_func=None,
             sampling_rate=None,
             state=None,
             comment=None):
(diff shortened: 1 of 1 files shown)


Agent49 (approved):
Looks good to me. Filters can now be saved (already implemented in #192.

PR #195, opened 2021-05-28
Replace the mock in the fractional octave test

### LINKED ISSUE #127: Fractional octaves: test_sum_bands_din
## Description

The implementation of the `impulse_mock` fixture used in `test_sum_bands_din` disagrees with the testing guideline. Unfortunately, the test only works for a sufficient high number of samples in combination with a specific sampling rate. So, if the default `impulse` fixture should be used, an adjustment of its parameters would be required. This could be an option, but maybe change of the tests logic would be a better idea?

mberz:

Closes #127 

- Replace the mock with a signal

### DIFF
--- a/tests/test_fractional_octaves.py
+++ b/tests/test_fractional_octaves.py
@@ -71,21 +71,23 @@ def test_fractional_coeff_oct_filter_iec():
     assert actual.shape == (1, order, 6)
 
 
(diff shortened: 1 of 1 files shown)


f-brinkmann (approved):
Looking good - only the error needs to be fixed, but that looks easy :)

PR #198, opened 2021-06-01
Approved: 2D Plots (will be merged after 0.3.0 release)

f-brinkmann:

First draft of a 2D plot for discussion.
The implmentation is not ready yet:
- New private utils might be introduced to reduce code redundancy
- Interaction will be added

- add `time2d` for discussion. Other plots would be similar
- for example use the following for experimenting:

```
import pyfar as pf
import numpy as np

%matplotlib qt

# test signal
delays = np.array(np.sin(np.linspace(0, 2*np.pi, 45))*50 + 55, dtype=int)
amplitudes = 1 - .5 * np.sin(np.linspace(0, 2*np.pi, 45))
x = pf.signals.impulse(256, delays, amplitudes)

pf.plot.time2d(x)
```


sikersten (changes requested):
Let's collect possibilities of plotting spatial data first, I opened issue #200 for that purpose

pyfar/plot/two_d.py
```diff
@@ -5,6 +5,98 @@
...
+        >>> delays = np.round(100 * np.sin(angles / 180 * np.pi)).astype(int)
+        >>> signal = pf.signals.impulse(128, delays)
+        >>> # plot the signal
+        >>> pf.plot.time2d(signal, points=angles)
+    """
+
+    with context(style):
+        ax, qm, cb = _two_d._time2d(
+            signal.flatten(), dB, log_prefix, log_reference, unit,
+            points, orientation, cmap, colorbar, ax, **kwargs)
```
sikersten, line 85: Signals of cshape higher than 1D are flattened. This is not compatible with points haveing the same shape as cshape:
```
import pyfar as pf
import numpy as np
# generate the signal
angles = np.ones((2, 180)) * np.arange(0, 180)
delays = np.round(100 * np.sin(angles / 180 * np.pi)).astype(int)
signal = pf.signals.impulse(128, delays)
# plot the signal
pf.plot.time2d(signal, points=angles, orientation='horizontal')
```
Suggestion:
```suggestion
            signal.flatten(), dB, log_prefix, log_reference, unit,
            np.asarray(points).flatten(), orientation, cmap, colorbar, ax, **kwargs)
```
What about describing this behavior explicitly in the docs?
Also, this is a bit different for spectrogram, where you cycle through the channels. Maybe it would make sense to have the cycling as a second option?

pyfar/plot/two_d.py
```diff
@@ -5,6 +5,98 @@
 from . import _interaction as ia
 
 
+def time2d(signal, dB=False, log_prefix=20, log_reference=1, unit=None,
+           points=None, orientation="vertical",
```
sikersten, line 9: What's the idea behind `orientation="vertical"` as default?

pyfar/plot/two_d.py
f-brinkmann: Tue, I did not think about that. In this case I would suggest to raise an error if a signal has a multi-dimensional `cshape`. Flattening the signal would make less sense in the cases that I am thinking of.

pyfar/plot/two_d.py
f-brinkmann: That is simply my preference, because I had to chose some default. Would you prefer `'horitzontal'` as the default?

sikersten: I do not have a preference, it's fine.

pyfar/plot/two_d.py
sikersten: Agree :)

tests/test_plot.py
```diff
@@ -326,6 +326,18 @@ def test_2d_colorbar_assertion(function, impulse_45_channels):
                  ax=[plt.gca(), plt.gca()])
 
 
+@pytest.mark.parametrize('function', [(plot.time2d)])
+def test_2d_cshape_assertion(function):
+    """
+    Test assertion when passing an array of axes but not having a colobar.
+    """
```
sikersten, line 333:
```suggestion

```

sikersten (commented):
Thanks, the restriction for `cshape` makes sense to me. Only on minor comment...

pyfar/plot/two_d.py
```diff
@@ -5,6 +5,99 @@
...
+
+    .. plot::
+        >>> import pyfar as pf
+        >>> import numpy as np
+        >>> # generate the signal
+        >>> angles = np.arange(0, 180)
+        >>> delays = np.round(100 * np.sin(angles / 180 * np.pi)).astype(int)
+        >>> signal = pf.signals.impulse(128, delays)
+        >>> # plot the signal
+        >>> pf.plot.time2d(signal, points=angles)
```
al-de, line 80: When I try to run `make html` in the `docs` folder I get the following warning message, that is pointing to these lines, if I get it right:
`/pyfar/pyfar/plot/__init__.py:docstring of pyfar.plot.two_d.time2d:53: WARNING: Error in "plot" directive:
maximum 2 argument(s) allowed, 43 supplied.`
I'm not sure how to solve this, as I see no problem with this code. Could you try to reproduce this warning?

al-de (changes requested):
Functionality and testing looks good to me, thanks a lot for implementing.
I only found that one doubling in the docstring and got a warning when I try to run `make html` in the `doc` folder.
Apart from this everything looks fine.

pyfar/plot/two_d.py
f-brinkmann: Ups, there was a blank line missing after `.. plot::`. Thanks for finding this!

pyfar/plot/two_d.py
```diff
@@ -5,6 +5,99 @@
...
+           style='light', **kwargs):
+    """
+    Surface plot of the time signal.
+
+    Plots ``signal.time`` and passes keyword arguments (`kwargs`) to
+    ``matplotlib.pyplot.pcolormesh()``.
+
+    Parameters
+    ----------
+
```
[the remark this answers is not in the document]
f-brinkmann: thanks, corrected.

al-de (approved):

sikersten (approved):

sikersten:
As discussed in the weekly, we'll not merge this PR into develop before releasing 0.3.0

PR #202, opened 2021-06-15
Feature/plot multidimensional signals

### LINKED ISSUE #197: Plots for multi-dimensional signals fail
## General

* pyfar version: all

## Description

Pyfar line plots to not work for multidimensional signals.
I'd propose flattening and adjusting the legend to represent the cshape of the plotted signal object, i.e. `'Ch 1,2'`. instead of the flattened index.

f-brinkmann:

Closes #197

- Flatten signals in `pyfar.line.*` in calls of private plot functions
- Change `pyfar.plot._interaction` to correctly handle and display multidimensional Signals and channel names
- Add test to `test_interaction`

### DIFF
--- a/pyfar/plot/_interaction.py
+++ b/pyfar/plot/_interaction.py
@@ -62,7 +62,7 @@ class Cycle(object):
         """
         Parameters
         ----------
-        chsape : tupel, array like, int
+        cshape : tupel, array like, int
             cshape of the signal
         index : int, optional
             index of the current channel. The default is 0
--- a/pyfar/plot/_interaction.py
+++ b/pyfar/plot/_interaction.py
@@ -62,7 +62,7 @@ class Cycle(object):
         """
         Parameters
         ----------
-        chsape : tupel
+        chsape : tupel, array like, int
             cshape of the signal
         index : int, optional
             index of the current channel. The default is 0
--- a/pyfar/plot/_interaction.py
+++ b/pyfar/plot/_interaction.py
@@ -58,17 +58,21 @@ class Cycle(object):
     """ Cycle class implementation inspired by itertools.cycle. Supports
     circular iterations into two directions.
     """
-    def __init__(self, n_channels, index=0):
+    def __init__(self, cshape, index=0):
         """
         Parameters
         ----------
-        n_channels : int
-            number of channels in the signal
+        chsape : tupel
+            cshape of the signal
         index : int, optional
             index of the current channel. The default is 0
         """
-        self._n_channels = n_channels
+        self._n_channels = np.prod(cshape)
         self._index = index
+
+        self._channels = []
+        for index in np.ndindex(cshape):
+            self._channels.append(index)
 
     def increase_index(self):
         self._index = (self.index + 1) % self.n_channels
@@ -81,11 +85,18 @@ class Cycle(object):
 
     @property
     def index(self):
-        return self._index
+        return int(self._index)
 
     @property
     def n_channels(self):
         return self._n_channels
+
+    @property
+    def current_channel(self):
+        channel = self._channels[self.index]
+        if len(channel) == 1:
+            channel = channel[0]
+        return channel
 
 
 class EventEmu(object):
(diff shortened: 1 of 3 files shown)


pyfar/plot/line.py
```diff
@@ -16,7 +16,8 @@ def time(signal, dB=False, log_prefix=20, log_reference=1, unit=None, ax=None,
     Parameters
     ----------
     signal : Signal, TimeData
-        The input data to be plotted.
+        The input data to be plotted. Multidimensional data are flattened for
+        plotting.
```
sikersten, line 20: I think the statement `Multidimensional data are flattened for plotting.` is a little confusing, as it only means that all channels are shown. When cycling through the channels, the dimensions are somehow kept in the sense of multidimensional indexes. What's meant with this statement is that `Multidimensional data are flattened for plotting, so all channels are shown.`, isn't it? Maybe it can be formulated a little more precisely.
This also holds for the other changed docstrings.

sikersten (commented):
Implementation is fine! Just a small comment on documentation

pyfar/plot/line.py
f-brinkmann: Good point. How about not mentioning this at all? One could argue that plotting all channels is the expected behavior and the detail that we internally flatten the signal to achieve this is not relevant for the user.

mberz: I'd describe in which order the channels are plotted (and cycled). The used operation - which in this case is flattening - isn't necessarily meaningful to the users.

tests/test_interaction.py
```diff
@@ -317,8 +317,8 @@ def test_toggle_colormap():
     plt.close("all")
 
 
-def test_cycle_and_toggle_lines():
-    """Test toggling and cycling channels of a Signal."""
+def test_cycle_and_toggle_lines_1D_signal():
```
mberz, line 320:
```suggestion
def test_cycle_and_toggle_lines_1d_signal():
```
snake case

tests/test_interaction.py
```diff
@@ -350,6 +350,42 @@ def test_cycle_and_toggle_lines():
     plt.close("all")
 
 
+def test_cycle_and_toggle_lines_2D_signal():
```
mberz, line 353:
```suggestion
def test_cycle_and_toggle_lines_2d_signal():
```
snake case

pyfar/plot/_interaction.py
```diff
@@ -58,18 +58,22 @@ class Cycle(object):
...
     circular iterations into two directions.
     """
-    def __init__(self, n_channels, index=0):
+    def __init__(self, cshape, index=0):
         """
         Parameters
         ----------
-        n_channels : int
-            number of channels in the signal
+        chsape : tupel
```
mberz, line 65:
```suggestion
        chsape : tupel, array_like, int
```

mberz (commented):
Plotting & cycling works well.

Only some small things.

pyfar/plot/_interaction.py
f-brinkmann: Yes, changed as it works with all! We only pass tuples so far...

tests/test_interaction.py
f-brinkmann: changed :)

tests/test_interaction.py
f-brinkmann: changed

pyfar/plot/line.py
f-brinkmann: Good point - I've left the mention of flattening and added an example of how multi-dimensional data would be plotted.

pyfar/plot/_interaction.py
```diff
@@ -58,18 +58,22 @@ class Cycle(object):
...
     circular iterations into two directions.
     """
-    def __init__(self, n_channels, index=0):
+    def __init__(self, cshape, index=0):
         """
         Parameters
         ----------
-        n_channels : int
-            number of channels in the signal
+        chsape : tupel, array like, int
```
sikersten, line 65:
```suggestion
        cshape : tupel, array like, int
```

sikersten (approved):

PR #205, opened 2021-07-02
Feature/save builtins

Agent49:
Support for writing/reading a subset of pythons builtin types.

### DIFF
--- a/pyfar/io/io.py
+++ b/pyfar/io/io.py
@@ -230,6 +230,11 @@ def read(filename):
                         'different versions of Pyfar.')
                 collection[name] = obj
 
+        if 'builtin_wrapper' in collection:
+            for key, value in collection['builtin_wrapper'].items():
+                collection[key] = value
+            collection.pop('builtin_wrapper')
+
     return collection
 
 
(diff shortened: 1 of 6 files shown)


tests/test_io.py
```diff
@@ -369,6 +369,32 @@ def test_write_read_numpy_ndarrays(tmpdir):
...
+        any_bool=any_bool,
+        **dict_of_builtins
+    )
+
+    actual = io.read(filename)
+    assert isinstance(any_int, int)
+    assert actual['any_int'] == any_int
+    assert isinstance(any_bool, bool)
+    assert actual['any_bool'] == any_bool
+    # checks if `dict_of_builtins` is a subset of `actual`
```
f-brinkmann, line 394: fancy - did not know that one!

pyfar/io/io.py
```diff
@@ -230,12 +230,18 @@ def read(filename):
...
+            collection.pop('builtin_wrapper')
+
     return collection
 
 
 def write(filename, compress=False, **objs):
     """
-    Write any compatible pyfar object or numpy array as .far file to disk.
+    Write any compatible pyfar object or numpy array and often used builtin
+    types as .far file to disk.
```
f-brinkmann, line 244: Are there any pyfar objects left, that we do not support? I would simplify this to:

```suggestion
    Write pyfar objects, numpy arrays, and Python built in types as .far file to disk.
```

f-brinkmann (approved):
Looks good to me - had only one minor suggestion for the docstring.

tests/test_io.py
Agent49: Yes it's nice and fancy :D

pyfar/io/io.py
Agent49: Well not all builtin types are actually supported, but those which are often used. But You're right with the pyfar objects. They are all supported at moment.

sikersten (approved):

Issue #210, opened 2021-09-08
Refactor plot mudule and testing thereof


f-brinkmann:
## General

* pyfar version: develop branch

## Description

The plot module should be prepared for upcoming new features:

- discuss usage of `plt.tight_layout()`. This is kind of annoying: In regular single plot and subplot it improves the layout, in complicated cases it fucks thinkgs up. There should be a way to only use tight layout in the former cases (check code of tight layout)
- actually assert if saved plots differ (use `assert compare_images() is None` instead of only `compare_images()`)
- separate line and 2D-Plots
- separate plot functions and helper/utilities
- improve code redundancy in tests

Issue #200, opened 2021-06-04
Feature: Plot spatial data


sikersten:
## Description

There are many possibilities to represent spatial data. Let's collect possible types of 3D plots here, to get an overview before we start to implement. This will enable to find proper function names and get an idea of the use cases.

sikersten:
- 2D surface plot: time/frequency data over spatial data (e.g., azimuth angle, x-coordinate ...), color-coded amplitude; example: HRTFs over azimuth
- 2D surface plot: spatial data on x- and y-axis, color-coded magnitude; example: 3D sound field at certain time step
- 3D surface plot: color-coded magnitude on 3D surface; example: 3D sound field on sphere, surface wave propagation on object
- 3D surface plot: (color-coded) balloon plot; example: directivity

f-brinkmann:
For this module, I would suggest to restrict ourselfs to planar plots and exclude ballon plots. In this case, we would have all plots from `pyfar.plot.line` except the spectrogram because it does not have a 3D version, and custom_subplots because this function could be used with the new plots as well.

sikersten:
Options:
- contour
- con

PR #211, opened 2021-09-08
Refactor colorbars

### LINKED ISSUE #208: plot.spectrogram does not work with subplots and does not a parameter to control if and where a colorbar is added
## General

* pyfar version: all

This is required before targeting #200

## Description

This is an issue consisting of two interwoven problems:

**Problem 1**: `plot.spectrogram` clears the entire figure and thus does not work with subplots. The entire figure is cleared to make sure that potentially existing colorbars are deleted before plotting. Unfortunately there is no way of detecting a colorbar that belongs to an already existing axis of a spectrogram

**Problem 2**: There is no way to control if and where a colorbar is added to the plot.

**Solution for 1**
- `plot.spectrogram` can receive and return either a single axis, or an array of two axes.
- In an array of two axes, the second axis belongs to the colorabar

**Solution for 2**
- add an optional parameter `colobar=True`

**New behavior depending on input**
- `colorbar=True, axis=None`: Use current axis, steal space for colorbar from current axis, return array of axis
- `colorbar=True, axis=ax`: Same as above, but use passed axes instead of current
- `colorbar=True, axis=[ax, ax]`: Same as above but use second axis for colorbar
- `colorbar=False, axis=None`: Use current axis, no colorbar, return single axis
- `colorbar=False, axis=ax`: Same as above, but use passed axis instead of current
- `colorbar=False, axis=[ax ax]`: Throw an error.

f-brinkmann:

Closes #208

- merged private functions `plot._line._spectrogram` and `plot._line._spectrogram_cb`
- added optional parameter to control if colorbar is plotted
- extended existing `ax` parameter to control where colorbar is plotted
- minor structural improvements
- added testing for colorbar options

Next steps #210, #214, #215, and #200


Agent49 (approved):
Looks good to me.

pyfar/plot/line.py
```diff
@@ -305,7 +305,19 @@ def spectrogram(signal, dB=True, log_prefix=20, log_reference=1,
     Returns
     -------
     ax : matplotlib.pyplot.axes
-        Axes or array of axes containing the plot.
+        If `colorbar` is ``True`` an array of two axes is returned. The first
+        is the axis on which the data is plotted, the second is the axis of the
+        coloarbar. If `colorbar` is ``False``, only the axis on which the data
```
sikersten, line 310:
```suggestion
        colorbar. If `colorbar` is ``False``, only the axis on which the data
```

pyfar/plot/_line.py
```diff
@@ -462,15 +462,22 @@ def _deal_time_units(unit='s'):
...
-                 cmap=mpl.cm.get_cmap(name='magma'), ax=None):
+                 cmap=mpl.cm.get_cmap(name='magma'), colorbar=True, ax=None):
     """Plot the magnitude spectrum versus time.
 
     See pyfar.line.spectogram for more information.
+
+    Note: this function always returns only the axis of the actual plot and
+    never an array of axes containing also the axis of the colorbar. This makes
+    handling interactions easier. The axis of the colorbar is added in
+    pyfar.line.spectrogram.
```
sikersten, line 473: This is a little confusing as ax[0], qm, cb are returned. Could you reformulate it to explain what these 3 objects are? :)

pyfar/plot/line.py
```diff
@@ -257,7 +257,7 @@ def group_delay(signal, unit=None, xscale='log', ax=None, style='light',
 def spectrogram(signal, dB=True, log_prefix=20, log_reference=1,
                 yscale='linear', unit=None, window='hann', window_length=1024,
                 window_overlap_fct=0.5, cmap=mpl.cm.get_cmap(name='magma'),
-                ax=None, style='light'):
+                colorbar=True, ax=None, style='light'):
```
sikersten, line 260: Colorbar parameter is missing in the parameter list below

sikersten (changes requested):
Nice improvement! Only minor comments. What do you think about an example on how to use the returned colorbar? Maybe in a configuration with several subplots?

pyfar/plot/line.py
f-brinkmann: changed :)

pyfar/plot/line.py
f-brinkmann: Thanks. I added that.

pyfar/plot/_line.py
f-brinkmann: It is documented in the public function - can you check if you find the information there sufficient?

sikersten:
```suggestion
    Note: this function always returns only the axis of the actual plot together with the quadmesh and colorbar, and
    never an array of axes containing also the axis of the colorbar. This makes
    handling interactions easier. The axis of the colorbar is added in
    pyfar.line.spectrogram.
```
It think, then it's clear :)

f-brinkmann:
> 
> 
> Nice improvement! Only minor comments. What do you think about an example on how to use the returned colorbar? Maybe in a configuration with several subplots?

Adding an example is a good idea - I will keep that in mind and would suggest to add it to ``pyfar.plot.__init__.py`. However, we have to wait until we decided on #214. At the moment complicated examples like the one you suggested are smashed by always using Matplotlib's tight layout.

sikersten (approved):

pyfar/plot/_line.py
f-brinkmann: ok added this :)

PR #225, opened 2021-10-08
Bug fix/frequency data fft norm

### LINKED ISSUE #223: Does FFT normalization make sense for FrequencyData objects?
## General

* pyfar version: all

## Description

Pyfar has FFT normalization and inverse FFT normalization. From a strict theoretic point of view, one could argue that normalization only exist for audio data that can be transformed between the time and frequency domain, hence the name FFT normalization. There are also reasons to argure the same from a practical point of view: All FFT normalizations handle the Nyquist-Frequency differently, and for FrequencyData objects, there is no sampling rate and thus not Nyquist frequency.

One exception to this is the FFT normalization `'none'` - which essentially is no normalization.

Currently FrequencyData objects, can be initiated with any FFT normalization, but they only have a getter function. Only Signal objects have the setter, because setting requires a sampling rate, which FrequencyData objects do not have. This handling is prone to errors, because it would require hidden knowlegde of the *intended* SamplingRate to correctly work with FrequencyData objects with FFT normalizations other than `'none'`.

What should we do about this?

I would suggest that FrequencyData objects are forced to have the `' none'` FFT norm. This way arithmetic operations could be consistently implemented for Signal and FrequencyData objects: #224

f-brinkmann:

Closes #223

- Only allow `'none'` FFT normalization for `FrequencyData` objects
- Update test and documentation
- Change default norm from `None` to `'none'`. This is more explicit and `None` was converted to `'non'` internally anyhow.

### DIFF
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -389,7 +389,7 @@ class FrequencyData(_Audio):
     incomplete spectra.
 
     """
-    def __init__(self, data, frequencies, fft_norm=None, comment=None,
+    def __init__(self, data, frequencies, fft_norm='none', comment=None,
                  dtype=complex):
         """Create FrequencyData with data, and frequencies.
 
@@ -444,8 +444,6 @@ class FrequencyData(_Audio):
                 len(self._frequencies) > 1:
             raise ValueError("Frequencies must be monotonously increasing.")
 
-        if fft_norm is None:
-            fft_norm = 'none'
         if fft_norm in self._VALID_FFT_NORMS:
             self._fft_norm = fft_norm
         else:
@@ -565,7 +563,7 @@ class Signal(FrequencyData, TimeData):
             sampling_rate,
             n_samples=None,
             domain='time',
-            fft_norm=None,
+            fft_norm='none',
             comment=None,
             dtype=np.double):
         """Create Signal with data, and sampling rate.
@@ -620,8 +618,6 @@ class Signal(FrequencyData, TimeData):
             self._data = np.atleast_2d(np.asarray(data, dtype=dtype))
             self._n_samples = self._data.shape[-1]
 
-            if fft_norm is None:
-                fft_norm = 'none'
             if fft_norm in self._VALID_FFT_NORMS:
                 self._fft_norm = fft_norm
             else:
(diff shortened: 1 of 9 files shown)


pyfar/classes/audio.py
```diff
@@ -393,7 +389,7 @@ class FrequencyData(_Audio):
     incomplete spectra.
 
     """
-    def __init__(self, data, frequencies, fft_norm=None, comment=None,
+    def __init__(self, data, frequencies, fft_norm='none', comment=None,
```
sikersten, line 392: I'd suggest the remove the obsolete parameter and move the corresponding explanation in the docstring as `Notes`
```suggestion
    def __init__(self, data, frequencies, comment=None,
```

pyfar/classes/audio.py
```diff
@@ -428,6 +422,16 @@ def __init__(self, data, frequencies, fft_norm=None, comment=None,
 
         _Audio.__init__(self, 'freq', comment, dtype)
 
+        self._VALID_FFT_NORMS = ["none"]
+
+        self._init_frequency_domain_data(data, frequencies, fft_norm)
+
+    def _init_frequency_domain_data(self, data, frequencies, fft_norm):
```
sikersten, line 429: This looks a little artificial to me... What about setting `self._VALID_FFT_NORMS` in the `fft_norm` setter and calling the `fft_norm` setter in the `__init__` functions of FrequencyData and Signal? This would allow for different values of  `self._VALID_FFT_NORMS` in FrequencyData and Signal despite the inheritance.

sikersten (changes requested):

al-de:
This looks good to me. Thanks a lot for implementing.

al-de (approved):

sikersten (commented):
+1 for the new structure with valid fft norms! Would it be possible to remove the `fft_norm ` property for FrequencyData completely?

f-brinkmann:
I would rather keep it to be more explicit. This is somewhat comparable to the `domain` Time and FrequenyData have a only a getter for the domain, Signals do additionally have the setter.

sikersten:
As explained in the Notes for FrequencyData, a `fft_norm` is only well defined for Signals due to the need for a sampling rate etv. From the user perspective, I think it would be fine to explain this in the Notes for FrequencyData and remove the `fft_norm` property. Otherwise, it might be more confusing that the property exists despite the fact that it is not well defined in a mathematical sense?

f-brinkmann:
Agreed and removed :) The property and setter now throw a value error. The documentation says that it will be completely removed in pyfar 0.5.0

sikersten (approved):

PR #230, opened 2021-11-03
import arithmetic functions to pyfar

f-brinkmann:

Arithmetic functions were hard to use and only available from `pyfar.classes.audio.add`. This makes them available as `pyfar.add`

### Changes

- add imports
- use new import syntax in tests


sikersten (approved):
Nice improvement, thanks!

al-de (approved):
Looks good to me, thanks for implementing. :)

PR #233, opened 2021-11-09
allow negative frequencies in LogFormatter

f-brinkmann:

Plot functions label negative frequencies with positive values, e.g., the label of '-100' Hz is '100'.

- Allow negative frequencies in the LogFormatter that is used to format frequency axis

Signals do not need this, but Frequency data could contain negative frequencies in case of analytical spectra.
I locally tested if this changes anything in the default behavior setting `compare = True` in *tests/test_plot.py* and everything was fine. Is there any reason for not allowing negative values?

### DIFF
--- a/pyfar/plot/ticker.py
+++ b/pyfar/plot/ticker.py
@@ -99,7 +99,7 @@ class LogFormatterITAToolbox(LogFormatter):
         if x == 0.0:  # Symlog
             return '0'
 
-        x = abs(x)
+        x = x
 
         vmin, vmax = self.axis.get_view_interval()
         vmin, vmax = mtransforms.nonsingular(vmin, vmax, expander=0.05)


sikersten:
How should the plot look like around f=0 for a spectrum with both negative and positive frequencies?

f-brinkmann:
This would be an example

``` python
import pyfar as pf
import numpy as np

omega = np.linspace(-np.pi, np.pi, 256)
freq = pf.FrequencyData(1 + 2 * np.cos(omega), omega)

ax = pf.plot.freq(freq, xscale="linear")
ax.set_xlim(-np.pi, np.pi)
ax.set_xlabel("Frequency in radians")
```
Screenshot 2021-11-10 121049

sikersten (approved):

al-de (approved):
good idea. :)

PR #234, opened 2021-11-10
Feature/read audio files

### LINKED ISSUE #229: pyfar.io.read_wav does not scale the data correct
## General

* pyfar version: all

I noticed that wav files are not read correctly - the data does not have the correct values. These are the options:

1. Stick with scipy to read wav files. I assume we need to normalize the data in this case as indicated by the table in the scipy documentation: https://docs.scipy.org/doc/scipy/reference/generated/scipy.io.wavfile.read.html
2. Use https://docs.python.org/3.8/library/wave.html which can read/write WAV and AIFF
3. Use https://pysoundfile.readthedocs.io/en/latest/ which can also read/write FLAC but requires libsndfile (has to be manually installed under linux)

sikersten:

Closes #229 

- added functions to read and write audio files based on soundfile
- updated read_wav and write_wav accordingly
- added tests, errors in underlying package are handled

### DIFF
--- a/pyfar/io/io.py
+++ b/pyfar/io/io.py
@@ -14,8 +14,9 @@ import warnings
 import sofar as sf
 import zipfile
 import io
-import soundfile as sf
+import soundfile
 import tempfile
+import numpy as np
 
 import pyfar
 from pyfar import Signal, FrequencyData, Coordinates
@@ -275,7 +276,7 @@ def read_audio(filename, dtype='float64', **kwargs):
       [-1.0, 1.0). If the file contains ``np.array([42.6], dtype='float32')``,
       you will read ``np.array([43], dtype='int32')`` for ``dtype='int32'``.
     """
-    data, sampling_rate = sf.read(
+    data, sampling_rate = soundfile.read(
         file=filename, dtype=dtype, always_2d=True, **kwargs)
     signal = Signal(data.T, sampling_rate, domain='time', dtype=dtype)
     return signal
@@ -339,7 +340,7 @@ def write_audio(signal, filename, subtype=None, overwrite=True, **kwargs):
                 subtype.upper() not in ['FLOAT', 'DOUBLE', 'VORBIS']):
             warnings.warn(
                 f'{format}-files of subtype {subtype} are clipped to +/- 1.')
-        sf.write(
+        soundfile.write(
             file=filename, data=data.T, samplerate=sampling_rate,
             subtype=subtype, **kwargs)
 
@@ -425,7 +426,7 @@ def audio_formats():
      'MAT5': 'MAT5 (GNU Octave 2.1 / Matlab 5.0)'}
 
     """
-    return sf.available_formats()
+    return soundfile.available_formats()
 
 
 def audio_subtypes(format=None):
@@ -449,7 +450,7 @@ def audio_subtypes(format=None):
      'PCM_S8': 'Signed 8 bit PCM'}
 
     """
-    return sf.available_subtypes(format=format)
+    return soundfile.available_subtypes(format=format)
 
 
 def default_audio_subtype(format):
@@ -468,7 +469,7 @@ def default_audio_subtype(format):
     'DOUBLE'
 
     """
-    return sf.default_subtype(format)
+    return soundfile.default_subtype(format)
 
 
 def _clipped_audio_subtypes():
@@ -488,7 +489,7 @@ def _clipped_audio_subtypes():
             for subtype in pyfar.io.audio_subtypes(format):
                 write_valid = not _soundfile_write_errors(format, subtype)
                 read_valid = not _soundfile_read_errors(format, subtype)
-                format_valid = sf.check_format(format, subtype)
+                format_valid = soundfile.check_format(format, subtype)
                 if write_valid and read_valid and format_valid:
                     if format == 'RAW':
                         write_audio(signal, filename, subtype=subtype)
--- a/pyfar/io/io.py
+++ b/pyfar/io/io.py
@@ -8,7 +8,6 @@ imported and exported as WAV files using :py:func:`read_wav` and
 :py:func:`write_wav`. :py:func:`read_sofa` provides functionality to read the
 data stored in a SOFA file.
 """
-import scipy.io.wavfile as wavfile
 import os.path
 import pathlib
 import warnings
@@ -16,6 +15,7 @@ import numpy as np
 import sofa
 import zipfile
 import io
+import soundfile as sf
 
 
 from pyfar import Signal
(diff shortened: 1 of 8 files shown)


sikersten:
Travis fails due to the missing sndfile library, any suggestions what to do?

f-brinkmann:
I think you can add `apt install sndfile` (pseudocode) to tox.ini. Probably for the environments `build` and `wheel`

pyfar/io/io.py
```diff
@@ -297,3 +228,210 @@ def write(filename, compress=False, **objs):
 
     with open(filename, 'wb') as f:
         f.write(zip_buffer.getvalue())
+
+
+def read_audio(filename, dtype='float64', **kwargs):
+    """
+    Import an audio file as :py:class:`~pyfar.classes.audio.Signal` object.
+
```
f-brinkmann, line 236: Maybe a bit more information to be convenient:
```suggestion
    Import an audio file as :py:class:`~pyfar.classes.audio.Signal` object.

    Reads 'wav', 'aiff', 'ogg', and 'flac' files among others. For a complete list see [link to function]
```

pyfar/io/io.py
```diff
@@ -297,3 +228,210 @@ def write(filename, compress=False, **objs):
...
+    Parameters
+    ----------
+    filename : string, Path
+        Input file.
+    dtype : {'float64', 'float32', 'int32', 'int16'}, optional
+        Data type of the returned signal, by default ``'float64'``.
+        Floating point audio data is typically in the range from
+        ``-1.0`` to ``1.0``.  Integer data is in the range from
+        ``-2**15`` to ``2**15-1`` for ``'int16'`` and from ``-2**31`` to
+        ``2**31-1`` for ``'int32'``.
```
f-brinkmann, line 246: I would be more explicit here:

```
        Note that ``'int16'`` and ``'int16'`` should only be used if the data was written in
        the same format. Integer data is in the range from ...
```

pyfar/io/io.py
```diff
@@ -297,3 +228,210 @@ def write(filename, compress=False, **objs):
...
+    signal : Signal
+        :py:class:`~pyfar.classes.audio.Signal` object containing the audio
+        data.
+
+    Notes
+    -----
+    * This function is based on ``soundfile.read``.
+    * Reading int values from a float file will *not* scale the data to
+      [-1.0, 1.0). If the file contains ``np.array([42.6], dtype='float32')``,
+      you will read ``np.array([43], dtype='int32')`` for ``dtype='int32'``.
```
f-brinkmann, line 263: Maybe we can omit this, if we are more explicit above.

pyfar/io/io.py
```diff
@@ -297,3 +228,210 @@ def write(filename, compress=False, **objs):
...
+    data, sampling_rate = sf.read(
+        file=filename, dtype=dtype, always_2d=True, **kwargs)
+    signal = Signal(data.T, sampling_rate, domain='time', dtype=dtype)
+    return signal
+
+
+def write_audio(signal, filename, subtype=None, overwrite=True, **kwargs):
+    """
+    Write a :py:class:`~pyfar.classes.audio.Signal` object as a audio file to
+    disk.
```
f-brinkmann, line 274: Add list of most common file types as above?

pyfar/io/io.py
```diff
@@ -297,3 +228,210 @@ def write(filename, compress=False, **objs):
...
+        The default is ``True``.
+    **kwargs
+        Other keyword arguments to be passed to ``soundfile.write``.
+
+    Notes
+    -----
+    * Signals are flattened before writing to disk (e.g. a signal with
+      ``cshape = (3, 2)`` will be written to disk as a six channel audio file).
+    * This function is based on ``soundfile.write``.
+    * Amplitudes larger than +/- 1 are clipped.
```
f-brinkmann, line 299: This is not always true. Writing wav-files with float or double will not clip. Suggestion - find out when this is the case and raise a warning if data is actually clipped. We could than omit this note.

pyfar/io/io.py
```diff
@@ -297,3 +228,210 @@ def write(filename, compress=False, **objs):
...
+    signal : Signal
+        :py:class:`~pyfar.classes.audio.Signal` object containing the audio
+        data from the WAV file.
+
+    Notes
+    -----
+    * This function is based on :py:func:`read_audio`.
+    """
+    signal = read_audio(filename)
+    return signal
```
f-brinkmann, line 341: Should we add a deprecation warning and tests thereof for pyfar 0.5.0 - a assume we want to get rid of this in the future...

pyfar/io/io.py
```diff
@@ -297,3 +228,210 @@ def write(filename, compress=False, **objs):
...
+    * Signals are flattened before writing to disk (e.g. a signal with
+      ``cshape = (3, 2)`` will be written to disk as a six channel wav file).
+    * This function is based on :py:func:`write_audio`.
+    * Amplitudes larger than +/- 1 are clipped.
+
+    """
+    # .wav file extension
+    filename = pathlib.Path(filename).with_suffix('.wav')
+
+    write_audio(signal, filename, subtype=subtype, overwrite=overwrite)
```
f-brinkmann, line 370: deprecation as well?

pyfar/io/io.py
```diff
@@ -297,3 +228,210 @@ def write(filename, compress=False, **objs):
...
+
+    Returns
+    -------
+    signal : Signal
+        :py:class:`~pyfar.classes.audio.Signal` object containing the audio
+        data.
+
+    Notes
+    -----
+    * This function is based on ``soundfile.read``.
```
f-brinkmann, line 260: Sphinx makes it possible to cross-reference the docs of other packages. Should we do that?

pyfar/io/io.py
```diff
@@ -297,3 +228,210 @@ def write(filename, compress=False, **objs):
...
+    * Amplitudes larger than +/- 1 are clipped.
+
+    """
+    # .wav file extension
+    filename = pathlib.Path(filename).with_suffix('.wav')
+
+    write_audio(signal, filename, subtype=subtype, overwrite=overwrite)
+
+
+def available_formats():
```
f-brinkmann, line 373: should we rename those to:
read_audio_formats
read_audio_subtypes
read_audio_default_subtype

This would make the concept more clear, since we also have other read write functions (soundfile does not)

f-brinkmann (changes requested):
Thanks for the effort - the code looks fine to me (I did not check the tests - I assumed they are identical but adopted for the new methods).

Suggestions are mainly for documentation and naming conventions

pyfar/io/io.py
sikersten: I added the cross-references, but I think we need to adjust the sphinx settings accordingly for it to work. I'd suggest to do that in a separate pull request together with the changes at other positions in the code and in CONTRIBUTING.rst.

pyfar/io/io.py
sikersten: I'd suggest to keep it as an illustrative example

pyfar/io/io.py
sikersten: As this is not straighforward for all implemented formats, I added the function ``write_audio_clipped``.

pyfar/io/io.py
sikersten: yes, good point, added that

pyfar/io/io.py
sikersten: done

pyfar/io/io.py
sikersten: Suggestion: audio_formats, audio_subtypes, default_audio_subtype?

f-brinkmann:
Thanks for the effort in finding out more about the clipping. I slightly changed you function and looked at the output:

f-brinkmann:
Thanks for the effort in finding out more about the clipping. I slightly changed your function and looked at the output to find an easy pattern (code below):

The types FLOAT, DOUBLE, VORBIS do *not* clip to +/- 1, *all* other types clip to +/- 1.

How about the following:
- We raise a warning if the users writes data with abs values > 1 if the subtype is not FLOAT, DOUBLE, or VORBIS
- We keep the modified version of the private function
- We would not need `clipped_audio_subtypes(format)` anymore

``` python
def _clipped_audio_subtypes_dict_creation():
    """Creates a dictionary of format/subtype combinations which are clipped by
    :py:func:´write_audio`.

    This function is not called directly due to the need of writing all files
    to disk. Instead, the dict is manually copied.
    """
    collection = {}
    signal = pyfar.Signal([-1.5, -1, -.5, 0, .5, 1, 1.5]*100, 44100)
    with tempfile.TemporaryDirectory() as tmpdir:
        formats = pyfar.io.audio_formats()
        for format in formats:
            filename = os.path.join(tmpdir, 'test_file.'+format)
            for subtype in pyfar.io.audio_subtypes(format):
                write_valid = not _soundfile_write_errors(format, subtype)
                read_valid = not _soundfile_read_errors(format, subtype)
                format_valid = sf.check_format(format, subtype)
                if write_valid and read_valid and format_valid:
                    if format == 'RAW':
                        write_audio(signal, filename, subtype=subtype)
                        signal_read = read_audio(
                            filename, samplerate=44100, channels=1,
                            subtype=subtype)
                    else:
                        write_audio(signal, filename, subtype=subtype)
                        signal_read = read_audio(filename)
                    if np.any(signal_read.time > 1.1) and np.any(signal_read.time < -1.1):
                        behavior = 'not clipping (' + format + ')'
                    elif np.any(signal_read.time > .1) and np.any(signal_read.time < -.1):
                        behavior = 'clipping to +/- 1 (' + format + ')'
                    else:
                        raise ValueError(f"{format}/{subtype}")

                    if subtype not in collection:
                        collection[subtype] = [behavior]
                    else:
                        collection[subtype] = collection[subtype] + [behavior]

    return collection
```

f-brinkmann (changes requested):
Thanks for the effort - I assume the bug fox in the last commit might have taken some time to figure out. I approve all the current code but have two suggestions:

1. In the top-level readme, we should add the need for `sudo apt-get install libsndfile1` if you are on linux.
2. What error do we get if *libsndfile* is not installed? Can we encapsulate this in a `try`, `catch` statement to give verbose and easy to understand feedback to Linux users who did not install it?

sikersten:
Added the libsndfile information to the README. I think it raises an [`OSError('sndfile library not found')`](https://stackoverflow.com/questions/55086834/cant-import-soundfile-python) when installing SoundFile. This is fine from my point of view.

f-brinkmann (approved):

Issue #224, opened 2021-10-03
Handling of the FFT normalization in arithmetic operations


f-brinkmann:
## General

* pyfar version: all

## Description

I thought about FFT normalization and arithmetic operations.

I see two goals:
1. Mathematically/physically correct operations
2. Uniform and therefore easy to understand handling (best everywhere in pyfar).

And two questions:
1. What do we do with the normalization before the operation?
2. What do we do with the normalization after the operation?

And here are my suggested answers:
1. Before the operation, I think we need to remove the normalization. This is the only consistent solution that gives correct results. If we don't remove the normalization, we would have to do different post-processing depending on the operation (`+`, `-`, `*`, `/`, `**`). e.g. the normalization (`n`) of two signals cancels out during division (`n * H1 / n * H2 = H1 / H2`) but during multiplication it in squares (`n * H1 * n * H2 = n**2 * H1 * H2`). There are more different cases that appear for different operations and if only one of the data has an FFT normalization. Currently we do remove the normalization before the operation (upon first glance). However, we handle it differently for Signal and FrequencyData objects. I think the reason for this is an error of thought in the FrequencyData implementation: #223
2. After the operation we can choose. Then the question is rather what is the most intuitive / explicit solution. Right now the result has the norm of the first audio object. Alternatively, we could return the result without normalization. I find the current variant more intuitive.

mberz:
I think it makes sense here to only support physically meaningful operations. Yes, I'm going to talk abount energy and power signals again ;)

The norms and their corresponding signal type:

energy signal:
- 'none'

power signal:
- 'unitary'
- 'rms'
- 'amplitude'

power spectrum (s(t)^2):
- 'power'
- 'psd'

The resulting signal types per operation are then:

energy * energy = energy
energy / energy = energy
power * energy = power (i.e. white noise fed into a band-pass filter will result in band-pass filtered noise)
power * power = power spectrum (i.e. white noise spectrum multiplied with itself will be the power spectrum)
power / power = energy (i.e. IR measurement.)

power * power * power = ? Doesn't really have a physical meaning to me.

- Altogether this would result in following the norm of the power signal if multiplying an energy signal with a power signal. 
- Returning a power spectrum if multiplying two power signals. (De-normalize -> Set to 'power' norm)
- Raising an error if power * power spectrum
- Returning a 'none' normalized signal if dividing two power type signals.

sikersten:
[...]

PR #235, opened 2021-11-10
Feature/match fft norm

al-de:

- add the function `match_fft_norm` which takes two fft_norms and returns one fft_norm resulting from arithmetic operations
  - raises errors, when invalid fft_norm combinations are passed
- basic testing of the function `match_fft_norm`

This pull request followed the discussion in #224

`match_fft_norm(fft_norm_1, fft_norm_2, division=False)`

Parameters:
`fft_norm_1` : str, first fft_norm for matching.
`fft_norm_2` : str, second fft_norm for matching.
`division` : bool, `False` if arithmetic operation is addition, subtraction or multiplication; `True` if arithmetic operation is division.

Returns:
`fft_norm_result` : str, the fft_norm resulting from arithmetic operation.

### DIFF
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -1286,9 +1286,9 @@ def match_fft_norm(fft_norm_1, fft_norm_2, division=False):
 
     # check if fft_norms are type string
     if isinstance(fft_norm_1, str) is False:
-        raise ValueError("Parameter fft_norm_1 must be type str.")
+        raise TypeError("Parameter fft_norm_1 must be type str.")
     if isinstance(fft_norm_2, str) is False:
-        raise ValueError("Parameter fft_norm_2 must be type str.")
+        raise TypeError("Parameter fft_norm_2 must be type str.")
 
     # check if fft_norms are valid
     valid_fft_norms = ['none', 'unitary', 'amplitude', 'rms', 'power', 'psd']
@@ -1303,7 +1303,7 @@ def match_fft_norm(fft_norm_1, fft_norm_2, division=False):
 
     # check if parameter division is type bool
     if isinstance(division, bool) is False:
-        raise ValueError("Parameter division must be type bool.")
+        raise TypeError("Parameter division must be type bool.")
 
     if division is False:
 
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -1317,7 +1317,7 @@ def _match_fft_norm(fft_norm_1, fft_norm_2, division=False):
                              "fft_norms must be the same, but they are ",
                              f"{fft_norm_1} and {fft_norm_2}.")
 
-    elif division is True:
+    else:
 
         if fft_norm_2 == 'none':
             fft_norm_result = fft_norm_1
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -1249,3 +1249,81 @@ def _divide(a, b):
 
 def _power(a, b):
     return a**b
+
+
+def match_fft_norm(fft_norm_1, fft_norm_2, division=False):
+    """
+    Helper function to determine the fft_norm resulting from an
+    arithmetic operation of two signals.
+
+    For addition, subtraction and multiplication:
+    Either: one signal has fft_norm ``'none'`` , the results gets the other
+    norm.
+    Or: both have the same fft_norm. Other combinations raise an error.
+
+    For division:
+    ###  more discussion needed  ###
+
+    Parameters
+    ----------
+    fft_norm_1 : str, ``'none'``, ``'unitary'``, ``'amplitude'``, ``'rms'``,
+    ``'power'`` or ``'psd'``
+        First fft_norm for matching.
+    fft_norm_2 : str, ``'none'``, ``'unitary'``, ``'amplitude'``, ``'rms'``,
+    ``'power'`` or ``'psd'``
+        Second fft_norm for matching.
+    division : bool
+        ``True`` if arithmetic operation is addition, subtraction or
+        multiplication;
+        ``False`` if arithmetic operation is division.
+
+    Returns
+    -------
+    fft_norm_result : str, ``'none'``, ``'unitary'``, ``'amplitude'``,
+    ``'rms'``, ``'power'`` or ``'psd'``
+        The fft_norm resulting from arithmetic operation.
+    """
+
+    # check if fft_norms are type string
+    if isinstance(fft_norm_1, str) is False:
+        raise ValueError("Parameter fft_norm_1 must be type str.")
+    if isinstance(fft_norm_2, str) is False:
+        raise ValueError("Parameter fft_norm_2 must be type str.")
+
+    # check if fft_norms are valid
+    valid_fft_norms = ['none', 'unitary', 'amplitude', 'rms', 'power', 'psd']
+    if fft_norm_1 not in valid_fft_norms:
+        raise ValueError("Parameter fft_norm_1 is not a valid fft_norm.\n" +
+                         f"valid fft_norms: {valid_fft_norms}\n" +
+                         f"found: {fft_norm_1}")
+    if fft_norm_2 not in valid_fft_norms:
+        raise ValueError("Parameter fft_norm_2 is not a valid fft_norm.\n" +
+                         f"valid fft_norms: {valid_fft_norms}\n" +
+                         f"found: {fft_norm_2}")
+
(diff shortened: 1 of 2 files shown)


pyfar/classes/audio.py
```diff
@@ -1249,3 +1249,92 @@ def _divide(a, b):
...
+    fft_norm_result : str, ``'none'``, ``'unitary'``, ``'amplitude'``,
+    ``'rms'``, ``'power'`` or ``'psd'``
+        The fft_norm resulting from arithmetic operation.
+    """
+
+    # check if fft_norms are type string
+    if isinstance(fft_norm_1, str) is False:
+        raise TypeError("Parameter fft_norm_1 must be type str.")
+    if isinstance(fft_norm_2, str) is False:
+        raise TypeError("Parameter fft_norm_2 must be type str.")
```
f-brinkmann, line 1294: You can omit this, because of the next check. If it is not a string the next will already raise a ValueError which would be fine for me.

pyfar/classes/audio.py
```diff
@@ -1249,3 +1249,92 @@ def _divide(a, b):
...
+
+    # check if fft_norms are valid
+    valid_fft_norms = ['none', 'unitary', 'amplitude', 'rms', 'power', 'psd']
+    if fft_norm_1 not in valid_fft_norms:
+        raise ValueError("Parameter fft_norm_1 is not a valid fft_norm.\n" +
+                         f"valid fft_norms: {valid_fft_norms}\n" +
+                         f"found: {fft_norm_1}")
```
f-brinkmann, line 1301: We usually do not raise multi-line errors. How about making the string shorter:

`f"fft_norm_1 is {fft_norm_1} but must be in {', '.join(valid_fft_norms)}"`

pyfar/classes/audio.py
```diff
@@ -1249,3 +1249,92 @@ def _divide(a, b):
 
 def _power(a, b):
     return a**b
+
+
+def match_fft_norm(fft_norm_1, fft_norm_2, division=False):
```
f-brinkmann, line 1254: Add `_` to clearly mark it as a private function
```suggestion
def _match_fft_norm(fft_norm_1, fft_norm_2, division=False):
```

f-brinkmann (changes requested):
Thanks for the effort. The code looks good for me. Only had some comments on code style.

pyfar/classes/audio.py
al-de: check, I wanted the raised errors to be as explicit as possible, but are absolutely fine with your suggestion.
I will implement it in my next commit.

pyfar/classes/audio.py
al-de: check

pyfar/classes/audio.py
al-de: check.

al-de:
Thanks a lot for your feedback, I implemented all your suggestions with my latest commit.

tests/test_match_fft_norm.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def test_input_fft_norm_type():
+    """Test assertion by passig non str type"""
+    with pytest.raises(TypeError,
+                       match='Parameter fft_norm_1 must be type str.'):
+        match_fft_norm(0, 'none')
+    with pytest.raises(TypeError,
+                       match='Parameter fft_norm_2 must be type str.'):
+        match_fft_norm('none', 0)
```
f-brinkmann, line 12: Is not needed anymore :)

f-brinkmann (approved):
Approved after tests are fixed :)

tests/test_match_fft_norm.py
al-de: You are totally right, I just forgot to look at the file containing the tests. Thaks for reminding me, I fixed this with my last commit.

One Question: It seems like travis is checking every commit I push, despite this is only a draft pull request. Is this behaviour what we want, as our monthly time on travis is limited?

al-de:
I finalized the function and testing with the results of our discussion today as stated in #224 and changed this pull request to non-draft.

pyfar/classes/audio.py
```diff
@@ -1249,3 +1249,85 @@ def _divide(a, b):
 
 def _power(a, b):
     return a**b
+
+
+def _match_fft_norm(fft_norm_1, fft_norm_2, division=False):
+    """
+    Helper function to determine the fft_norm resulting from an
+    arithmetic operation of two signals.
```
f-brinkmann, line 1257: Also works for Time/FrequencyData :)
```suggestion
    arithmetic operation of two audio objects.
```

pyfar/classes/audio.py
```diff
@@ -1249,3 +1249,85 @@ def _divide(a, b):
...
+    valid_fft_norms = ['none', 'unitary', 'amplitude', 'rms', 'power', 'psd']
+    if fft_norm_1 not in valid_fft_norms:
+        raise ValueError(f"fft_norm_1 is {fft_norm_1} but must be in ",
+                         f"{', '.join(valid_fft_norms)}")
+    if fft_norm_2 not in valid_fft_norms:
+        raise ValueError(f"fft_norm_2 is {fft_norm_2} but must be in ",
+                         f"{', '.join(valid_fft_norms)}")
+
+    # check if parameter division is type bool
+    if isinstance(division, bool) is False:
```
f-brinkmann, line 1301: more common I think:
```suggestion
    if not isinstance(division, bool):
```

pyfar/classes/audio.py
```diff
@@ -1249,3 +1249,85 @@ def _divide(a, b):
...
+        raise TypeError("Parameter division must be type bool.")
+
+    if division is False:
```
f-brinkmann, line 1304: more common / easier to read :)
```suggestion
    if not division:
```

pyfar/classes/audio.py
```diff
@@ -1249,3 +1249,85 @@ def _divide(a, b):
...
+
+        if fft_norm_1 == fft_norm_2:
+            fft_norm_result = fft_norm_1
+
+        elif fft_norm_1 == 'none':
+            fft_norm_result = fft_norm_2
+
+        elif fft_norm_2 == 'none':
+            fft_norm_result = fft_norm_1
```
f-brinkmann, line 1313: shortcut, if you want:
```suggestion
        else:
            fft_norm_result = fft_norm_2 if fft_norm_1 == 'none' else fft_norm_1
```

f-brinkmann (approved):

pyfar/classes/audio.py
al-de: totally right, thx. :)

pyfar/classes/audio.py
al-de: check

pyfar/classes/audio.py
al-de: check

pyfar/classes/audio.py
al-de: If we do this, we can't use the `else` for raising the ValueError when an invalid combination of norms is passed. Do you have an idea, how we could solve this?

f-brinkmann: Oh, true. Didn't think that through. Let's leave it in this case...

pyfar/classes/audio.py
```diff
@@ -1249,3 +1249,85 @@ def _divide(a, b):
...
+
+        else:
+            raise ValueError("Either one fft_norm has to be 'none' or both ",
+                             "fft_norms must be the same, but they are ",
+                             f"{fft_norm_1} and {fft_norm_2}.")
+
+    elif division is True:
```
f-brinkmann, line 1320: because of the check above, you know that it is a bool
```suggestion
    else:
```

pyfar/classes/audio.py
al-de: check

pyfar/classes/audio.py
al-de: that's true. :)

sikersten (commented):
Functionality looks good, thanks! I'd suggest to replace the loops inside the tests with pytest's [parametrize functionality](https://docs.pytest.org/en/6.2.x/parametrize.html) :)

al-de:
Thanks for this hint. I implemented the tests with pytest's parametrize functionality now. :)

sikersten (approved):

PR #247, opened 2021-12-07
Add return_window parameter to time_window

### LINKED ISSUE #242: Possibility to return time window
## General

* pyfar version: 0.2.3 

## Description

Should we make it possible for `pyfar.dsp.time_window` to return the window itself? This would be equivalent to pass a signal containing only ones. This would be more convenient if applying the same window at multiple points or to multiple signals.

sikersten:

Closes #242 

- A ``return_ window`` parameter is added to `pyfar.dsp.time_window` to optionally inspect to time window. This is easier than calling the function with ones.

### DIFF
--- a/pyfar/dsp/dsp.py
+++ b/pyfar/dsp/dsp.py
@@ -320,7 +320,7 @@ def spectrogram(signal, dB=True, log_prefix=20, log_reference=1,
 
 
 def time_window(signal, interval, window='hann', shape='symmetric',
-                unit='samples', crop='none'):
+                unit='samples', crop='none', return_window=False):
     """Apply time window to signal.
 
     This function uses the windows implemented in ``scipy.signal.windows``.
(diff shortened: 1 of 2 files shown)


pyfar/dsp/dsp.py
```diff
@@ -370,11 +370,16 @@ def time_window(signal, interval, window='hann', shape='symmetric',
             cropped, so the original phase is preserved.
 
         The default is ``'none'``.
+    return_window: bool, optional
+        If ``True``, both the windowed signal and the time window are returned.
```
f-brinkmann, line 374:
```suggestion
        If ``True``, both the windowed signal and the time window are returned.
        The default is ``False``.
```

f-brinkmann (approved):
Thanks for the work. Just had one minor comment on the docstring.

al-de (approved):
looks good to me, thanks for implementing.

JMA-JMA (approved):
Works. Thanks! :)

PR #248, opened 2021-12-08
More verbose filternames

### LINKED ISSUE #170: More verbose filter names
e.g., butterworth instead of butter

f-brinkmann:

Closes #170 

- rename filter functions to more verbose names
- add deprecation warnings and tests for old  filter functions
- improve documentation
- add plot showing all filters to `docs/resources` for later use in *Concepts*


pyfar/dsp/filter.py
```diff
@@ -353,10 +420,10 @@ def bessel(signal, N, frequency, btype='lowpass', norm='phase',
         return signal_filt
 
 
-def peq(signal, center_frequency, gain, quality, peq_type='II',
-        quality_warp='cos', sampling_rate=None):
+def bell(signal, center_frequency, gain, quality, bell_type='II',
```
sikersten, line 423: I'd suggest a different name:
```suggestion
def parametric_eq(signal, center_frequency, gain, quality, bell_type='II',
```

sikersten (commented):
Thanks for changing, looks good! I'd only suggest a different name for `bell`

pyfar/dsp/filter.py
f-brinkmann: That would be fine for me. Let's wait for more votes and finalize the name during the next weekly.

al-de: I'm not sure which name I'd prefer, as I think `parametric_eq` is more intuitive, but `bell` is more precise as one could see a high- and low-shelf as a kind of parametric eq, too. When forced to decide I'd probably go with `parametric_eq`.

al-de (approved):
Thanks a lot for implementing, I especially like the new overview plot for the filter types.

pyfar/dsp/filter.py
f-brinkmann: Following the comment from @al-de, I googled 'peq filter' to see how the term is generally used. Lots of the prominent results did not clearly distinguish between peq/bell and shelve filters and use peq as the general term for both (e.g. https://de.mathworks.com/help/audio/ug/parametric-equalizer-design.html).

Googling 'bell filter' shows that this term seems to be less ambiguous but maybe slightly less familiar. Other terms that you find in this search are 'bell eq' and 'peaking filter'.

I would still vote for bell, to avoid mixing peq/bell and shelve filters. It might also be more intuitive for a non-audio peer group, because bell well describes the filter shape.

PR #252, opened 2021-12-17
Feature/filter class coefficient setter

### LINKED ISSUE #249: Coefficients setter for filter classes
## General

* pyfar version: 0.2.3

## Description

The filter classes have a getter for `coefficients` but a setter is missing. Having a setter would make it possible to mimic time invariant systems by a block-wise update of the coefficients.

f-brinkmann:

Closes #249 

- add setter for `coefficients`.  Did not require new tests, because the setter is tested as part of the initialization.
- add `comment` property to all filter classes and new default for comment (`'None'` instead of `None`, which is more consistent with audio classes).

### DIFF
--- a/pyfar/classes/filter.py
+++ b/pyfar/classes/filter.py
@@ -122,6 +122,8 @@ class Filter(object):
             The sampling rate of the filter in Hz.
         state : array, double, optional
             The state of the buffer elements.
+        comment : str, optional
+            A comment. The default is ``'None'``.
 
         Returns
         -------
@@ -146,7 +148,7 @@ class Filter(object):
 
         self._state = state
         self._sampling_rate = sampling_rate
-        self._comment = comment
+        self.comment = comment
 
     def init_state(self, state='zeros'):
         self._state = state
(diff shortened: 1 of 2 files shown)


al-de (approved):
Thanks for implementing, this look good to me.

mberz:
See my comment in issue #249. I don't think having a setter like this is the way to go here. Let's discuss this further.

sikersten (commented):
Looks good, thanks! I'd generally approve, only two minor comments:

- Do I get it right, that the `coefficients` setter is only implicitly tested during the initialization and the use case of time variant filtering is more or less tested in filter_concepts.py?
- Maybe a little explanation of the intended use case in the documentation of the `coefficients` setter itself would be nice to hint to the problem mentioned by @mberz in #249?

f-brinkmann:
- Blockwise processing was tested for all filter calsses in `test_filter_classes.py::test_blockwise_processing`. I now added a test with exchanging the filter coefficients for the FIR filter `test_blockwise_processing_with_coefficients_exchange`. Together with the implicit tests of the setter functions this should cover all use cases.
- I added a link from  `Filter*.coefficients` to the Filter concepts.

Issue #244, opened 2021-12-03
Update handling of FFT normalization in pyfar


f-brinkmann:
## General

* pyfar version: 0.2.3

- [x] Concepts: arithmetic (include, after #243 ),
- [x] FFT normalization (link to arithmetic, after #243 )
- [x] regulated inversion
- [x] arithmetics

f-brinkmann:
I tried not to allow 1 / power-signal in the regulated inversion, however, there is a problem. The regulated inversion is used as part of the deconvolution where we can pass to power signals (fine with our rules for arithmetics). The second power signal is then passed to the regulated inversion, which would throw a value error, which we do not want. I thus would suggest to keep the regulated inversion as an exception to the 1 / power-signal rule.

PR #253, opened 2021-12-17
Complete rules for arithmetics

### LINKED ISSUE #224: Handling of the FFT normalization in arithmetic operations

f-brinkmann:

Closes #224 contributes to #244 

- check `fft_norm` in cases like `1 / signal`
- additional tests not required. Case was already tested, but now the behavior is different, e.g., `1 / pf.signals.sine(100, 100)` is not possible anymore

### DIFF
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -1075,21 +1075,25 @@ def _assert_match_for_arithmetic(data: tuple, domain: str, division: bool):
     # properties that must match
     sampling_rate = None
     n_samples = None
-    fft_norm = None
+    fft_norm = 'none'
     times = None
     frequencies = None
     audio_type = type(None)
 
     # check input types and meta data
     found_audio_data = False
-    for d in data:
+    for n, d in enumerate(data):
         if isinstance(d, (Signal, TimeData, FrequencyData)):
             # store meta data upon first appearance
             if not found_audio_data:
                 if isinstance(d, Signal):
                     sampling_rate = d.sampling_rate
                     n_samples = d.n_samples
-                    fft_norm = d.fft_norm
+                    # if a signal comes first (n==0) its fft_norm is taken
+                    # directly. If a signal does not come first, (n>0, e.g.
+                    # 1/signal, the fft norm is matched)
+                    fft_norm = d.fft_norm if n == 0 else \
+                        _match_fft_norm(fft_norm, d.fft_norm, division)
                 elif isinstance(d, TimeData):
                     if domain != "time":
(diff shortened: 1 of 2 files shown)


al-de (approved):
Looks good, thanks for implementing. :)

pyfar/classes/audio.py
```diff
@@ -1075,21 +1075,25 @@ def _assert_match_for_arithmetic(data: tuple, domain: str, division: bool):
...
         if isinstance(d, (Signal, TimeData, FrequencyData)):
             # store meta data upon first appearance
             if not found_audio_data:
                 if isinstance(d, Signal):
                     sampling_rate = d.sampling_rate
                     n_samples = d.n_samples
-                    fft_norm = d.fft_norm
+                    # if a signal comes first (n==0) its fft_norm is taken
+                    # directly. If a signal does not come first, (n>0, e.g.
+                    # 1/signal, the fft norm is matched)
```
JMA-JMA, line 1094:
```suggestion
                    # 1/signal), the fft norm is matched
```

JMA-JMA (approved):
Ok. Just small changes in comments...

sikersten (approved):
Looks good, thanks!
Which tests do you refer to?

f-brinkmann:
> Looks good, thanks! Which tests do you refer to?

I remember I had to change a test after changing the behavior. But now I did not find the test myself and quickly added an explicit test for this :)

PR #260, opened 2022-01-17
Enhancement/log prefix defaults

### LINKED ISSUE #259: Default dB Parameter in frequency domain plots
## General

* pyfar version: 0.2.3

The FFT normalization 'PSD', 'Power' require `10 * log10(signal.freq)` while all other normalizations require  `20 * log10(signal.freq)`. We should change the default for these cases.

sikersten:

Closes #259 

- changed default behavior of spectrum plots for signals with `fft_norm`  ``'power'`` and ``'psd'`` to ``10 * log10(signal.freq)`` instead of ``20 * log10(signal.freq)``
- added `log_prefix_time` and `log_prefix_freq` parameters to `plot.time_freq` instead of a single `log_prefix`

### DIFF
--- a/pyfar/plot/_utils.py
+++ b/pyfar/plot/_utils.py
@@ -279,7 +279,7 @@ def _deal_time_units(unit='s'):
     return factor, string
 
 
-def _log_prefix(fft_norm):
+def _log_prefix(signal):
     """Return prefix for dB calculation in frequency domain depending on
     fft_norm.
 
@@ -291,7 +291,7 @@ def _log_prefix(fft_norm):
     fft_norm : str
         FFT normalization
     """
-    if fft_norm in ('power', 'psd'):
+    if isinstance(signal, Signal) and signal.fft_norm in ('power', 'psd'):
         log_prefix = 10
     else:
         log_prefix = 20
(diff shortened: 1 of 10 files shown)


pyfar/plot/_utils.py
```diff
@@ -277,3 +277,22 @@ def _deal_time_units(unit='s'):
         factor = 1
         string = ''
     return factor, string
+
+
+def _log_prefix(fft_norm):
```
f-brinkmann, line 282: Would it make sense to also move this code inside this function
```python
if log_prefix is None:
    if isinstance(signal, Signal):
        log_prefix = _utils._log_prefix(signal.fft_norm)
    else:
        log_prefix = 20
```
Otherwise we would have to repeat it in every freq-like plot function

f-brinkmann (changes requested):
Looks good - but I think the new default `None` is missing in `plot.two_2.spectrogram` and `plot._two_2.spectrogram`

pyfar/plot/_utils.py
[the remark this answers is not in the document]
sikersten: Yes, added that

sikersten: I changed the function to be called with the signal and kept the None check outside. It could be included in a general dB calculation function?

f-brinkmann (approved):

al-de (approved):
Thanks a lot for implementing, this looks good to me.

PR #263, opened 2022-01-25
Bugfix/micro seconds label in plots

sikersten:

Plotting with time unit µs as axis label (e.g., in time, group_delay) raises an "Unknown symbol" error:

```
s = pf.signals.impulse(10, sampling_rate=44100)
pf.plot.time(s)
```

- exchanged  ``r'$\mathrm{\mu s}$'`` with unicode ``'µs'`` in private function ``_deal_time_units``
- added test

### DIFF
--- a/pyfar/plot/_utils.py
+++ b/pyfar/plot/_utils.py
@@ -269,7 +269,7 @@ def _deal_time_units(unit='s'):
         string = 'ms'
     elif unit == 'mus':
         factor = 1 / 1e-6
-        string = 'µs'
+        string = r'$\mathrm{\mu}$s'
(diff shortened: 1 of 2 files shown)


f-brinkmann:
Both versions work for me, i.e., I do not get the error. I always thought using Tex would be the safer way. Should we look a bit more into detail why the error is thrown?

sikersten:
Thanks for noting. I realized that the error is thrown only with the space between `\mu` and `s` inside the `\mathrm` command, so I moved the `s` outside the `$...$` and it works :)

f-brinkmann (approved):

PR #274, opened 2022-02-07
Refactor Signal.freq

### LINKED ISSUE #262: Handling of FFT-Normalizations in pyfar
## General

* pyfar version: all

## Description

FFT normalization is implemented in `pyfar.fft.normalization` and `Signal.freq` always returns the normalized spectrum (probably inspired by the ITA-Toolbox for Matlab). During coding pyfar, we came across a couple of places where this might not be the best solution, and where we have to

- remove the normalization,
- do some processing, and
- add the normalization again.

This is not only additional code, but also makes coding pyfar prone to errors. So far, the only place, where we directly used the normalized spectrum were the plot functions. This made me think of two questions

1. Do we want to change the current handling?
2. If yes, do we want to break backwards compatibility and move to pyfar 1.0?

**If the answer to 1. is yes, I would propose the following**
Add a property and setter to `Signal`, e.g., `Signal.freq_raw`. Internally we would only store the raw, i.e., not normalized frequency data. The old property `Signal.freq` would return the normalized data. The old setter `Signal.freq` would apply the inverse normalization to the input data and save it.

**If the answer to 2. is yes, I would propose the following**
The property `Signal.freq` returns the non-normalized data. This seems to be the most common case and more intuitive - at least for me. We add a property, e.g., `Signal.freq_norm` that behaves as `Signal.freq` described in the previous paragraph.

**Third option**
Add `freq_raw` and `freq_norm` and keep `freq` for two minor versions with a deprecation warning. In this case `freq_norm` would only have a getter, and return real values spectra in some cases.

What are your thoughts on this?

sikersten:

Closes #262 

- the Signal property `freq_raw` is introduced, which contains the unnormalized spectrum 
- `freq_raw` is used to replace internal denormalizations of `freq`
- the `freq` property inherited from FrequencyData is overwritten accordingly


pyfar/classes/audio.py
```diff
@@ -669,16 +656,18 @@ def domain(self, new_domain):
             # Only process if we change domain
```
f-brinkmann, line 656: Lets simplify the line above to `if self._domain != new_domain`

pyfar/classes/audio.py
```diff
@@ -726,20 +715,14 @@ def fft_norm(self, value):
...
-                self._data, self._n_samples, self._sampling_rate,
-                self._fft_norm, inverse=True)
-            # normalize
-            self._data = fft.normalization(
-                self._data, self._n_samples, self._sampling_rate,
-                value, inverse=False)
-
         self._fft_norm = value
 
+    @property
```
f-brinkmann, line 720: Suggestion: Lets move this up right below the `freq`property to make the concept more clear is someone sees the code for the first time

Question: Don't we need a setter for `freq_raw`? I thought it would be the easiest way if we change `freq_raw`, e.g., in the DSP module to also put it back in to the Signal object using a corresponding setter.

pyfar/classes/audio.py
```diff
@@ -637,27 +624,27 @@ def time(self):
 
     @FrequencyData.freq.getter
     def freq(self):
-        """Return the data in the frequency domain."""
-        # converts the data from 'time' to 'freq'
-        self.domain = 'freq'
-
-        return super().freq
+        """Return the (normalized) frequency domain data."""
```
f-brinkmann, line 627: I would suggest to remove the parentheses around 'normalized' and to add a second line to the Docstring, for example
'The normalized data is usually used for inspecting the data, e.g., using plots or when extracting information such as the amplitude of harmonic components. Most processing operations, e.g., frequency domain convolution require the non-normalized data.'

The opposite could be added to the `freq_raw` docstring

f-brinkmann (changes requested):
Thanks for the effort, this looks good to me. Note that we added functionality but ended up with less code for the audio classes, which makes me think this really was a good idea. I was wondering why there is no setter for `freq_raw` (see detailed comments)

pyfar/classes/audio.py
sikersten: changed it :)

pyfar/classes/audio.py
sikersten: Yes, good point. Changed the docstrings accordingly :)

pyfar/classes/audio.py
sikersten: Good point, moved it.
I added a setter for `freq_raw` and removed the normalization in __arithmetic_.
Which positions in the code do you refer to in the dsp module?

f-brinkmann: I was not referring to any specific position in DSP. I thought it was the next step, to check where we should work with `freq_raw` in follow up Pull requests.

f-brinkmann (approved):

tests/test_signal.py
```diff
@@ -78,15 +75,12 @@ def test_domain_setter_time_when_time():
     assert signal.domain == domain
 
 
-def test_domain_setter_time_when_freq(sine_stub):
-    signal = Signal(
-        sine_stub.freq, sine_stub.sampling_rate, domain='freq',
-        fft_norm=sine_stub.fft_norm)
+def test_domain_setter_time_when_freq():
+    signal = Signal([1, 2, 3, 4], 44100, domain='time', fft_norm='rms')
```
al-de, line 79:
```suggestion
    signal = Signal([1, 2, 3, 4], 44100, domain='freq', fft_norm='rms')
```
If I get it right, the domain of the Signal defined for this test should have the 'freq' domain in order to test the setter below.

al-de (approved):
Thanks a lot for your effort adding signal.freq_raw and the tests. The code look good to me, only had one minor comment. Besides this approved.

examples/pyfar_demo.ipynb
```diff
@@ -174,8 +175,7 @@
    "outputs": [],
    "source": [
     "time_data = x_power.time\n",
-    "freq_data_normalized = x_power.freq\n",
-    "freq_data_raw = x_power.freq_raw"
+    "freq_data = x_power.freq"
```
f-brinkmann, line 178: It looks as if, we lost changes here red lines 177 and 178 are new and green line 178 is outdated

examples/pyfar_demo.ipynb
```diff
@@ -184,9 +184,9 @@
    "source": [
     "Two things are important here:\n",
     "\n",
-    "1. The data are mutable! That means `x_power.time` changes if you change `time_data`. If this is not what you want use `time_data = x_power.time.copy()` instead.\n",
+    "1. `time_data` and `freq_data` are mutable! That means `x_power.time` changes if you change `time_data`. If this is not what you want use `time_data = x_power.time.copy()` instead.\n",
     "\n",
-    "2. The frequency data of signals is available without normalization from `x_power.freq_raw` and with normalization from `x_power.freq`. The latter depends on the Signal's `fft_norm`."
+    "2. The frequency data of signals depends on the Signal's `fft_norm`."
```
f-brinkmann, line 189: Same here - I think the red lines are the latest

f-brinkmann (changes requested):
The code looks good - thanks for implementing. I spotted to parts in the examples notebook where I think that something went wrong.

examples/pyfar_demo.ipynb
sikersten: thanks for noting, changed it

f-brinkmann (approved):

PR #275, opened 2022-02-08
Bugfix/remove C._t file from tests

sikersten:

For some audio formats a separate hidden file called `C._t` is created when writing/reading files (see Q13 at https://libsndfile.github.io/libsndfile/FAQ.html). The corresponding tests are extended to remove the file, so is not recognized as new file by Git every time the tests are run.


f-brinkmann (approved):
Thanks for figuring out. I was wondering about that for a while...

PR #276, opened 2022-02-10
Feature/3d remaining plots

f-brinkmann:

Add remaining 2D Plots

- Add `freq2d`, `phase2d`, `group_delay2d`, `time_freq2d`, `freq_phase2d`, `freq_groupd_delay2d`
- Change custom shading from `gouraud` to `nearest`. The old default seemed a bit deceiving for me, as it might make data look more smooth / high res. than they actually are (https://matplotlib.org/devdocs/gallery/images_contours_and_fields/pcolormesh_grids.html)
- Add new functionality to the interaction module to account for correct handling of axis operations (move, zoom, log/lin) after changing the orientation of 2D plots
- separate handling of `kwargs` (keyword arguments) for different plotfunctions (`matplotlib.pyplot.plot`, `matplotlib.pyplot.pcolormesh`)
- attach all axes and colorbars to interaction instance to improve testing

### Review process

- inspect all baseline plots (consider using the GitHub Desktop App)
- test interaction excessively
- build & check docs and examples


pyfar/plot/two_d.py
```diff
@@ -85,22 +118,747 @@ def time2d(signal, dB=False, log_prefix=None, log_reference=1, unit=None,
...
+
+        >>> import pyfar as pf
+        >>> import numpy as np
+        >>> # generate the signal
+        >>> angles = np.arange(0, 180) / 180 * np.pi
+        >>> delays = np.round(100 * np.sin(angles)).astype(int)
+        >>> amplitudes = .5 + .5 * np.abs(np.cos(angles))
+        >>> signal = pf.signals.impulse(128, delays, amplitudes)
+        >>> # plot the signal
+        >>> pf.plot.time2d(signal)
```
sikersten, line 587:
```suggestion
        >>> pf.plot.time_freq2d(signal)
```

pyfar/plot/two_d.py
```diff
@@ -85,22 +118,747 @@ def time2d(signal, dB=False, log_prefix=None, log_reference=1, unit=None,
...
```
sikersten, line 711:
```suggestion
        >>> pf.plot.freq_phase2d(signal)
```

pyfar/plot/two_d.py
```diff
@@ -85,22 +118,747 @@ def time2d(signal, dB=False, log_prefix=None, log_reference=1, unit=None,
...
```
sikersten, line 834:
```suggestion
        >>> pf.plot.freq_group_delay2d(signal)
```

sikersten (changes requested):
Many thanks for implementing! Plots look very nice and interaction works smoothly :)
Some general comments we can discuss:

- I'm not completely happy with the default name _Points_ for the second axis. What about something like _Index_? For specific values, one needs to add a specific label anyway.
- I'd suggest to change the functions' suffix to `_2d` instead of `2d`. In my opinion, it matches the underscore convention (e.g., in `freq_phase`) better by clearly separating it from the last phrase.
- In my case the frequency ticker only showed 1k and 10k when plotting the examples from the functions. I'd suggest to have the same as for `freq`. The difference is nicely illustrated when toggling between `freq2d` and `freq` :)

f-brinkmann:
@sikersten : all suggestions implemented :)

sikersten (approved):

PR #277, opened 2022-02-10
Feature/matrix multiplication

sikersten:

- add matrix multiplication along first two axes to arithmetic operations for audio classes
- enable the `@` operator accordingly


pyfar/classes/audio.py
```diff
@@ -1011,6 +1025,41 @@ def power(data: tuple, domain='freq'):
...
+    data : tuple of the form (data_1, data_2, ..., data_N)
+        Data to be multiplied. Can contain pyfar audio objects, array likes,
+        and scalars. Pyfar audio objects can not be mixed, e.g.,
+        :py:func:`TimeData` and :py:func:`FrequencyData` objects do not work
+        together. See
+        :py:mod:`arithmetic operations <pyfar._concepts.arithmetic_operations>`
+        for possible combinations of Signal FFT normalizations.
+    domain : ``'time'``, ``'freq'``, optional
+        Flag to indicate if the operation should be performed in the time or
+        frequency domain. If working in the frequency domain, the FFT
```
f-brinkmann, line 1047: Should we already change this to 'Frequency domain operations work on the raw spectrum' and also change this in the other docstrings?

pyfar/classes/audio.py
```diff
@@ -1011,6 +1025,41 @@ def power(data: tuple, domain='freq'):
     return _arithmetic(data, domain, _power)
 
 
+def matrix_multiplication(data: tuple, domain='freq'):
+    """Matrix multiplication along first two axes of pyfar audio objects
+    and array likes.
+
+    Pyfar audio objects are: :py:func:`Signal`, :py:func:`TimeData`, and
```
f-brinkmann, line 1032: Maybe add a hint on which function is used in the background (np.dot or np.matmul come to my mind).

f-brinkmann (changes requested):
I like the idea of having this but I'm unsure about the following: depending on the kind of matrix multiplication, the underlying `frequencies` and `times` could be changed. Lets assume our data (time or frequency) is of shape (10 channels, 256 times/bins). If we apply any non-square matrix to the right, what would we end up with?

f-brinkmann:
We discussed that multiplication should never alter the number of time/frequency bins and that the axis on which it operates should become a parameter (default first two axes) and that broadcasting to higher dimensional Audio objects should be possible.

pyfar/classes/audio.py
```diff
@@ -1011,6 +1025,41 @@ def power(data: tuple, domain='freq'):
     return _arithmetic(data, domain, _power)
 
 
+def matrix_multiplication(data: tuple, domain='freq'):
+    """Matrix multiplication along first two axes of pyfar audio objects
```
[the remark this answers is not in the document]
sikersten: Added axes parameter to specify

pyfar/classes/audio.py
```diff
@@ -1021,7 +1039,122 @@ def power(data: tuple, domain='freq'):
...
 
 
-def _arithmetic(data: tuple, domain: str, operation: Callable):
+def matrix_multiplication(
+        data: tuple, domain='freq', axes=[(-2, -1), (-2, -1), (-2, -1)]):
+    """Matrix multiplication of multidimensional pyfar audio objects and/or
+    array likes.
+
+    This function is based on ``numpy.matmul`` and only acts on the channels,
+    not on the frequency/time dimension. Pyfar audio objects are:
```
f-brinkmann, line 1048: I'm not sure if this makes it clear. Maybe make more clear that the number of samlpes/frequency bins always stays the same.

pyfar/classes/audio.py
```diff
@@ -1021,7 +1039,122 @@ def power(data: tuple, domain='freq'):
...
+    :py:func:`Signal`, :py:func:`TimeData`, and :py:func:`FrequencyData`.
+
+    Parameters
+    ----------
+    data : tuple of the form (data_1, data_2, ..., data_N)
+        Data to be multiplied. Can contain pyfar audio objects or array likes.
+        The data is `broadcasted
+        <https://numpy.org/doc/stable/user/basics.broadcasting.html>`_
+        to the data with the highest number of channel dimensions. Pyfar audio
```
f-brinkmann, line 1057: maybe explicitly mention that this means `len(data.cshape)` (or is it `prod(data.cshape)`)?

pyfar/classes/audio.py
```diff
@@ -1021,7 +1039,122 @@ def power(data: tuple, domain='freq'):
...
+        See notes and examples below.
+
+    Returns
+    -------
+    results : Signal, TimeData, FrequencyData, numpy array
+        Result of the operation as numpy array, if `data` contains only array
+        likes and numbers. Result as pyfar audio object if `data` contains an
+        audio object. The `fft_norm` is ``'none'`` if all FFT norms are
+        ``'none'``. Otherwise the first `fft_norm` that is not ``'none'`` is
+        used.
```
f-brinkmann, line 1086: Is this still intended, or should it use our new 'broadcasting' rules for FFT normalization?

pyfar/classes/audio.py
```diff
@@ -1021,7 +1039,122 @@ def power(data: tuple, domain='freq'):
...
+    - In case of more than two arguments in ``data``, ``axes`` refers to all
+      partial operations.
+    - If the ``cshape`` of the arguments in data is 2-D, the arguments are
+      multiplied like conventional matrices along the time/frequency dimension.
+    - If either ``cshape`` is N-D, N > 2, the channels are treated as a stack
+      of matrices residing in the last two indexes and broadcasted accordingly.
+    - If the first ``cshape`` is 1-D, it is promoted to a 2-D ``cshape`` by
+      prepending a 1 to its dimensions.
+    - If the second ``cshape`` is 1-D, it is promoted to a 2-D ``cshape`` by
+      appending a 1 to its dimensions.
```
f-brinkmann, line 1101: What is the idea behind this case dependent handling? I'm wondering if this would mess with (AB)C = A(BC). Should we maybe insist that shapes match right away and not do any prepending/appending?

pyfar/classes/audio.py
```diff
@@ -1021,7 +1039,122 @@ def power(data: tuple, domain='freq'):
...
+        A list of tuples with the indices of the channels the
+        multiplication should operate on. The signature is (i,j),(j,k)->(i,k),
+        so the base elements of the multiplication are two-dimensional
+        matrices along the time/frequency dimension taken to be stored in the
+        indicated channels of the result. Negative values in ``axes`` refer to
+        ``cshape``, as the frequency/time axis is excluded. For instance, to
+        multiply matrices residing (stacked) in the last two channel
+        dimensions to be stored in the last two channel dimensions, this is
+        indicated by ``[(-2, -1), (-2, -1), (-2, -1)]``, which is the default.
+        See notes and examples below.
```
f-brinkmann, line 1077: This was not clear upon first reading to me. Should we make more clear that the first tuple gives the axes (defines the matrix) for the first input, the second axes do the same for the second input and the third define in what axes the result is saved?

pyfar/classes/audio.py
```diff
@@ -1021,7 +1039,122 @@ def power(data: tuple, domain='freq'):
...
+        objects can not be mixed, e.g., :py:func:`TimeData` and
+        :py:func:`FrequencyData` objects do not work together. See
+        :py:mod:`arithmetic operations <pyfar._concepts.arithmetic_operations>`
+        for possible combinations of Signal FFT normalizations.
+    domain : ``'time'``, ``'freq'``, optional
+        Flag to indicate if the operation should be performed in the time or
+        frequency domain. If working in the frequency domain, the FFT
+        normalization is removed before the operation (See
+        :py:func:`~pyfar.dsp.fft.normalization`). The default is ``'freq'``.
+    axes : list of 3 tuples
```
f-brinkmann, line 1067: What happens the input `data` contains more than two arrays/audio objects? Wouldn't this have list of `len(data)+1` tuples in this case?

pyfar/classes/audio.py
```diff
@@ -357,6 +359,14 @@ def __pow__(self, data):
     def __rpow__(self, data):
         return power((data, self), 'time')
 
+    def __matmul__(self, data):
+        return matrix_multiplication(
+            (self, data), 'time', axes=[(-2, -1), (-2, -1), (-2, -1)])
```
f-brinkmann, line 364: We would not have to specify `axes` here (and below), since we are using the default. We could of course if we want to be explicit.

f-brinkmann (changes requested):
Thanks for the effort. I marked a few possible improvements for the documentation and a few things to discuss. The code looks fine to me otherwise.

pyfar/classes/audio.py
sikersten: This is in line with the docstrings of the other arithmetic operations. What do you refer to?

pyfar/classes/audio.py
sikersten: Basically, it refers to the fact that - depending on the position of the signal in operation - a 1-D signal is either extended in the "row" or "column" channel dimension for the matrix multiplication before broadcasting. I'm not sure if this answers your question?

I would prefer allowing signals with a 1-D cshape. But we can also refer the ``reshape``, if we want to restrict.

pyfar/classes/audio.py
f-brinkmann: I was thinking about this: https://pyfar.readthedocs.io/en/latest/concepts/pyfar.arithmetic_operations.html. For example, we do not allow two signals with norms 'amplitude' and 'power' to be multiplied. Are the docstrings of the arithmetic functions outdated?

f-brinkmann: Fix according to #320

pyfar/classes/audio.py
sikersten: Mentioned it in issue #320. I'd suggest to solve in a separate PR for all arithmetic operations.

pyfar/classes/audio.py
mberz: I think the numpy default handling is very reasonable in this case.

f-brinkmann (approved):

pyfar/classes/audio.py
```diff
@@ -1301,6 +1492,36 @@ def _power(a, b):
...
+
+    # adjust axes according to time/frequency dimension
+
+    #  handle first input
+    if dtypes[0] in (Signal, TimeData, FrequencyData):
+        a = np.expand_dims(a, 0) if a.ndim == 2 else a
+        axes[0] = tuple([ax-1 if ax < 0 else ax for ax in axes[0]])
+    elif a.ndim < b.ndim:
+        a = np.expand_dims(a, -1)
+        axes[0] = tuple([ax-1 if ax < 0 else ax for ax in axes[0]])
```
f-brinkmann, line 1508: Is this required? An axis is already appended in `_get_arithmetic_data` (line 1466) if the input is an array and any input is a pyfar audio object. If all inputs are are arry likes, we probably shouldn't do any processing.

If it is not required, could the function could be simplified?
```python
#  handle first input
    if dtypes[0] in (Signal, TimeData, FrequencyData):
        a = np.expand_dims(a, 0) if a.ndim == 2 else a

    # handle second input
    if dtypes[1] in (Signal, TimeData, FrequencyData):
        b = np.expand_dims(b, 1) if b.ndim == 2 else b

    # adjust axis for output if output is a pyfar audio object
    if any(dt in (Signal, TimeData, FrequencyData) for dt in dtypes):
        axes[0] = tuple([ax-1 if ax < 0 else ax for ax in axes[0]])
        axes[1] = tuple([ax-1 if ax < 0 else ax for ax in axes[1]])
        axes[2] = tuple([ax-1 if ax < 0 else ax for ax in axes[2]])
```

f-brinkmann (commented):
I did to small commits for comments and code structure and have one question.

pyfar/classes/audio.py
sikersten: Thanks for noting the appending in ``_get_arithmetic_data``. It was possible to simplify the code by using the ``audio_type`` variable in ``_get_arithmetic_data`` and ``_matrix_multiplication``.

mberz (approved):

Issue #279, opened 2022-02-18
Rename xscale parameter in spectrum plots


sikersten:
## General

* pyfar version: 0.3.0

## Description

For plots including a frequency axis (line plots and 2d plots) the `xscale` parameter can be `'lin'` or `'log'`. It makes sense to rename it to `freq_scale` to be consistent between line and 2d plots (where the frequency can be on the y-axis).
This is planned to be done after #198 is merged.

PR #282, opened 2022-02-25
freq_scale parameter for 2D plots

sikersten:
 I'd suggest to already change to frequency axis parameter from `xscale` to `freq_scale` for the 2D plots before introducing them in develop.
Note: in this PR, only the top level functions (the user interface) is modified.  Changing all `xscale` parameters in the background (e.g., for interaction) can be done together with the line plots in the future, see #279.


al-de (approved):
looks good to me, thanks.

PR #283, opened 2022-02-25
Deprecate xscale parameter from all plots

### LINKED ISSUE #279: Rename xscale parameter in spectrum plots

f-brinkmann:

Closes #279 

- remove `xscale` parameter in favor or `freq_scale`
- add deprecation warnings


f-brinkmann (commented):
Looks good to me - thanks for the effort. I only wonder if we want to deprecate in 0.6.0 or 0.7.0? I think up to now we waited two minor versions which would be 0.6.0

sikersten:
Thanks, yes you are right, changed that!

sikersten (approved):

al-de (approved):
Looks good to me, thanks.

PR #300, opened 2022-04-07
Switch to CircleCI

mberz:
Use CircleCI instead of TravisCI

CircleCI has a favorable model of auto-replenishing open source credits on a monthly basis

- Add the respective config file
- Cleanup of pytest configs
- CircleCI does not work directly with tox, but with pytest
- Remove the Travis config file


f-brinkmann (approved):
Thanks for the effort - looks good to me. Should we take the opportunity and add Python 3.10 to the testing?

mberz:
I'll check if it's supported

sikersten (approved):

PR #303, opened 2022-04-25
Minimum phase algorithm for real world signals

f-brinkmann:

Add minimum-phase algorithm for *real-world* signals. The existing algorithms fail for creating minimum phase HRTFs and I assume they fail at comparable signals as well. I added an algorithm that is better in approximating the magnitude response at the cost of a worst approximation of the analytical minimum phase. Here are examples of a filter

filter

and an HRTF

HRTF

The code for the first plot is in the docstring. This is the code for the second (requires a SOFA file from http://dx.doi.org/10.14279/depositonce-5718.5)

```python
import pyfar as pf
import numpy as np

file = "FABIAN_HRIR_measured_HATO_0.sofa"
signal, sources, _ = pf.io.read_sofa(file)
idx, _ = sources.find_nearest_k(0, 1, 0)
signal = signal[idx, 0]

min_phase = pf.dsp.minimum_phase(signal, "hilbert_2", n_fft=4096)
urg_phase = pf.dsp.minimum_phase(signal)

pf.plot.time_freq(signal, unit="s", label="original")
pf.plot.time_freq(urg_phase, unit="s", label="Hilbert", linestyle="--")
pf.plot.time_freq(min_phase, unit="s", label="Hilbert 2", linestyle=":")

plt.legend()
```


mberz:
In my opinion, it makes sense to drop the scipy implementations.
The method implemented there is intended to be used for creating equivalent band-pass systems.
In both methods implemented in scipy, the magnitude spectrum is approximated from the square root of the power spectrum, under the constraint of finding a best fitting minimum-phase system with minimum pass-band ripple. This potentially leads to distortions in the magnitude spectrum for systems that only comprise a pass-band part with large ripple (This is typically the case for HRIRs, etc.).
The new method approximates only the phase for a given magnitude spectrum using the Hilbert transform of the real-valued cepstrum. The magnitude spectrum remains unchanged, resulting in no distortions of the magnitude spectrum. A drawback is that the zeros of the resulting minimum-phase system are closer to the unit circle of the pole-zero diagram compared to the scipy methods.

mberz (changes requested):

PR #313, opened 2022-05-23
Enhancement: signals parameters

f-brinkmann:

The parameters for the functions `pyfar.signals.impulse` and `pyfar.signals.sine` had to be scalars or of the same shape. Now any parameters that can be broadcasted to the same shape can be passed. This is more flexible, maintains backwards compatibility, and made the code shorter and easier to read.

- added new parameter handling
- added new tests


sikersten (commented):
Good idea, thanks for implementing!
Maybe a short broadcasting example in each of the function's docstring would be nice, as the explanations on the numpy homepage are very general?
Otherwise appoved :)

f-brinkmann:
Good idea - I added examples in the Notes sections.

sikersten (approved):

mberz (approved):

PR #322, opened 2022-06-21
Insert prefix_return in dB function

twennemann:
New Parameter 'prefix_return' in dB function.
Default False will do nothing.
Set True and the function will also return 'log_prefix'.

Added, because it is used in two functions currently being developed in branches (normalization and soft limiting)

### DIFF
--- a/pyfar/dsp/dsp.py
+++ b/pyfar/dsp/dsp.py
@@ -1384,7 +1384,7 @@ def convolve(signal1, signal2, mode='full', method='overlap_add'):
 
 
 def decibel(signal, domain='freq', log_prefix=None, log_reference=1,
-            prefix_return=False):
+            return_prefix=False):
     r"""Convert data of the selected signal domain into decibels (dB).
 
     The converted data is calculated by the base 10 logarithmic scale:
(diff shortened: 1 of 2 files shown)


pyfar/dsp/dsp.py
```diff
@@ -1444,10 +1445,16 @@ def decibel(signal, domain='freq', log_prefix=None, log_reference=1):
         +---------------------------------+--------------+
 
         The default is 1.
+    prefix_return : bool, optional
+        If prefix_return is ``True``, the function will also return the
+        log_prefix value. This can be used to delogrithmize the data. The
```
f-brinkmann, line 1450: We highlight parameters in italic:
```suggestion
        `log_prefix` value. This can be used to delogrithmize the data. The
```

f-brinkmann (approved):
Only two micro comments for the docstring - approved otherwise.

sikersten (approved):

sikersten (commented):
What about naming it ``return_prefix``? This would be more conform with our general strategy of "name it how you say it". I personally do not have a strong preference...

sikersten (approved):

PR #334, opened 2022-07-29
Hotfix for broken soundfile import on some machines

### LINKED ISSUE #332: Make python-soundfile optional
## General

python-soundfile sometines makes problems depending on the os and processor. It would be good to make pyfar importable in that case and raise a warning if soundfile could not be imported.

f-brinkmann:

Closes #332 

Check if soundfile can be imported. If not a warning is raised if a function is used that required soundfile.

I tried if the warnings are raised on my local machine. I suggest not to include this in testing. It woud require a separate test environment where we `pip uninstall soundfile`.


pyfar/io/io.py
```diff
@@ -14,10 +14,19 @@
...
 import numpy as np
 
+try:
+    import soundfile
+    soundfile_imported = True
+except ModuleNotFoundError or OSError:
+    soundfile_imported = False
+    soundfile_warning = (
+        "python-soundfile could not be imported. Try to install it using `pip "
+        "install soundfile. If this not works search the documentation for "
```
sikersten, line 27:
```suggestion
        "install soundfile`. If this not works search the documentation for "
```

sikersten (approved):

pyfar/io/io.py
f-brinkmann: committed :)

mberz (approved):

PR #338, opened 2022-08-10
feature/energy, power and rms function

twennemann:
We decided to create three new functions for energy, power and rms calculation in dsp. 

Implemented the functions, but the math command in the docsstring doesn't work proper yet. 
Created a test file, but also the tests are not working yet.


mberz:
The flake8 issues you're getting should be fixed on the main branch and can be ignored for now.

pyfar/dsp/dsp.py
```diff
@@ -1498,3 +1498,156 @@ def decibel(signal, domain='freq', log_prefix=None, log_reference=1,
...
+        \sum_{n=0}^{N-1}|x[n]|^2=\frac{1}{N}\sum_{k=0}^{N-1}|X[k]|^2,
+
+    as the calculation in time domain is equivalent to the frequency domain,
+    just the time domain is used in this function. Nevertheless, for the energy
+    calculation in the frequency domain the unitary norm needs to be used.
+
+    Parameters
+    ----------
+    signal : Signal
+        The signal to computed the energy from.
```
sikersten, line 1519:
```suggestion
    Compute the energy of a signal.

    According to the Parseval's theorem [#]_ the energy is

    .. math::

        \sum_{n=0}^{N-1}|x[n]|^2=\frac{1}{N}\sum_{k=0}^{N-1}|X[k]|^2,

    It is calculated based on the time data and therefore independent of the
    signal's ``fft_norm``.

    Parameters
    ----------
    signal : Signal
        The signal to compute the energy from.
```

pyfar/dsp/dsp.py
```diff
@@ -1498,3 +1498,156 @@ def decibel(signal, domain='freq', log_prefix=None, log_reference=1,
...
+           Transform”, Biomedical Signal and Image Processing, Spring 2005.
+    """
+    # check input
+    if not isinstance(signal, pyfar.Signal):
+        raise ValueError(f"signal is type '{signal.__class__}'"
+                         " but must be of type 'Signal'.")
+    # cumpute energy data
+    data = np.sum(signal.time**2, keepdims=keepdims)
+    if keepdims:
+        data = np.squeeze(data, axis=-1)
```
sikersten, line 1549: In my opinion, an ``axis=-1`` is missing here to sum over the time steps per channel. Accordingly, the returned `data` would have a shape which equals the signals ``cshape``.
``keepdims=True`` then leads to a shape of `data` which should be broadcastable to the shape of ``signal.time``, but it might make sense to just append an axis instead of calling ``np.sum`` with ``keepdims=True``, to be able to also broadcast it against ``signal.freq``.
```suggestion
    data = np.sum(signal.time**2, axis=-1)
    if keepdims:
        data=data[..., None]
```

pyfar/dsp/dsp.py
```diff
@@ -1498,3 +1498,156 @@ def decibel(signal, domain='freq', log_prefix=None, log_reference=1,
...
+    Parameters
+    ----------
+    signal : Signal
+        The signal to computed the energy from.
+    keepdims : bool, optional
+        If this is true, the axis which are reduced during summing up are
+        kept as a dimension with size one. Also, the output will broadcast
+        correctly with the input Signal. Thereafter, the last dimension of the
+        data gets squeezed.
+        The default is ``False``.
```
sikersten, line 1525: See comment on the computation along the time axis

pyfar/dsp/dsp.py
```diff
@@ -1498,3 +1498,156 @@ def decibel(signal, domain='freq', log_prefix=None, log_reference=1,
...
+    Compute the power of a signal in time domain. Therefore, the
+    Parseval's theorem [#]_ is used to calculate the energy of the signal with:
+
+    .. math::
+
+        \sum_{n=0}^{N-1}|x[n]|^2=\frac{1}{N}\sum_{k=0}^{N-1}|X[k]|^2,
+
+    As the calculation in time domain is equivalent to the frequency domain,
+    just the time domain is used in this function. Nevertheless, for the energy
+    calculation in the frequency domain the unitary norm needs to be used.
```
sikersten, line 1564: I'd suggest to just give the formula for the power computation here, the relation to energy and rms could be given in the ``Notes``

pyfar/dsp/dsp.py
```diff
@@ -1498,3 +1498,156 @@ def decibel(signal, domain='freq', log_prefix=None, log_reference=1,
...
+    Then, the rms is calculated by
+    :math:`\sqrt{\frac{1}{N}*energy}`.
```
sikersten, line 1617: I'd suggest to just give the formula for the rms computation here

tests/test_dsp_energy_power_rms.py
```diff
@@ -0,0 +1,24 @@
...
+    signal = pf.signals.sine(freq, n_samples, amplitude)
+    energy = pf.dsp.energy(signal)
+    answer_e = n_samples/2 * amplitude**2
+    power = pf.dsp.power(signal)
+    answer_p = amplitude**2 / 2
+    rms = pf.dsp.rms(signal)
+    answer_r = np.sqrt(1/2) * amplitude
+    npt.assert_almost_equal(energy, answer_e)
+    npt.assert_almost_equal(power, answer_p)
+    npt.assert_almost_equal(rms, answer_r)
```
sikersten, line 24: A test for a multichannel signal would be good to add here

sikersten (changes requested):
Thanks for implementing! Main comment refers to the sum along the time axis

pyfar/dsp/dsp.py
```diff
@@ -1498,3 +1498,144 @@ def decibel(signal, domain='freq', log_prefix=None, log_reference=1,
...
+    Compute the energy of a signal.
+
+    According to the Parseval's theorem [#]_ the energy is
+
+    .. math::
+
+        \sum_{n=0}^{N-1}|x[n]|^2=\frac{1}{N}\sum_{k=0}^{N-1}|X[k]|^2,
+
+    The function calculates the energy from the time data for each channel
+    separately.
```
f-brinkmann, line 1514: I would switch it
```suggestion
    Computes the channel wise energy in the time domain

    .. math::

        \sum_{n=0}^{N-1}|x[n]|^2=\frac{1}{N}\sum_{k=0}^{N-1}|X[k]|^2,

    , which is equivalent to the frequency domain computation according to Parseval's theorem [#]_.
```

pyfar/dsp/dsp.py
```diff
@@ -1498,3 +1498,144 @@ def decibel(signal, domain='freq', log_prefix=None, log_reference=1,
...
+    keepdims : bool, optional
+        If ``False``, the shape of the returned array equals ``signal.cshape``.
+        If ``True``, an axis will be appended, so it's broadcastable to the
+        shape of ``signal.time`` and ``signal.freq``.
+        The default is ``False``.
+
+    Returns
+    -------
+    data : numpy.ndarray
+        The total energy of the input signal.
```
f-brinkmann, line 1529:
```suggestion
        The channel-wisel energy of the input signal.
```

f-brinkmann (approved):
I would only suggest minor changes in the docstring. I commented on one docstring only, but I think they apply to all

sikersten (approved):

mberz:
@f-brinkmann [eac63a6](https://github.com/pyfar/pyfar/pull/338/commits/eac63a680043f84f2c7962a680aab90068ae08b7) 🍺 +1

pyfar/dsp/dsp.py
```diff
@@ -1498,3 +1498,144 @@ def decibel(signal, domain='freq', log_prefix=None, log_reference=1,
...
+    -----
+    Due to the calculation based on the time data, the returned energy is
+    independent of the signal's ``fft_norm``.
+    :py:func:`~pyfar.dsp.power` and :py:func:`~pyfar.dsp.rms` can be used
+    to compute the power and the rms of a signal.
+
+    References
+    -----------
+    .. [#] B. Delgutte and J. Greenberg , “The discrete Fourier
+           Transform”, Biomedical Signal and Image Processing, Spring 2005.
```
mberz, line 1541: I'd rather use literature concerned with audio signal processing, as people will have that more readily available.

pyfar/dsp/dsp.py
```diff
@@ -1498,3 +1498,144 @@ def decibel(signal, domain='freq', log_prefix=None, log_reference=1,
...
+
+    Parameters
+    ----------
+    signal : Signal
+        The signal to compute the energy from.
+    keepdims : bool, optional
+        If ``False``, the shape of the returned array equals ``signal.cshape``.
+        If ``True``, an axis will be appended, so it's broadcastable to the
+        shape of ``signal.time`` and ``signal.freq``.
+        The default is ``False``.
```
mberz, line 1524: 1. This is a bit misleading in my opinion. Keeping the dimensions to me sounds like "we maintain the dimensionality of the signal for which we calculate the energy" so, naturally I would expect a True here.
2. What's do we actually need this for?

pyfar/dsp/dsp.py
```diff
@@ -1498,3 +1498,144 @@ def decibel(signal, domain='freq', log_prefix=None, log_reference=1,
...
+    The RMS equals the square root of the signal's power.
+    :py:func:`~pyfar.dsp.energy` and :py:func:`~pyfar.dsp.power` can be used
+    to compute the energy and the power.
+    """
+    # check input
+    if not isinstance(signal, pyfar.Signal):
+        raise ValueError(f"signal is type '{signal.__class__}'"
+                         " but must be of type 'Signal'.")
+    # cumpute rms data
+    data = np.sqrt(np.sum(signal.time**2, axis=-1)/signal.n_samples)
```
mberz, line 1638: As you say in the note: sqrt(power), so why not do something like this:

```suggestion
    data = np.sqrt(power(signal))
```

pyfar/dsp/dsp.py
```diff
@@ -1498,3 +1498,144 @@ def decibel(signal, domain='freq', log_prefix=None, log_reference=1,
...
+
+    Parameters
+    ----------
+    signal : Signal
+        The signal to compute the power from.
+    keepdims : bool, optional
+        If ``False``, the shape of the returned array equals ``signal.cshape``.
+        If ``True``, an axis will be appended, so it's broadcastable to the
+        shape of ``signal.time`` and ``signal.freq``.
+        The default is ``False``.
```
mberz, line 1574: see above

pyfar/dsp/dsp.py
```diff
@@ -1498,3 +1498,144 @@ def decibel(signal, domain='freq', log_prefix=None, log_reference=1,
...
+
+    Parameters
+    ----------
+    signal : Signal
+        The signal to compute the RMS from.
+    keepdims : bool, optional
+        If ``False``, the shape of the returned array equals ``signal.cshape``.
+        If ``True``, an axis will be appended, so it's broadcastable to the
+        shape of ``signal.time`` and ``signal.freq``.
+        The default is ``False``.
```
mberz, line 1620: see above

mberz (changes requested):

f-brinkmann:
We decided to remove keepdims.

f-brinkmann (approved):
Thanks for the effort!

sikersten (approved):
Thanks for implementing!

mberz (approved):

PR #344, opened 2022-08-12
More convenient handling of data types in audio objects

### LINKED ISSUE #343: Data type for audio objects
## General

* pyfar version: all

Problem: The audio classes have the dtype argument, but it is not used

Solution:
- FrequencyData and TimeData should keep it, to make it possible to store real valued spactra and complex valued time signals
- Signal.time is always double, Signal.freq always complex.
- The new dtype default will be None in all cases and the dtype is inferred from the input data if not passed

f-brinkmann:

Closes #343 

- more consistent structure of `__init__` functions: TimeData and Frequency data call `Audio.__init__`. Signal calls `TimeData.__init__` or `FrequencyData.__init__`. Before it also called `Audio.__init__`
- `Audio.__init__` now also initializes the data. Otherwise the checks for the dtype would have to be made in all classes.
- Adjust tests and other function to changes
- Unify names of test files for audio objects

**Note:** if we want to allow complex TimeData later, we can revert https://github.com/pyfar/pyfar/pull/344/commits/5874f6fcc83a3eb5b5f183bdf1220cc355ecb629


pyfar/classes/audio.py
```diff
@@ -25,23 +25,33 @@ class _Audio():
...
+        # set data and check dtype
+        if dtype not in (None, float, complex):
+            raise ValueError("dtype must be None, float, or complex")
+        elif dtype is None:
+            self._data = np.atleast_2d(np.asarray(data))
+
+            # convert ints to float and check dtype of input
+            if str(self._data.dtype).startswith("int"):
+                self._data = self._data.astype(float)
+            elif str(self._data.dtype) not in ["float64", "complex128"]:
```
sikersten, line 49: This is system dependend, isn't it?

pyfar/classes/audio.py
```diff
@@ -56,7 +66,7 @@ def domain(self):
     def dtype(self):
         """The data type of the audio object. This can be any data type and
         precision supported by numpy."""
-        return self._dtype
+        return float if str(self._data.dtype).startswith("float") else complex
```
sikersten, line 69: What's the reason for not just returning the actual numpy dtype?

pyfar/classes/audio.py
```diff
@@ -365,11 +373,11 @@ class FrequencyData(_Audio):
     """Class for frequency data.
 
     Objects of this class contain frequency data which is not directly
-    convertable to the time domain, i.e., non-equidistantly spaced bins or
-    incomplete spectra.
+    convertible to the time domain, i.e., non-equidistantly spaced bins or
+    incomplete spectra. Can store real and complex valued data.
```
sikersten, line 377:
```suggestion
    incomplete spectra. Can store real or complex valued data.
```

sikersten (commented):
Thanks for going through it!
Looks good, I'm only a little confused, it the dtype of our objects match numpy's convention. Let's discuss in the weekly.

pyfar/classes/audio.py
f-brinkmann: yes at least for the int. I'm not sure about float and complex. Should we do
`elif not str(self._data.dtype).startswith(("float", "complex")):`

pyfar/classes/audio.py
f-brinkmann: to be consistent with the init functions. We init numpy arrays with `float` but the dtype than is `float64` or maybe `float32` depending on the system

f-brinkmann:
We decided to completely remove the dtype parameters from the audio classes.

pyfar/classes/audio.py
```diff
@@ -197,45 +194,34 @@ def __setitem__(self, key, value):
...
 
 
 class TimeData(_Audio):
     """Class for time data.
 
-    Objects of this class contain time data which is not directly convertable
-    to frequency domain, i.e., non-equidistant samples.
+    Objects of this class contain time data which is not directly convertible
+    to frequency domain, i.e., non-equidistant samples. Can store real and
+    complex valued data.
```
sikersten, line 205:
```suggestion
    to frequency domain, i.e., non-equidistant samples.
```

pyfar/classes/audio.py
```diff
@@ -365,26 +350,24 @@ class FrequencyData(_Audio):
     """Class for frequency data.
 
     Objects of this class contain frequency data which is not directly
-    convertable to the time domain, i.e., non-equidistantly spaced bins or
-    incomplete spectra.
+    convertible to the time domain, i.e., non-equidistantly spaced bins or
+    incomplete spectra. Can store real or complex valued data.
```
sikersten, line 354:
```suggestion
    incomplete spectra.
```

pyfar/io/io.py
```diff
@@ -311,12 +311,13 @@ def read_audio(filename, dtype='float64', **kwargs):
...
     dtype : {'float64', 'float32', 'int32', 'int16'}, optional
-        Data type of the returned signal, by default ``'float64'``.
+        Data type with which the wav file is read, by default ``'float64'``.
         Floating point audio data is typically in the range from
         ``-1.0`` to ``1.0``.  Note that ``'int16'`` and ``'int32'`` should only
         be used if the data was written in the same format. Integer data is in
         the range from ``-2**15`` to ``2**15-1`` for ``'int16'`` and from
-        ``-2**31`` to ``2**31-1`` for ``'int32'``.
+        ``-2**31`` to ``2**31-1`` for ``'int32'``. In any case, the data is
+        converted to float when writing it to the Signal.
```
sikersten, line 320:
```suggestion
        converted to float.
```

sikersten (changes requested):
Only one small mistake in a docstring, otherwise approved!

sikersten (approved):

pyfar/classes/audio.py
```diff
@@ -25,23 +25,33 @@ class _Audio():
...
+        # set data and check dtype
+        if dtype not in (None, float, complex):
+            raise ValueError("dtype must be None, float, or complex")
+        elif dtype is None:
+            self._data = np.atleast_2d(np.asarray(data))
+
+            # convert ints to float and check dtype of input
+            if str(self._data.dtype).startswith("int"):
+                self._data = self._data.astype(float)
+            elif not str(self._data.dtype).startswith(("float", "complex")):
```
mberz, line 49:
```suggestion
            elif not self._data.dtype.kind in ('f', 'c'):
```

pyfar/classes/audio.py
```diff
@@ -643,18 +636,24 @@ def freq(self):
     @freq.setter
     def freq(self, value):
         """Set the normalized frequency domain data."""
+        # check data type
         data = np.atleast_2d(np.asarray(value))
+        if not str(data.dtype).startswith(("int", "float", "complex")):
+            raise ValueError(
+                "data must be int, float, or complex and is casted to complex")
```
mberz, line 643:
```suggestion
                "data must be int, float, or complex and is cast to complex")
```

pyfar/classes/audio.py
```diff
@@ -643,18 +636,24 @@ def freq(self):
     @freq.setter
     def freq(self, value):
         """Set the normalized frequency domain data."""
+        # check data type
         data = np.atleast_2d(np.asarray(value))
+        if not str(data.dtype).startswith(("int", "float", "complex")):
```
mberz, line 641: see comment above

mberz (changes requested):
Thanks for implementing!
I'm not so happy that it's possible to create complex time data, which is then cast into float (in some cases without any warning of feedback).
```
pf.TimeData(np.ones(20, dtype=complex), np.arange(20)).time
pf.Signal(np.ones(20, dtype=complex), 1).time
```
These cases don't even give a warning:
```
sig = pf.signals.sine(1e3, 512)
sig.time = np.ones(512, dtype=complex)

sig = pf.TimeData(np.ones(20, dtype=float), np.arange(20))
sig.time = np.ones(20, dtype=complex)
sig.time
```
The more I think about this, the more I'd expect getting an error. This behavior is very or even too implicit in my opinion.

pyfar/classes/audio.py
```diff
@@ -197,47 +178,36 @@ def __setitem__(self, key, value):
...
 class TimeData(_Audio):
     """Class for time data.
 
-    Objects of this class contain time data which is not directly convertable
+    Objects of this class contain time data which is not directly convertible
     to frequency domain, i.e., non-equidistant samples.
 
     """
-    def __init__(self, data, times, comment=None, dtype=np.double):
+    def __init__(self, data, times, comment=None):
```
mberz, line 191:
```suggestion
    def __init__(self, data, times, comment=""):
```

pyfar/classes/audio.py
```diff
@@ -373,26 +349,25 @@ class FrequencyData(_Audio):
     """Class for frequency data.
 
     Objects of this class contain frequency data which is not directly
-    convertable to the time domain, i.e., non-equidistantly spaced bins or
+    convertible to the time domain, i.e., non-equidistantly spaced bins or
     incomplete spectra.
 
     """
-    def __init__(self, data, frequencies, comment=None, dtype=complex):
+    def __init__(self, data, frequencies, comment=None):
```
mberz, line 356:
```suggestion
    def __init__(self, data, frequencies, comment=""):
```

pyfar/classes/audio.py
```diff
@@ -409,16 +384,20 @@ def __init__(self, data, frequencies, comment=None, dtype=complex):
...
+        # init
         freqs = np.atleast_1d(np.asarray(frequencies).flatten())
-        if np.any(np.diff(freqs) <= 0) and len(frequencies) > 1:
-            raise ValueError("Frequencies must be monotonously increasing.")
         self._frequencies = freqs
-        # init data
         self.freq = data
 
+        # check frequencies
+        if np.any(np.diff(freqs) <= 0) and len(frequencies) > 1:
```
mberz, line 395: Do we not want to allow single frequency data?

tests/test_audio_signal.py
```diff
@@ -216,10 +271,10 @@ def test_setter_fft_norm():
...
 
 def test_dtype():
-    """Test for the getter of dtype."""
-    dtype = float
-    signal = Signal([1, 2, 3], 44100, dtype=dtype)
-    assert signal.dtype == dtype
+    """Test for converting int to float upon init."""
+
+    signal = Signal([1, 2, 3], 44100)
+    assert str(signal._data.dtype).startswith("float")
```
mberz, line 277:
```suggestion
    assert signal._data.dtype.kind == "f"
```

mberz (changes requested):
- I'd vote for using empty strings as default for comments instead of None types
- Why do we not allow setting single frequency FrequencyData
- A couple of minor things

f-brinkmann:
> * I'd vote for using empty strings as default for comments instead of None types
> 
>     * Why do we not allow setting single frequency FrequencyData
> 
>     * A couple of minor things

All make sense to me. I'll do the changes later and check what is required to allow single frequency data.

PR #346, opened 2022-09-01
fix FrequencyData dtype, dtype was ignored

ahms5:

no issue

- paramater dtype in FrequencyData is working now if not complex

### Labels

Label your issue to make it easier for us to assign and track:

Use one of these labels:
- **bug:** For bugs not on the master branch

Use a second label to let us know to which module the issue is related:

- **signal**

## Project

**Do not assign your issue to a project.** This is done automatically.

### DIFF
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -419,7 +419,7 @@ class FrequencyData(_Audio):
     @freq.setter
     def freq(self, value):
         """Set the frequency data."""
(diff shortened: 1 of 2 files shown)


f-brinkmann:
Is this a new fix or is it branched from #344 ? #344 is allowing double and float values for FrequencyData objects

ahms5:
It's a fix branched from developer

ahms5:
ok, I see. I think with #344 it does not make sence here.

PR #368, opened 2022-10-06
add `freq_range` kwarg to `deconvolve`

artpelling:
Calling `deconvolve` without a `freq_range` argument leads to an error in `regularized_spectrum_inversion`, because `freq_range` is mandatory. I've added a mechanism for setting a default value.

### DIFF
--- a/pyfar/dsp/dsp.py
+++ b/pyfar/dsp/dsp.py
@@ -1472,7 +1472,13 @@ def deconvolve(system_output, system_input, freq_range=None, fft_length=None, **
         The system input signal (e.g., used to perform a measurement).
         The system input signal is zero padded, if it is shorter than the
         system output signal.
-    fft_length: int or None
+    freq_range : tuple, array_like, double
+        The upper and lower frequency limits outside of which the
+        regularization factor is to be applied. The default ``None``
+        bypasses the regularization, which might cause numerical
+        instabilities in case of band-limited `system_input`. Also see
+        :py:func:`~pyfar.dsp.regularized_spectrum_inversion`.
+    fft_length : int or None
         The length the signals system_output and system_input are zero padded
         to before deconvolving. The default is None. In this case only the
         shorter signal is padded to the length of the longer signal, no padding
--- a/pyfar/dsp/dsp.py
+++ b/pyfar/dsp/dsp.py
@@ -1506,9 +1506,7 @@ def deconvolve(system_output, system_input, freq_range=None, fft_length=None, **
     if not system_output.sampling_rate == system_input.sampling_rate:
         raise ValueError("The two signals have different sampling rates!")
 
-    if freq_range is not None:
-        assert hasattr(freq_range, '__iter__') and len(freq_range) == 2
-    else:
+    if freq_range is None:
         freq_range = (0, system_input.sampling_rate)
 
     # Set fft_length to the max n_samples of both signals,
--- a/pyfar/dsp/dsp.py
+++ b/pyfar/dsp/dsp.py
@@ -1438,7 +1438,7 @@ def time_shift(signal, shift, unit='samples'):
     return shifted.reshape(signal.cshape)
 
 
-def deconvolve(system_output, system_input, fft_length=None, **kwargs):
+def deconvolve(system_output, system_input, freq_range=None, fft_length=None, **kwargs):
     r"""Calculate transfer functions by spectral deconvolution of two signals.
 
     The transfer function :math:`H(\omega)` is calculated by spectral
@@ -1506,6 +1506,11 @@ def deconvolve(system_output, system_input, fft_length=None, **kwargs):
     if not system_output.sampling_rate == system_input.sampling_rate:
         raise ValueError("The two signals have different sampling rates!")
 
+    if freq_range is not None:
+        assert hasattr(freq_range, '__iter__') and len(freq_range) == 2
+    else:
+        freq_range = (0, system_input.sampling_rate)
+
     # Set fft_length to the max n_samples of both signals,
     # if it is not explicitly set to a value
     if fft_length is None:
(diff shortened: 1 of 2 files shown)


pyfar/dsp/dsp.py
```diff
@@ -1506,6 +1506,11 @@ def deconvolve(system_output, system_input, fft_length=None, **kwargs):
     if not system_output.sampling_rate == system_input.sampling_rate:
         raise ValueError("The two signals have different sampling rates!")
 
+    if freq_range is not None:
+        assert hasattr(freq_range, '__iter__') and len(freq_range) == 2
+    else:
```
f-brinkmann, line 1511: We don't need to do the check here. It is done in the regularized inversion.
```suggestion
    if freq_range is None:
```

pyfar/dsp/dsp.py
```diff
@@ -1438,7 +1438,7 @@ def time_shift(signal, shift, unit='samples'):
     return shifted.reshape(signal.cshape)
 
 
-def deconvolve(system_output, system_input, fft_length=None, **kwargs):
+def deconvolve(system_output, system_input, freq_range=None, fft_length=None, **kwargs):
```
f-brinkmann, line 1441: It would usually be bad style to change the order of arguments, but in this case I'm also in favor of it. Any other thoughts?

f-brinkmann (changes requested):
Thx for reporting. I made some minor comments and would suggest to add the new parameter to the docstring:

```
freq_range : tuple, array_like, double
        The upper and lower frequency limits outside of which the
        regularization factor is to be applied. The default ``None``
        bypasses the regularization, which might cause numerical
        instabilities in case of band-limited `system_input`. Also see
        :py:func:`~pyfar.dsp.regularized_spectrum_inversion`.
```

f-brinkmann:
@mberz @sikersten, I would consider this a bug fix for the upcoming release. Do you agree? I checked our tests. We indeed always provided the parameter.

pyfar/dsp/dsp.py
artpelling: I added it to prevent unecessary computations. I guess the zero padding is not that bad :)

f-brinkmann (approved):
Good to go from my side

pyfar/dsp/dsp.py
sikersten: I not sure, if it is worth do break backwards compatibility here, unless we do not provide a default. Using the keywords allows an arbitrary order anyway... Any other opinions?
```suggestion
def deconvolve(system_output, system_input, fft_length=None, freq_range=None, **kwargs):
```

pyfar/dsp/dsp.py
```diff
@@ -1472,7 +1472,13 @@ def deconvolve(system_output, system_input, fft_length=None, **kwargs):
         The system input signal (e.g., used to perform a measurement).
         The system input signal is zero padded, if it is shorter than the
         system output signal.
-    fft_length: int or None
+    freq_range : tuple, array_like, double
+        The upper and lower frequency limits outside of which the
+        regularization factor is to be applied. The default ``None``
+        bypasses the regularization, which might cause numerical
+        instabilities in case of band-limited `system_input`. Also see
```
sikersten, line 1479:
```suggestion
        instabilities in case of a band-limited `system_input`. Also see
```

sikersten (commented):
Thanks for fixing!
I'm not sure about the default ``freq_range=None``. If I remember correctly, we decided to not provide a default for ``regularized_spectrum_inversion`` as there is not really a meaningful one. (e.g., what are the cases where (0, sampling_rate) to not cause artifacts? Why not (20, sampling_rate)?....). We should consider to have it the same for ``deconvolve``.
In that case, I would agree to add it as a parameter before `fft_length`.

f-brinkmann:
> I'm not sure about the default freq_range=None. If I remember correctly, we decided to not provide a default for regularized_spectrum_inversion as there is not really a meaningful one

I would argue that for the deconvolution, it makes sense to not have a restriction on the frequency range as default. This would give you the plain H = Y/X, that can be adjusted to H = Y * X^-1 including regularization if the problem requires it.

sikersten:
Yes, good point, convinced that the default makes sense here.

pyfar/dsp/dsp.py
mberz: I'll have to agree with Simon here, please do not change the order of input arguments for existing functions. It breaks backwards compatibility. Prior to this PR, freq_range could have been passed as part of kwargs.

pyfar/dsp/dsp.py
```diff
@@ -1506,6 +1512,9 @@ def deconvolve(system_output, system_input, fft_length=None, **kwargs):
     if not system_output.sampling_rate == system_input.sampling_rate:
         raise ValueError("The two signals have different sampling rates!")
 
+    if freq_range is None:
+        freq_range = (0, system_input.sampling_rate)
```
mberz, line 1516:
```suggestion
        freq_range = (0, system_input.sampling_rate/2)
```
The Nyquist limit is half the sampling rate.

pyfar/dsp/dsp.py
```diff
@@ -1528,7 +1537,7 @@ def deconvolve(system_output, system_input, fft_length=None, **kwargs):
     # multiply system_output signal with regularized inversed system_input
     # signal to get the system response
     system_response = (system_output *
-                       regularized_spectrum_inversion(system_input, **kwargs))
+                       regularized_spectrum_inversion(system_input, freq_range, **kwargs))
```
mberz, line 1540: 1. Not flake8 conform and will not pass on CircleCI
2. Please consider saving the inverse spectrum in a separate variable for improved readability

mberz (changes requested):
Thanks for implementing, I only have a few comments.

PR #369, opened 2022-10-07
Enhancement/coords

### LINKED ISSUE #353: Coordinate set and get inconsistant
## Description

inconsistant behavior with setting and getting coordinates. For the setter you get an array x 3 and for the setter you have to set each individual. See below

## What I Did

```
    sph = coords.get_sph()
    sph[:, :, 2] = 2
    # coords.set_sph(sph) # Does not work, would be more consistant
    coords.set_sph(sph[..., 0], sph[..., 1], sph[..., 2])
```

ahms5:

Closes #353

- general: all old functions should work as before, obsolte are marked and will be removed later (such as get_cart()), new properties are introduced and coordinates are always saved in carteasian coords.
- getter and setter for all coordinatesystem (azimuth, x, radius, etc.) parameters are added and testes (internal converters are used)
- getter and setter for each system and convention is also added (sph_top_elev, cart, etc) 
- get_x() and set_x() and sh_order are working but give PendingDeprecationWarning
- data is stored in cart and converted for other setter and getter
- system concept is not needed anymore
- doc not ready and not tested
- subclass for sh_order named SamplingSphere, is the name ok? 

### Open Questions

- find_mearest_k, find_mearest_sph, find_mearest_cart could be replaced by find_nearest(coords, ..) (what should be returned (mask, index, element)??
- SamplingSphere is the name ok?
- are the full setter and getter useful, or is there a better way (sph_top_elev, cart, etc)


ahms5:
todo @ahms5:
- [ ] make diff easier to read (use old order)
- [ ] more meaningful property names such as spherical instead of sph

ahms5:
branch was broken, see #398

PR #380, opened 2022-11-09
Filterbanks to classes

jake-is-ESD-protected:

-

- Fractional octave filter banks as class
- Inherit from FilterFIR or FilterSOS
- 2 seperate classes for either energy or amplitude preservation
- Docs and examples updated accordingly

### Labels

**enhancement:**
**documentation:**
**dsp:**
**filter:**

## Project

**Do not assign your issue to a project.** This is done automatically.


jake-is-ESD-protected:
btw, `flake8` and `pytest` failed for files which were not related to my enhancements and are of very low priority, e.g. an assertion fails for `1 == 0.99999999999` and `flake8` complains about an unused variable. Let me know if i should add any infos.

f-brinkmann (changes requested):
Thanks for the effort - most of the code looks pretty good already. I found three general things
- We have to keep the old functions for two minor releases. Can you please bring those back? We also have to add deprecation warnings and test for them - but lets wait with that. We are currently changing the way we handle the deprecation warnings.
- It would be more clean to make separate pulls for changing parameter names. Can you revert those? We would also have to make deprecation warnings if we decide to change them.
- The only thing you missed is testing if the new filter objects can be read and written with pyfar.io.write/read. For this to work you have to edit `_is_pyfar_type` in pyfar.io._codec and write a test in pyfar.tests.test_io (you could use `test_write_gammatone_bands` as a starting point)

pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -1,11 +1,608 @@
...
+        >>> y = FOFB.process(x)
+        >>> # frequency domain plot
+        >>> y_sum = pf.FrequencyData(np.sum(np.abs(y.freq)**2, 0),
+        ...     y.frequencies)
+        >>> pf.plot.freq(y)
+        >>> ax = pf.plot.freq(y_sum, color='k', log_prefix=10, linestyle='--')
+        >>> ax.set_title(
+        ...     "Filter bands and the sum of their squared magnitudes")
+        >>> plt.tight_layout()
+        >>> plt.show()
```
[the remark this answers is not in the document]
jake-is-ESD-protected: No reference in the previous versions, but it's **now** added to the filterbank object's stringdoc. I left the deprecated functions untouched.

pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -1,11 +1,608 @@
...
+        time. For an amplitude preserving filter bank with -6 dB cut-off
+        frequencies see
+        :py:func:`~pyfar.dsp.filter.ReconstructingFractionalOctaveBands`.
+
+    Parameters
+    ----------
+    num_fractions : int, optional
+        The number of bands an octave is divided into. Eg., ``1`` refers to
+        octave bands and ``3`` to third octave bands. The default is ``1``.
+    sampling_rate : int, optional
```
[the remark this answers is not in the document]
jake-is-ESD-protected: I'm not sure if i understand this correctly, which argument are you referring to?

pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -151,361 +752,3 @@ def _center_frequencies_fractional_octaves_iec(nominal, num_fractions):
     exact = reference_freq * octave_ratio**exponent
 
     return nominal, exact
-
-
-def fractional_octave_bands(
```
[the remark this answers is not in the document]
jake-is-ESD-protected: I added them back with the standard (not pyfar) deprecation warnings. This way i can refactor them quicker as soon as the own version of deprecation warnings gets included.

pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -226,6 +80,10 @@ def fractional_octave_bands(
         >>> plt.tight_layout()
 
     """
+    warnings.warn((
+            "This function will be deprecated in pyfar 0.6.0. "
```
f-brinkmann, line 84: Can you correct this to 0.8.0. The new classes will be released in 0.6.0 and deprecated to minor releases later. (See also below)

pyfar/dsp/filter/fractional_octaves.py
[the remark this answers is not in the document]
f-brinkmann: Thanks, for the effort. This style would be more common:

IEC 61260-1:2014, Octave-band and fractional-octave-band filters - Part 1: Specifications.

pyfar/dsp/filter/fractional_octaves.py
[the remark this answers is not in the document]
f-brinkmann: Sorry that could have been more clear. I would suggest the order: num_fractions, frequency_range, order, sampling_rate. That would be more consistent with the other filter functions and classes

pyfar/dsp/filter/__init__.py
```diff
@@ -16,8 +16,8 @@
 )
 
 from .fractional_octaves import (
-    fractional_octave_bands,
-    reconstructing_fractional_octave_bands,
+    FractionalOctaveBands,
```
f-brinkmann, line 19: Can you also bring back the imports for the old functions. Otherwise they won't be available for the user

f-brinkmann (changes requested):
Thanks again, I only had minor comments left. The final thing from my point is to add a test to pyfar/tests/test_io.py to make sure that writing and reading the filter Objects works. You can take `test_write_gammatone_bands` as a starting point

pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -971,8 +971,8 @@ def _encode(self):
     def _decode(cls, obj_dict):
         # initialize new class instance
         obj = cls(obj_dict["_num_fractions"], obj_dict["_frequency_range"],
-                  obj_dict["_sampling_rate"], obj_dict["_overlap"],
-                  obj_dict["_slope"], obj_dict["_n_samples"])
+                  obj_dict["_overlap"], obj_dict["_slope"],
+                  obj_dict["_n_samples"], obj_dict["_sampling_rate"])
```
f-brinkmann, line 975: `"_state"` and `"_initialized"` must also be added. Otherwise they would be lost when they do not have their default values. This must probably also be done for the reconstructing filter bank

tests/test_io.py
```diff
@@ -263,6 +263,38 @@ def test_write_filterSOS(filterSOS, tmpdir):
     assert actual == filterSOS
 
 
+def test_write_fractionaloctave_bands(tmpdir):
+    """ dsp.filter.FractionalOctaveBands
+    Make sure `read` understands the bits written by `write`
+    """
+    filename = os.path.join(tmpdir, 'fractionaloctave_bands.far')
+    fractionaloctave_bands = pyfar.dsp.filter.FractionalOctaveBands(
+        frequency_range=(20, 10000))
```
f-brinkmann, line 272: It would be good to save the filter with an non-empty state to see if everything is written and read. Could for example be done by
```python
fractionaloctave_bands.init_state((1, ))
fractionaloctave_bands.process(pf.signals.sine(1000, 441), True)
```

Would also be good for the reconstructing filter bank

f-brinkmann (changes requested):
Thanks for the effort looks very good now. I only had one straight forward fix for writing the filter state to disk.

pyfar/dsp/filter/fractional_octaves.py
jake-is-ESD-protected: Does that mean that the `FractionalOctaveBands` needs [this](https://github.com/jake-is-ESD-protected/pyfar/blob/bf187e53bdf55a9aa118ddc68f349345933bf412/pyfar/dsp/filter/gammatone.py#L388-L411) implementation? As for `_initialized`, where is that used? I can't find it in `GammatoneBands` as a reference.

f-brinkmann: Its already implemented in the filter-class from which you inherit. So no need to implement it here, but you need to save the parameters `"_state"` and `"_initialized"` , and use them for loading class objects. Does that make it more clear?

f-brinkmann (approved):
Thanks - all good to go from my side...

mberz (changes requested):
Thanks for implementing.
I do have a couple of major comments on the design and implementation. 
I've added these comments primarily to the FractionalOctaveBands class, but the same comments apply to the Reconstructing class.
Please also clean up the diff. Code which is not related changed functions/implemented classes should not be changed in this PR. This complicates tracking issues in the future due to messy diffs.

pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -308,12 +171,12 @@ def _coefficients_fractional_octave_bands(
...
             Wn = Wn[0]
             btype = 'highpass'
-            sos_hp = spsignal.butter(order, Wn, btype=btype, output='sos')
+            sos_hp = sgn.butter(order, Wn, btype=btype, output='sos')
             sos_coeff = pf.classes.filter._extend_sos_coefficients(
                 sos_hp, order)
         else:
             btype = 'bandpass'
-            sos_coeff = spsignal.butter(
+            sos_coeff = sgn.butter(
```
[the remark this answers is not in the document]
jake-is-ESD-protected: refactored back.

pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -308,12 +171,12 @@ def _coefficients_fractional_octave_bands(
                 bandpass'.format(np.round(freqs_upper[idx], decimals=1)))
             Wn = Wn[0]
             btype = 'highpass'
-            sos_hp = spsignal.butter(order, Wn, btype=btype, output='sos')
+            sos_hp = sgn.butter(order, Wn, btype=btype, output='sos')
```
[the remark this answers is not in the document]
jake-is-ESD-protected: ditto

pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -493,7 +360,7 @@ def reconstructing_fractional_octave_bands(
     time = pf.dsp.fft.irfft(g, n_samples, sampling_rate, 'none')
 
     # window
-    time *= spsignal.windows.hann(time.shape[-1])
+    time *= sgn.windows.hann(time.shape[-1])
```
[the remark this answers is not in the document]
jake-is-ESD-protected: ditto

pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -509,3 +376,758 @@ def reconstructing_fractional_octave_bands(
...
+        """Get the frequency range of the filter bank in Hz"""
+        return self._frequency_range
+
+    @property
+    def order(self):
+        """Get the filter order"""
+        return self._order
+
+    @property
+    def norm_frequencies(self):
```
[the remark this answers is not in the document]
jake-is-ESD-protected: Refactored to `nominal_frequencies`

pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -509,3 +376,758 @@ def reconstructing_fractional_octave_bands(
...
+        """Get the IEC center frequencies of the (fractional)
+         octave filters in Hz"""
+        return self._norm_frequencies
+
+    @property
+    def exact_frequencies(self):
+        """Get the exact center frequencies of the (fractional)
+        octave filters in Hz"""
+        return self._exact_frequencies
```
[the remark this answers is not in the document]
jake-is-ESD-protected: Refactored to `center_frequencies`

pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -509,3 +376,758 @@ def reconstructing_fractional_octave_bands(
...
+
+    @property
+    def frequency_range(self):
+        """Get the frequency range of the filter bank in Hz"""
+        return self._frequency_range
+
+    @property
+    def order(self):
+        """Get the filter order"""
+        return self._order
```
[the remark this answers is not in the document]
jake-is-ESD-protected: removed

pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -509,3 +376,758 @@ def reconstructing_fractional_octave_bands(
...
+
+    @property
+    def n_bands(self):
+        """Get the number of bands in the filter bank"""
+        return len(self._exact_frequencies)
+
+    @property
+    def sampling_rate(self):
+        """Get the sampling rate of the filter bank in Hz"""
+        return self._sampling_rate
```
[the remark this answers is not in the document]
jake-is-ESD-protected: removed

jake-is-ESD-protected:
Due to me syncing my fork after not contributing for about 3 months, there are now merge conflicts within `test_filter.py`. Should I resolve that? Let me know.

pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -509,3 +376,758 @@ def reconstructing_fractional_octave_bands(
...
+        exponent = (indices/num_fractions)
+    else:
+        indices = np.around(
+            2.0*num_fractions *
+            np.log(nominal/reference_freq) / np.log(octave_ratio) - 1)/2
+        exponent = ((2*indices + 1) / num_fractions / 2)
+
+    exact = reference_freq * octave_ratio**exponent
+
+    return nominal, exact
```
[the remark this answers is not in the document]
mberz: Yes, no worries about the import. The diff already looks much cleaner.

pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -509,3 +529,595 @@ def reconstructing_fractional_octave_bands(
...
+            Second order section filter coefficients with shape (.., 6)
+
+        Notes
+        -----
+        This function uses second order sections of butterworth filters for
+        increased numeric accuracy and stability.
+        """
+
+        f_crit = fractional_octave_frequencies(
+            num_fractions, frequency_range, return_cutoff=True)[2]
```
mberz, line 721: Examplary for the remainder of this method:
```suggestion
        f_crit = fractional_octave_frequencies(
            self.num_fractions, self.frequency_range, return_cutoff=True)[2]
```

pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -509,3 +529,595 @@ def reconstructing_fractional_octave_bands(
...
+    def n_bands(self):
+        """Get the number of bands in the filter bank"""
+        return len(self._center_frequencies)
+
+    def _get_coefficients(
+            self,
+            sampling_rate,
+            num_fractions,
+            frequency_range,
+            order):
```
mberz, line 692:
```suggestion
    @property
    def coefficients(self):
```
This can be converted into a property, overloading the existing coefficients property.
All input variables to this method are part of the class and hence can be accessed inside the class without passing them as input parameters.

pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -509,3 +529,595 @@ def reconstructing_fractional_octave_bands(
...
+    def nominal_frequencies(self):
+        """Get the IEC center frequencies of the (fractional)
+         octave bands in Hz"""
+        return self._nominal_frequencies
+
+    @property
+    def center_frequencies(self):
+        """Get the exact center frequencies of the (fractional)
+        octave bands in Hz"""
+        return self._center_frequencies
```
mberz, line 674: The calculation can be done dynamically using the classes' properties.
```suggestion
        return f_crit = fractional_octave_frequencies(
            self.num_fractions, self.frequency_range, return_cutoff=True)[0]
```

pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -509,3 +529,595 @@ def reconstructing_fractional_octave_bands(
...
+
+    @property
+    def cutoff_frequencies(self):
+        """Get the cutoff frequencies of the (fractional)
+        octave bands in Hz"""
+        return self._cutoff_frequencies
```
mberz, line 680: see above

pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -509,3 +529,595 @@ def reconstructing_fractional_octave_bands(
...
+
+    @property
+    def n_bands(self):
+        """Get the number of bands in the filter bank"""
+        return len(self._center_frequencies)
```
mberz, line 685: I think this is redundant with Filter.n_channels

pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -509,3 +529,595 @@ def reconstructing_fractional_octave_bands(
...
+        self._nominal_frequencies,\
+            self._center_frequencies,\
+            self._cutoff_frequencies = fractional_octave_frequencies(
+                self._num_fractions, self._frequency_range, True)
+
+        coefficients = self._get_coefficients(
+                sampling_rate=self._sampling_rate,
+                num_fractions=self._num_fractions,
+                frequency_range=self._frequency_range,
+                order=self._order)
```
mberz, line 633: These do not necessarily need to be saved bit can be calculated in their respective getters depending on the parameters of the filter bank above.

pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -509,3 +529,595 @@ def reconstructing_fractional_octave_bands(
...
+    @property
+    def frequency_range(self):
+        """Get the frequency range of the filter bank in Hz"""
+        return self._frequency_range
+
+    @property
+    def nominal_frequencies(self):
+        """Get the IEC center frequencies of the (fractional)
+         octave bands in Hz"""
+        return self._nominal_frequencies
```
mberz, line 668: See comment below (my bad, I missed this)

mberz (changes requested):
Thanks, for implementing the changes. 
I have a few additional comments on the conceptual design part.
I think it makes sense to use a couple more advantages that a class implementation enables here.
Your implementation is very static with regard to the configuration of the filter bank. Once created, most parameters are fixed. 
A more convenient implementation could be to allow setting parameters such as the sampling rate, the filer order, the frequency range and updating the filter coefficients accordingly. The calculation of the filter coefficients can then also be done as a read-only property (instead of storing the actual coefficients). 
I've added a few comments to the IIR version of the filter bank. However, the ideas apply to the other filter bank as well. Feel free to send me a message if things are unclear or need further discussion.

pyfar/dsp/filter/fractional_octaves.py
jake-is-ESD-protected: Doing that causes an `AttributeError` in `pyfar/classes/filter.py:139`. It seems that me overloading the `coefficients` property disables the `@coefficients.setter` in the base class, causing the property to be impossible to set. If I then add this setter to the `FractionalOctaveBands` class as well, i don't know where to call it, as the property is not recognized in the constructor. I'm essentially trying to recreate `pyfar/classes/filter.py:139`.

jake-is-ESD-protected: There's a tricky issue with that. Simply overloading an returning the coefficients causes an `AttributeError: can't set attribute 'coefficients'`. Taking a look at lines `171`-`174` of `filter.py` shows that I need to specify the property's `setter` explicitly. However, this `setter` gets called on line `139` if `coefficients` is passed along to the filter's constructor, which does not happen in my case, as the constructor of `FractionalOctaveBands` does not take any coefficients. These get created from more basic arguments. In essence, i think that the parent class is forcing me to go with a `setter` approach, even though I don't have/require a `setter`. A simple 
```python
@property
def coefficients(self):
    # calc
    return sos_result
```
should be enough, but again, this throws an error.

Btw, giving the property a different name fixes the issue, so it's definetely the base class messing with the implementation.

f-brinkmann: Not sure if I fully understand the issue. Can you commit or give a minimal example that raises the error. It is possible to explicitly override properties, would this help: https://stackoverflow.com/questions/7019643/overriding-properties-in-python

Issue #358, opened 2022-09-27
Inconsistent mode descriptors in zero padding function


mberz:
## Description

The mode descriptors for zero-padding are named inconsistently:

modes 'before' and 'after' are subject to a reference, 'center' is not. Since in a one word descriptor, references are absent.
I'd propose renaming them to something like 'beginning' and 'end'.

PR #381, opened 2022-11-10
zero padding descriptors

twennemann:
- Modi "before" and "after" are called "beginning" and "end" now. 
- Wrote PendingDeprecationWarning, that modi "before" and "after" can no longer be used in Pyfar Version 0.6.0. 
- PendingDeprecationWarning was ignored by default, so I used 'warnings.simplefilter('always')' to show the warning when modi "before" and "after" will be used. Is there a way to activate PendingDeprecationWarning permanent without using this line everytime?
- Insert 'test_pad_zero_modi' into test_deprecations.py

### DIFF
--- a/pyfar/dsp/dsp.py
+++ b/pyfar/dsp/dsp.py
@@ -978,7 +978,7 @@ def pad_zeros(signal, pad_width, mode='end'):
         ``'end'``
             Append zeros to the end of the signal
         ``'beginning'``
-            Pre-pend zeros before the starting time of the signal
+            Prepend zeros to the beginning of the signal
         ``'center'``
             Insert the number of zeros in the middle of the signal.
             This mode can be used to pad signals with a symmetry with respect
--- a/pyfar/dsp/dsp.py
+++ b/pyfar/dsp/dsp.py
@@ -1003,10 +1003,9 @@ def pad_zeros(signal, pad_width, mode='end'):
         raise TypeError('Input data has to be of type: Signal.')
 
     if mode in ['before', 'after']:
-        warnings.simplefilter('always')
         warnings.warn(('Mode "before" and "after" will be renamed into '
                        '"beginning" and "end" and can no longer be used in '
-                       'Pyfar 0.5.8.'), PendingDeprecationWarning)
+                       'Pyfar 0.8.0.'), DeprecationWarning)
 
         mode = 'beginning' if mode == 'before' else 'end'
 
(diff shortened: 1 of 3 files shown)


twennemann:
We decided to use DeprecationWarning instead of PendingDeprecationWarning.

pyfar/dsp/dsp.py
```diff
@@ -995,18 +995,26 @@ def pad_zeros(signal, pad_width, mode='after'):
...
     """
 
     if not isinstance(signal, pyfar.Signal):
         raise TypeError('Input data has to be of type: Signal.')
 
+    if mode in ['before', 'after']:
+        warnings.simplefilter('always')
+        warnings.warn(('Mode "before" and "after" will be renamed into '
+                       '"beginning" and "end" and can no longer be used in '
+                       'Pyfar 0.5.8.'), PendingDeprecationWarning)
```
f-brinkmann, line 1009: This would be released in 0.6.0 and deprecated two minor releases later:
```suggestion
                       'Pyfar 0.8.0.'), PendingDeprecationWarning)
```

tests/test_deprecations.py
```diff
@@ -167,3 +167,9 @@ def test__check_time_unit():
             # remove xscale from pyfar 0.6.0!
             create_figure()
             pf.plot._utils._check_time_unit(None)
+
```
f-brinkmann, line 170: Can you also add a test to check if the function was deleted if the pyfar version if 0.8.0 or larger. I think we did that for the other cases. Here's an example:
```
if version.parse(pf.__version__) >= version.parse('0.5.0'):
        with pytest.raises(AttributeError):
            # remove get_nearest_k() from pyfar 0.5.0!
            coords.get_nearest_k(1, 0, 0)
```

f-brinkmann (changes requested):
Thanks for implementing. I had some small comments and I think that the parameter values should be changed in the tests for `pad_zeros` as well. Otherwise the tests would raise warnings after switching to `DeprecationWarning`

sikersten:
Closes issue #358

pyfar/dsp/dsp.py
```diff
@@ -975,16 +975,16 @@ def pad_zeros(signal, pad_width, mode='after'):
     mode : str, optional
         The padding mode:
 
-        ``'after'``
+        ``'end'``
             Append zeros to the end of the signal
-        ``'before'``
+        ``'beginning'``
             Pre-pend zeros before the starting time of the signal
```
f-brinkmann, line 981:
```suggestion
            Prepend zeros to the beginning of the signal
```

f-brinkmann (changes requested):
Just one small comment for the docstring and I found that 'before' and 'after' are still used in `test_dsp_pad_zeros`

f-brinkmann (approved):

sikersten (approved):

PR #397, opened 2022-11-25
Feature/warning class

twennemann:
We decided to implement a custom class for DeprecationWarnings called "PyfarDeprecationWarning" to avoid using `warnings.simplefilter("always")`.

### DIFF
--- a/pyfar/__init__.py
+++ b/pyfar/__init__.py
@@ -4,7 +4,7 @@
 
 __author__ = """The pyfar developers"""
 __email__ = ''
-__version__ = '0.5.0'
+__version__ = '0.5.1'
 
 
 from .classes.audio import Signal, TimeData, FrequencyData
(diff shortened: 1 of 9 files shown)


f-brinkmann (changes requested):
Thanks for the effort. Looks good to me, but I would suggest to hide the access to the warning from the user, i.e, to remove it from `pyfar/__init__.py`

pyfar/__init__.py
```diff
@@ -14,6 +14,7 @@
 from .classes.orientations import Orientations
 from .classes.filter import FilterFIR, FilterIIR, FilterSOS
 
+
```
sikersten, line 17:
```suggestion

```

sikersten (approved):

ahms5:
I usually use **PendingDeprecationWarning** and this works very well. The warning raises in my case.

ahms5 (commented):
I usually use PendingDeprecationWarning and this works very well. The warning raises in my case.
Then we might not need any new class.

f-brinkmann (approved):

f-brinkmann:
> I usually use PendingDeprecationWarning and this works very well.
We initially also used that Warning but it is not shown to the user by default. Thats why we thought making a new class is the cleanest way. If we wanted to use PendingDeprecationWarning, we had to manually switch it on every time we use it.

sikersten (approved):

ahms5 (approved):

PR #398, opened 2022-11-29
New Coordinate structure - update

### LINKED ISSUE #353: Coordinate set and get inconsistant

ahms5:

Closes #353

- general: all old functions should work as before, obsolte are marked and will be removed later (such as get_cart()), new properties are introduced and coordinates are always saved in carteasian coords.
- getter and setter for all coordinatesystem (azimuth, x, radius, etc.) parameters are added and testes (internal converters are used)
- getter and setter for each system and convention is also added (sph_top_elev, cart, etc) 
- get_*() and set_*() and sh_order are working but give PendingDeprecationWarning
- data is stored in cart and converted for other setter and getter
- system concept is not needed anymore, but still there
- doc not ready and not tested
- subclass for sh_order named SamplingSphere, is the name ok? 
- demo need to be updated

### Open Questions

- find_nearest_k, find_nearest_sph, find_nearest_cart could be replaced by find_nearest(coords, ..) (what should be returned (mask, index, element)??, maybe a mask could be good, so that it can be used like `coords[coords.find_nearest(other_coords)]`, maybe sth like find_nearest_mask(...), find_nearest_index(...)
- SamplingSphere is the name ok?
- are the full setter and getter useful, or is there a better way (spherical_top_elevevation, cartesian, etc) 
- additional setter such as from_array, to set Coordinates from a single array, not 3 for each dimension

### TODO

- [ ] update doc and test it
- [ ] Demo need to be updated
- [ ] include new structure to pyfar which uses Coordinates


pyfar/classes/coordinates.py
```diff
@@ -606,7 +651,261 @@ def csize(self):
         The channel size gives the number of points stored in the coordinates
         object.
         """
-        return self._points.size // 3
+        return self._x.size
+
+    @property
+    def cartesian(self):
```
ahms5, line 657: all properties are according to [this figure](https://pyfar.readthedocs.io/en/latest/concepts/pyfar.coordinates.html), does it make sence? phi and theta are usally used as in [wiki](https://en.wikipedia.org/wiki/Spherical_coordinate_system) what should we use?

pyfar/classes/coordinates.py
```diff
@@ -919,24 +1237,133 @@ def find_nearest_sph(self, points_1, points_2, points_3, distance,
...
         # get the points
         distance, index, mask = self._find_nearest(
-            points_1, points_2, points_3,
-            domain, convention, unit, show, distance, 'sph', atol,
-            np.max(radius))
+            coords, show, distance, 'sph', atol)
+
+        return index, mask
+
+    def find_nearest_points(self, coords, number_of_points=1):
```
ahms5, line 1259: replaces 
`find_nearest_k(self, points_1, points_2, points_3, k=1,
                       domain='cart', convention='right', unit='met',
                       show=False):`
what should be the return value, if we fix one we can do sth like 
`found = coords[coords.find_nearest(coords2find)]`
which would be very clean. I would also go for indexes then the `coords2find` Coordinates can be more than one and it can be used to restructure arrays. (same for slice)

pyfar/classes/coordinates.py
```diff
@@ -919,24 +1237,133 @@ def find_nearest_sph(self, points_1, points_2, points_3, distance,
...
+            number_of_points <= self.csize,\
+            "number of points must be an integer > 0 and <= self.csize."
+
+        distance, index, mask = self._find_nearest(
+            coords, False, value=number_of_points, measure='k')
+
+        index = np.where(mask)
+        return distance, index, mask
+
+    def find_nearest_by_distance(self, coords, distance, measure='direct'):
```
ahms5, line 1293: replaces `find_nearest_sph()` and `find_nearest_cart()`, measure defines which one is used. I was thinking this is a cleaner way of implementaion.

ahms5: And also removed the plot option, it is described in the example, how to do it if needed.

pyfar/classes/coordinates.py
```diff
@@ -1456,20 +1878,7 @@ def _make_system(self, domain=None, convention=None, unit=None):
 
         return system
 
-    def _return_system(self, pts1, pts2, pts3, new_system, convert):
```
ahms5, line 1459: I think the internal _system is not needed in the future, since it is always saved in cartesian.

tests/test_coordinates.py
```diff
@@ -269,7 +269,7 @@ def test_show():
     plt.close("all")
```
ahms5, line 269: old tests should work as they are, for new features new tests are added

pyfar/classes/coordinates.py
```diff
@@ -19,44 +20,26 @@ class Coordinates():
...
-                 weights=None, sh_order=None, comment=""):
+    _x: np.array = np.empty
+    _y: np.array = np.empty
+    _z: np.array = np.empty
+    _weights: np.array = None
+    _sh_order: int = None
+    _comment: str = None
+    _system: dict = None
+
+    def __init__(
```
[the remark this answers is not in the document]
ahms5: we could also introduce sth like `from_*_array([azimuth, elevation, radius], weights=None, sh_order=None, comment=""))`
sth like this was done [here](https://github.com/mberz/spharpy/blob/main/spharpy/samplings/coordinates.py)

tests/test_coordinates.py
ahms5: I just changed them for private variables or methods

pyfar/classes/coordinates.py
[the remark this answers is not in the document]
mberz: I'm in favor of the `@classmethod` `from_*` option, to de-clutter  `__init__`. The points_1, etc. naming is not very intuitive to use, as one quickly forgets which number corresponds to which coordinate dimension.

ahms5:
We decided the following:
- init will be changed to just x,y,z
- from_* methods will be introduced for all coordinate systems
- from_array will not be implemented
- find_nearest_sph -> find_nearest_angular and include toleratance for radius
- find_nearest_cart -> find nearest_euclidean 
- find_nerest(Coordinates) for one closest point
- find_nearest_k(Coordinates)  for many multiple points
- redefine show(index) (use index instead of mask)
- find* functions  just return index not mask
- systems, which just show docu in the consol,  will be deleted

Thinks to be decided later
- SamplingSphere in pyfar or spharpy

please comment if i got sth wrong!

ahms5:
We decided to split it into different PR see #429
find nearest and find slice will follow

f-brinkmann:
moved to #429

Issue #389, opened 2022-11-11
Normalize fails for TimeData, FrequencyData containing nan elements


mberz:
## General

* pyfar version: v0.5.x

## Description

`pyfar.dsp.normalize` fails if the input data contains nans, normalizing the result to contain nan values only.
This only affects `TimeData` and `FrequencyData` which support storing nan values.

f-brinkmann:
I wasn't aware that those should be able to hold NaNs. I'm not sure if we are testing that anywhere. If it is the case, we should probably add tests for all functions that accept TimeData and FrequencyData objects.

PR #399, opened 2022-12-01
Accept Nans in Normalize 

twennemann:
Solves #389 

We decided
- We allow NaNs in normalize


f-brinkmann:
Thanks for the effort. The code looks good. But I wonder if we should discuss this in more detail. Which functions should allow NaNs? Should there be a flag, that has to be set or should it be allowed by default? Is there a good way of testing this for all functions that allow it?

ahms5 (commented):
looks good to me, I think we should add that nans are ignored in the doc.

f-brinkmann:
We discuss that we explicitly want to add an `ignore_nan` flag to the functions that except time data and NaNs would be a problem. This is also relevant for `dsp.average`.

mberz:
This is an example of the naming convention scipy follows:

https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.pmean.html#scipy.stats.pmean

twennemann:
Added `nan_policy` parameter to normalize function. It follows scipys naming convention. 
Use:
- `propagate` --> Output will be NaN if Signal includes NaNs (with np.max/np.mean)
- `omit` --> NaNs will be omitted in the normalization (with np.nanmax/np.nanmean)
- `raise`--> A `ValueError` will be raised if the signal includes NaNs

I thought `nan_policy='raise'` might be a logical default, so the user will be informed that the signal includes NaNs and to choose a policy.

pyfar/dsp/dsp.py
```diff
@@ -1997,7 +1997,7 @@ def average(signal, mode='linear', caxis=None, weights=None, keepdims=False):
 
 def normalize(signal, reference_method='max', domain='time',
               channel_handling='individual', target=1, limits=(None, None),
-              unit=None, return_reference=False):
+              unit=None, nan_policy='raise', return_reference=False):
```
f-brinkmann, line 2000: I think the new parameter should go last to maintain backwards compatability

pyfar/dsp/dsp.py
```diff
@@ -2183,9 +2206,13 @@ def normalize(signal, reference_method='max', domain='time',
             input_data = np.abs(signal.freq)
     # get values for normalization max or mean
         if reference_method == 'max':
-            reference = np.max(input_data[..., limits[0]:limits[1]], axis=-1)
+            get_max = np.nanmax if nan_policy == 'omit' else np.max
+            reference = get_max(input_data[..., limits[0]:limits[1]],
+                                axis=-1)
```
f-brinkmann, line 2211: Why not directly use `np.nanmax(..., nan_policy=nan_policy)`?
Might also be changed in the elif case below.

f-brinkmann (changes requested):
Looks good to me. I had only one minor comment and question.

f-brinkmann (approved):

sikersten (changes requested):
Thanks for implementing!
I tested with a Signal object, `s = pf.Signal([2, -1, 0, np.nan], 44100)`. The mode `nan_policy='omit' ` does not seem to work in this case. Could you please add a test and check what the reason is?

twennemann:
> Thanks for implementing! I tested with a Signal object, `s = pf.Signal([2, -1, 0, np.nan], 44100)`. The mode `nan_policy='omit' ` does not seem to work in this case. Could you please add a test and check what the reason is?

To apply normalization we calculate the normalized signal as followed:
`normalized_signal = signal.copy() * target / reference_norm`

If the signal object, which is used for a calculation, is type `pf.Signal` and it includes NaN values, every value will be returned with NaN.  So this problem needs to be fixed in the Signal class.

sikersten (approved):

Issue #405, opened 2022-12-10
Improve plots in documentation


f-brinkmann:
## General

* pyfar version: 0.5.1

## Description

The plots in the documentation that have a time axis could be improved. The time axis is given in seconds in most cases but using milliseconds, samples, etc. would often be better. After building the documentation using `sphinx make html`, the plots are locally stored at **pyfar/docs/_build/plot_directive**

PR #406, opened 2022-12-10
Remove private tight_layout plot functionality

### LINKED ISSUE #363: Deprecation of Matplotlib's tight_layout module
## General

* pyfar >= 0.4.0
* Matplotlib >= 3.6

We use the tight_layout module in pyfar.plot._utils to test if tight layout can be applied. But the module will be deprecated:
https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.6.0.html#modules

We might be able to maintain functionality using new layout options:
https://matplotlib.org/stable/users/prev_whats_new/whats_new_3.6.0.html#new-layout-engine-module

The simplest way might be to switch to the constrained layout, which is no longer experimental:
https://matplotlib.org/stable/users/prev_whats_new/whats_new_3.6.0.html#constrained-layout-is-no-longer-considered-experimental

f-brinkmann:

Closes #363

- update plotstyles to use constrained layout
- remove use of plot._utils._tight_layout (uses Matplotlib functionality that will be removed)
- update tests
- update and check baseline plots
- remove use of matplotlib.pyplot.tight_layout everywhere in pyfar (not needed anymore)
- Check and fix plots in documentation
- opened #405 instead of fixing it directly to avoid having to buy to many beers for the others


ahms5 (approved):

sikersten (commented):
Thanks for the effort! Implementation looks good!
The test plots with incl. group delay look different (e.g., freq_group_delay_hold.png). Where does it come from and was there an error before?

f-brinkmann:
> The test plots with incl. group delay look different (e.g., freq_group_delay_hold.png)

We still had auto-unit the last time the baseline plots were created. In that case the first signal was plotted in milliseconds (see that in freq_group_delay_default) and the second signal changed the time axis to seconds. Now they are both plotted in seconds.

PR #407, opened 2022-12-10
Update testing guidelines

### LINKED ISSUE #403: Improve Test Guidlines
## General

* pyfar version: 0.5.1

## Description

We should extend the testing guidelines for a minimal required set of tests. This should include

- Test if input audio classes are of the correct type (Signal, TimeData, FrequencyData)
- Test all errors and warnings
- Test single and multi-dimensional input data

- add a note for testing and generated plots

f-brinkmann:

Closes #403 

Update testing guidelines


ahms5 (dismissed):

CONTRIBUTING.rst
```diff
@@ -89,6 +89,28 @@ In the following, you'll find a guideline. Note: these instructions are not gene
...
+pyfar functions should raise errors if
+
+- audio objects do not have the correct type
+- strings have invalid values
+- invalid parameter combinations are used
+
+pyfar functions should raise warnings if
+
+- results might be wrong or unexpected
+- possibly bad parameter combinations are used
```
mberz, line 112: Doesn't this categorize as function behavior, rather than test behavior?

mberz (changes requested):
One comment: It's a bit confusing that we have guidelines on test containing guidelines on function design. Maybe we can move this to a more fitting position in the doc.

twennemann (approved):

CONTRIBUTING.rst
```diff
@@ -76,7 +76,21 @@ Before you submit a pull request, check that it meets these guidelines:
...
+pyfar functions and classes should raise errors if
+
+- audio objects do not have the correct type (e.g. a TimeData instance is passed but a Signal instance is required)
+- strings have invalid values
+- invalid parameter combinations are used
+
+pyfar functions and classes should raise warnings if
+
+- results might be wrong or unexpected
+- possibly bad parameter combinations are used
```
sikersten, line 93: I'd suggest to have it consistent within the file
```suggestion
- Audio objects do not have the correct type (e.g. a TimeData instance is passed but a Signal instance is required)
- Strings have invalid values
- Invalid parameter combinations are used

pyfar functions and classes should raise warnings if

- Results might be wrong or unexpected
- Possibly bad parameter combinations are used
```

CONTRIBUTING.rst
```diff
@@ -89,6 +103,17 @@ In the following, you'll find a guideline. Note: these instructions are not gene
 - Make sure that all important parts of pyfar are covered by the tests. This can be checked using *coverage* (see below).
 - In case of pyfar, mainly **state verification** is applied in the tests. This means that the outcome of a function is compared to a desired value (``assert ...``). For more information, it is refered to `Martin Fowler's article <https://martinfowler.com/articles/mocksArentStubs.html.>`_.
 
+Required Tests
+~~~~~~~~~~~~~~
+
+The testing should include
+
+- Test all errors and warnings (also function and class guidelines above)
```
sikersten, line 111:
```suggestion
- Test all errors and warnings (see also function and class guidelines above)
```

CONTRIBUTING.rst
```diff
@@ -76,7 +76,21 @@ Before you submit a pull request, check that it meets these guidelines:
...
+
+pyfar functions and classes should raise errors if
+
+- audio objects do not have the correct type (e.g. a TimeData instance is passed but a Signal instance is required)
+- strings have invalid values
+- invalid parameter combinations are used
+
+pyfar functions and classes should raise warnings if
+
+- results might be wrong or unexpected
```
sikersten, line 92:
```suggestion
- results might be wrong or unexpected (from the user's perspective)
```

CONTRIBUTING.rst
sikersten, line 93: What do you mean with "bad" here?

CONTRIBUTING.rst
```diff
@@ -76,7 +76,21 @@ Before you submit a pull request, check that it meets these guidelines:
...
 2. If the pull request adds functionality, the docs should be updated. Put your new functionality into a function with a docstring.
-3. If checks to not pass, have a look at https://travis-ci.com/pyfar/pyfar/pull_requests for more information.
+3. If checks do not pass, have a look at https://app.circleci.com/pipelines/github/pyfar/pyfar for more information.
+
+Function and Class Guidelines
+-----------------------------
+
+pyfar functions and classes should raise errors if
+
+- audio objects do not have the correct type (e.g. a TimeData instance is passed but a Signal instance is required)
```
sikersten, line 86: Do we or should we have guidelines on checking the type of parameters?
E.g. in case of bools, if private properties of classes are set

sikersten (commented):
Thanks, nice improvement!
I have the feeling that we can formulate the function and class guidelines a little more precise and added some ideas. I'd suggest that we discuss in a weekly, if we understand it all in the same way?

CONTRIBUTING.rst
f-brinkmann: I think I read somewhere that the general Python recommendation is to document the expected input type but not explicitly check it. This, we would already follow.

ahms5 (approved):

CONTRIBUTING.rst
```diff
@@ -76,7 +76,37 @@ Before you submit a pull request, check that it meets these guidelines:
...
+
+Function and Class Guidelines
+-----------------------------
+
+Functions and classes should
+
+* have a single clear purpose and a functionality limited to that purpose. Conditional arguments are fine in some cases but are an indicator that a function or class does not have a clear purpose. Conditional arguments are
+
+  - arguments that are obsolete if another argument is provided
+  - arguments that or necessary only if another argument is provided
```
ahms5, line 89:
```suggestion
  - arguments that are necessary only if another argument is provided
```

CONTRIBUTING.rst
```diff
@@ -76,7 +76,37 @@ Before you submit a pull request, check that it meets these guidelines:
...
+  - arguments that must have a specific value depending on other arguments
+
+* be split into multiple functions or classes if their purpose is not clear or the functionality not well limited.
```
ahms5, line 92:
```suggestion

```
if the purpose is not clear we should maybe remove it.

ahms5 (commented):
maybe we should be clear if we use ``parameter`` or ``arguments``. If I get it right in python they are called ``arguments``

mberz (approved):

CONTRIBUTING.rst
mberz: Good point

f-brinkmann: True, I removed 'if their purppose is not clear'

f-brinkmann:
> maybe we should be clear if we use parameter or arguments. If I get it right in python they are called arguments

PR #418, opened 2023-01-06
Deprecate len(Signal)

### LINKED ISSUE #412: deprecate `TimeData.__len__`?
## General

* pyfar version: 0.5.1

## Description

We have  `TimeData.__len__` that returns the number of samples of a TimeData/Signal object. I would suggest to deprecate this for the following reasons:

- We also have `TimeData.n_samples`, which is more explicit
- We do not have `FrequencyData.__len__` which is inconsistent
- `len(time_data_instance)` does not equal `len(time_data_instance.time)` which conflicts with general python use of the function.

f-brinkmann:

Closes #412 

- deprecate `len(Signal)`
- test deprecation

### DIFF
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -12,6 +12,7 @@ import deepdiff
 import numpy as np
 import pyfar.dsp.fft as fft
 from typing import Callable
+from .warnings import PyfarDeprecationWarning
 
 
 class _Audio():
@@ -823,6 +824,9 @@ class Signal(FrequencyData, TimeData):
     def __len__(self):
         """Length of the object which is the number of samples stored.
         """
+        warnings.warn(
+            ("len(Signal) wil be deprecated in pyfar 0.8.0 "
+             "Use Signal.n_samples instead"), PyfarDeprecationWarning)
         return self.n_samples
 
     def __iter__(self):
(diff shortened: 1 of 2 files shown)


pyfar/classes/audio.py
```diff
@@ -823,6 +824,9 @@ def __repr__(self):
     def __len__(self):
         """Length of the object which is the number of samples stored.
         """
+        warnings.warn(
+            ("len(Signal) wil be deprecated in pyfar 0.8.0 "
```
sikersten, line 828:
```suggestion
            ("len(Signal) will be deprecated in pyfar 0.8.0 "
```

tests/test_deprecations.py
```diff
@@ -178,3 +181,14 @@ def test_pad_zero_modi():
         with pytest.raises(ValueError):
             # remove mode 'before' and 'after' from pyfar 0.8.0!
             pf.dsp.pad_zeros(pf.Signal([1], 44100), 5, mode='before')
+
+
+def test_signal_len():
+    with pytest.warns(PyfarDeprecationWarning,
+                      match=re.escape("len(Signal) wil be deprecated")):
```
sikersten, line 188:
```suggestion
                      match=re.escape("len(Signal) will be deprecated")):
```

sikersten (approved):

pyfar/classes/audio.py
f-brinkmann: Wenn ich groß bin, kann ich irgendwann Rechtschreibung...

tests/test_deprecations.py
f-brinkmann: fixed :)

PR #425, opened 2023-01-17
NaNs in dsp.average

twennemann:
In #399 we decided

- that we also allow NaNs in dsp.average function

### DIFF
--- a/pyfar/dsp/dsp.py
+++ b/pyfar/dsp/dsp.py
@@ -1932,6 +1932,7 @@ def average(signal, mode='linear', caxis=None, weights=None, keepdims=False,
         will be squeezed after averaging. The default is ``False``.
     nan_policy: string, optional
         Define how to handle NaNs in input signal.
+
         ``'propagate'``
            If the input signal includes NaNs, the resulting averaged output
            signal values are NaN.
@@ -1940,7 +1941,8 @@ def average(signal, mode='linear', caxis=None, weights=None, keepdims=False,
         ``'raise'``
             A ``'ValueError'`` will be raised, if the input signal includes
             NaNs.
-        The default is 'raise'.
+
+        The default is ``'raise'``.
 
     Returns
     --------
--- a/pyfar/dsp/dsp.py
+++ b/pyfar/dsp/dsp.py
@@ -1933,7 +1933,7 @@ def average(signal, mode='linear', caxis=None, weights=None, keepdims=False,
     nan_policy: string, optional
         Define how to handle NaNs in input signal.
         ``'propagate'``
-           If the input signal includes NaNs, the resulting averaged output 
+           If the input signal includes NaNs, the resulting averaged output
            signal values are NaN.
         ``'omit'``
            NaNs will be omitted while averaging.
--- a/pyfar/dsp/dsp.py
+++ b/pyfar/dsp/dsp.py
@@ -1934,10 +1934,11 @@ def average(signal, mode='linear', caxis=None, weights=None, keepdims=False,
         Define how to handle NaNs in input signal.
 
         ``'propagate'``
-           If the input signal includes NaNs, the resulting averaged output
-           signal values are NaN.
+           If the input signal includes NaNs, the corresponding averaged output
+           signal value will be NaN.
         ``'omit'``
-           NaNs will be omitted while averaging.
+           NaNs will be omitted while averaging. For each NaN value, the number
+           of values used for the average division is also reduced by one.
         ``'raise'``
             A ``'ValueError'`` will be raised, if the input signal includes
             NaNs.
--- a/pyfar/dsp/dsp.py
+++ b/pyfar/dsp/dsp.py
@@ -1936,7 +1936,7 @@ def average(signal, mode='linear', caxis=None, weights=None, keepdims=False,
            If the input signal includes NaNs, the resulting averaged output 
            signal values are NaN.
         ``'omit'``
-           NaNs will be omitted while averaging.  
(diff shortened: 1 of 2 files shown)


twennemann:
As there is no weighted average function including `nan_policy` parameter and also accepting negative values, we used `np.ma.masked_array()` to create a masked array if `nan_policy = "omit"`.

pyfar/dsp/dsp.py
```diff
@@ -1929,6 +1930,19 @@ def average(signal, mode='linear', caxis=None, weights=None, keepdims=False):
         If this is ``True``, the axes which are reduced during the averaging
         are kept as a dimension with size one. Otherwise, singular dimensions
         will be squeezed after averaging. The default is ``False``.
+    nan_policy: string, optional
+        Define how to handle NaNs in input signal.
+
+        ``'propagate'``
+           If the input signal includes NaNs, the resulting averaged output
+           signal values are NaN.
```
sikersten, line 1938: This could maybe be formulated a bit more precise. Not all values in the resulting signal are NaN, only at the corresponding samples or bins, right?

pyfar/dsp/dsp.py
```diff
@@ -1929,6 +1930,19 @@ def average(signal, mode='linear', caxis=None, weights=None, keepdims=False):
...
+        ``'omit'``
+           NaNs will be omitted while averaging.
```
sikersten, line 1940: For me, it would not be directly clear, that this refers both to the numerator and denumerator in the average operation. (so average of `[1,1,np.nan]` is 2 and not 2/3). But I might be the only one...

sikersten (commented):
Thanks for implementing, looks good.
Only had some minor comments on the docstring. It may also make sense to add tests for Signal objects?

pyfar/dsp/dsp.py
```diff
@@ -1929,6 +1930,20 @@ def average(signal, mode='linear', caxis=None, weights=None, keepdims=False):
...
         will be squeezed after averaging. The default is ``False``.
+    nan_policy: string, optional
+        Define how to handle NaNs in input signal.
+
+        ``'propagate'``
+           If the input signal includes NaNs, the corresponding averaged output
+           signal value will be NaN.
+        ``'omit'``
+           NaNs will be omitted while averaging. For each NaN value, the number
+           of values used for the average division is also reduced by one.
```
f-brinkmann, line 1941: I would suggest 'average operation' instead of 'average division'.

f-brinkmann, line 1941: I would suggest operation instead of division
```suggestion
           of values used for the average operation is also reduced by one.
```

pyfar/dsp/dsp.py
```diff
@@ -1991,7 +2008,12 @@ def average(signal, mode='linear', caxis=None, weights=None, keepdims=False):
             """mode must be 'linear', 'magnitude_zerophase', 'power',
             'magnitude_phase' or 'log_magnitude_zerophase'."""
             )
-
+    # check if data includes NaNs and raise error or create masked array
+    if nan_policy == 'raise' and True in np.isnan(data):
+        raise ValueError("The signal includes NaNs. Change 'nan_policy' to "
+                         "'propagate' or 'omit'.")
+    elif nan_policy == 'omit' and True in np.isnan(data):
```
f-brinkmann, line 2015:
```suggestion
    elif nan_policy == 'omit' and np.any(np.isnan(data)):
```

tests/test_dsp_average.py
```diff
@@ -59,24 +60,47 @@ def test_keepdims_parameters():
     assert len(signal.cshape) == len(ave2.cshape)
 
 
+@pytest.mark.parametrize('data', (
+                        pf.TimeData([[1, np.nan], [1, 2]], [1, 2]),
+                        pf.FrequencyData([[1, np.nan], [1, 2]], [1, 2])))
+def test_nan_value_averaging(data):
+    # Test average with data including NaNs.
+    if data.domain == 'time':
```
f-brinkmann, line 68: the if case could be avoided if you pass 'time' and 'freq' as part of parametrize and get the data by `getattr(signal, 'time')` or `getattr(signal, 'freq')`

pyfar/dsp/dsp.py
```diff
@@ -1929,6 +1930,20 @@ def average(signal, mode='linear', caxis=None, weights=None, keepdims=False):
         If this is ``True``, the axes which are reduced during the averaging
         are kept as a dimension with size one. Otherwise, singular dimensions
         will be squeezed after averaging. The default is ``False``.
+    nan_policy: string, optional
+        Define how to handle NaNs in input signal.
+
+        ``'propagate'``
+           If the input signal includes NaNs, the corresponding averaged output
+           signal value will be NaN.
+        ``'omit'``
```
f-brinkmann, line 1939: I would suggest to add: "If a signal contains only NaN values in a specific dimensions, the output will be zero. For example if the second sample of a multi channel signal is allways NaN, the average will be zero at the second sample"

See this example:
```python
import pyfar as pf
import numpy as np

time = pf.TimeData(np.array([[1, np.nan, 1], [1, np.nan, 1]]), [1, 2, 3])
avg = pf.dsp.average(time, nan_policy="omit")
avg.time # yields [[1, 0, 1]]
```

f-brinkmann (changes requested):
Thanks for the effort I only had minor comments on the docstring and code style

sikersten (approved):

f-brinkmann (approved):

PR #430, opened 2023-02-02
Ignore axis in broadcasting

twennemann:
We decided:

- that we want to have an `ignore_axis` parameter in `utils.broadcast_cshape` which will be used in #65.

### DIFF
--- a/pyfar/utils.py
+++ b/pyfar/utils.py
@@ -62,7 +62,7 @@ def broadcast_cshapes(signals, cshape=None, ignore_caxis=None):
         if not isinstance(signal, (pf.Signal, pf.TimeData, pf.FrequencyData)):
             raise TypeError("All input data must be pyfar audio objects")
     if cshape is not None and ignore_caxis is not None:
-        raise ValueError(f"Use ignore_axis = 'None' to broadcast to cshape ="
+        raise ValueError(f"Use ignore_caxis = 'None' to broadcast to cshape ="
                          f" {cshape}.")
     if cshape is None:
         if ignore_caxis is not None:
--- a/pyfar/utils.py
+++ b/pyfar/utils.py
@@ -67,7 +67,12 @@ def broadcast_cshapes(signals, cshape=None, ignore_caxis=None):
                          f" {cshape}.")
     if cshape is None:
         if ignore_caxis is not None:
+            # Adjust cshapes to largest dimention
             data_shapes = [s.cshape for s in signals]
+            max_dim = np.max([len(sh) for sh in data_shapes], axis=0)
+            for i, sh in enumerate(data_shapes):
+                for _ in range(max_dim-len(sh)):
+                    data_shapes[i] = (1,) + data_shapes[i]
             # Finds broadcast cshape without the axis to ignore
             cshape = np.broadcast_shapes(*np.delete(data_shapes, ignore_caxis,
                                          axis=-1))
--- a/pyfar/utils.py
+++ b/pyfar/utils.py
@@ -48,6 +48,9 @@ def broadcast_cshapes(signals, cshape=None, ignore_axis=None):
         The cshape to which the signals are broadcasted. If `cshape` is
         ``None`` it is determined from the cshapes of the input signals using
         ``numpy.broadcast_shapes``. The default is ``None``.
+    ignore_axis : int, optional
+        Axis which will be ignore while broadcasting. Has to be ``None``, if
+        `cshape` is not ``None``. The default is ``None``.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


pyfar/utils.py
```diff
@@ -48,6 +48,9 @@ def broadcast_cshapes(signals, cshape=None):
         The cshape to which the signals are broadcasted. If `cshape` is
         ``None`` it is determined from the cshapes of the input signals using
         ``numpy.broadcast_shapes``. The default is ``None``.
+    ignore_caxis : int, optional
+        Axis which will be ignore while broadcasting. Has to be ``None``, if
+        `cshape` is not ``None``. The default is ``None``.
```
f-brinkmann, line 53: Lets add a link to the concepts and explain the default behavior:
```suggestion
        Channel axis (:py:mod:`caxis <pyfar._concepts.audio_classes>`) which will be ignore during broadcasting.
        Must be ``None``, if `cshape` is not ``None``. The default is ``None``, which broadcasts all caxis.
```

pyfar/utils.py
```diff
@@ -58,9 +61,29 @@ def broadcast_cshapes(signals, cshape=None):
...
-
+    if cshape is not None and ignore_caxis is not None:
+        raise ValueError(f"Use ignore_caxis = 'None' to broadcast to cshape ="
+                         f" {cshape}.")
     if cshape is None:
-        cshape = np.broadcast_shapes(*[s.cshape for s in signals])
+        if ignore_caxis is not None:
+            data_shapes = [s.cshape for s in signals]
+            # Finds broadcast cshape without the axis to ignore
+            cshape = np.broadcast_shapes(*np.delete(data_shapes, ignore_caxis,
```
f-brinkmann, line 71: Should we make sure that all cshapes have the same length before we do this by calling `pyfar.utils.match_cdims` beforehands? Otherwise the behavior would be hard to predict if signals with different `cdims` would be passed. If we do this we could add to the docstring of `caxis`: caxis refers to the channel axis after broadcasting the input signals to the same `cdim` by calling :py:func:`~pf.utils.broadcast_cdims`.

tests/test_utils.py
```diff
@@ -40,12 +41,30 @@ def test_broadcast_cshapes(cshape, reference):
         assert isinstance(broadcast, type(signal))
 
 
+@pytest.mark.parametrize("shape1, shape2, caxis, reference1, reference2", [
+    ((1, 2, 3, 1), (1, 1, 2, 1), -1, (1, 2, 3), (1, 2, 2)),
```
f-brinkmann, line 45: would be easier to understand if you passed the cshapes and added the number of samples like this
`pf.Signal(np.ones(cshape1 + (n_samples, )), 44100)`

f-brinkmann (changes requested):
Thanks, I found one thing that we might want to discuss in the weekly and minor comments on the docstring and testing

pyfar/utils.py
```diff
@@ -58,9 +62,34 @@ def broadcast_cshapes(signals, cshape=None):
...
         if not isinstance(signal, (pf.Signal, pf.TimeData, pf.FrequencyData)):
             raise TypeError("All input data must be pyfar audio objects")
-
+    if cshape is not None and ignore_caxis is not None:
+        raise ValueError(f"Use ignore_caxis = 'None' to broadcast to cshape ="
+                         f" {cshape}.")
     if cshape is None:
-        cshape = np.broadcast_shapes(*[s.cshape for s in signals])
+        if ignore_caxis is not None:
+            # Adjust cshapes to largest dimention
```
f-brinkmann, line 70:
```suggestion
            # Adjust cshapes to largest dimension
```

pyfar/utils.py
```diff
@@ -58,9 +62,34 @@ def broadcast_cshapes(signals, cshape=None):
...
+            data_shapes = [s.cshape for s in signals]
+            max_dim = np.max([len(sh) for sh in data_shapes], axis=0)
```
f-brinkmann, line 72: and `max_dim` to `max_cdim` (sorry for being so picky)

pyfar/utils.py
```diff
@@ -48,6 +48,10 @@ def broadcast_cshapes(signals, cshape=None):
         The cshape to which the signals are broadcasted. If `cshape` is
         ``None`` it is determined from the cshapes of the input signals using
         ``numpy.broadcast_shapes``. The default is ``None``.
+    ignore_caxis : int, optional
+        Channel axis (:py:mod:`caxis <pyfar._concepts.audio_classes>`) which
+        will be ignore during broadcasting. Must be ``None``, if `cshape` is
+        not ``None``. The default is ``None``, which broadcasts all caxis.
```
f-brinkmann, line 54: Maybe add "Note that `caxis` refers to the channel axis after :py:func:`~broadcast.cdims` is applied".

tests/test_utils.py
```diff
@@ -40,12 +41,30 @@ def test_broadcast_cshapes(cshape, reference):
         assert isinstance(broadcast, type(signal))
 
 
+@pytest.mark.parametrize("shape1, shape2, n_samples, caxis, ref1, ref2", [
```
f-brinkmann, line 44: would suggest to rename the parameters to `cshape1` and `cshape2` to make it more clear. Is n_samples really needed as a variable? If not the test could always use the same number of samples and the variable could be removed.

f-brinkmann (changes requested):
Thanks. I have only minor comments left

f-brinkmann:
We decided that the new functionality should rather be moved to the concatenate function

f-brinkmann:
now handled in #452

PR #433, opened 2023-02-03
Bug/single frequency signals

### LINKED ISSUE #360: Audio Classes do not allow data for a single time/frequency
## General

but they should

f-brinkmann:

Closes #360 

- fix `Signal.__init__`, `Signal.freq`, and `Signal.freq_raw` for the case of passing single frequency data but not the number of samples
- Reduce code redundancy in `Signal.freq`, and `Signal.freq_raw` though shared functions `Signal._freq`

### DIFF
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -654,24 +654,7 @@ class Signal(FrequencyData, TimeData):
     @freq.setter
     def freq(self, value):
         """Set the normalized frequency domain data."""
-        # check data type
-        data = np.atleast_2d(np.asarray(value))
-        if data.dtype.kind not in ["i", "f", "c"]:
-            raise ValueError((f"frequency data is {data.dtype} must be int, "
-                              "float, or complex"))
-        # Check n_samples
-        if data.shape[-1] != self.n_bins:
-            self._n_samples = max(1, (data.shape[-1] - 1)*2)
-            warnings.warn(
-                f"Number of samples not given, assuming {self.n_samples} "
-                f"samples from {data.shape[-1]} frequency bins.")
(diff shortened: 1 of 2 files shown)


pyfar/classes/audio.py
```diff
@@ -614,9 +614,9 @@ def __init__(
             # check and set n_samples
             if n_samples is None:
                 warnings.warn(
-                    "Number of time samples not given, assuming an even "
-                    "number of samples from the number of frequency bins.")
-                n_samples = (data.shape[-1] - 1)*2
+                    f"Number of samples not given, assuming {n_samples} "
+                    f"samples from {data.shape[-1]} frequency bins.")
+                n_samples = max(1, (data.shape[-1] - 1)*2)
```
tluebeck, line 619: Does it make sense to write {n_samples} in the f-string here? n_samples is None. I suggest to print soemthing like "Number of samples not given, assuming the number of samples from {data.shape[-1]} frequency bins."

tluebeck (commented):
Everything looks fine to me. Just one minor comment. 

But I need to update it in the complex_time_signal branch.

pyfar/classes/audio.py
sikersten: Good point, could be solved by changig the order of the warning and `n_samples=...`, right?

sikersten (approved):

pyfar/classes/audio.py
f-brinkmann: Thanks for noting. Changed the order as suggested to be able to output the assumed number of samples

tluebeck (approved):

PR #396, opened 2022-11-22
add complex argument to TimeData

tluebeck:
### Changes proposed in this pull request:

- allow time data to be complex valued
- chose FFT accordinlgy, and wrap fft and ifft for two-sided Fourier transform

PR #435, opened 2023-02-10
Enforce numeric data in signals

### LINKED ISSUE #431: Calculations for Signal containing NaN elements return only NaN.
If a Signal object including NaN values is used for a calculation, every value will be returned with NaN. Example:
`calc = pf.Signal([1, 2, np.nan], 44100) / 2`
I would expect that `calc.time == [0.5, 1, nan]` but the actual result is `calc.time == [nan, nan, nan]`.

f-brinkmann:

Closes #431, requires #436 / #453

- Check for non-numerical data in `Signal.__init__`

I did not find a way to check for boolean data, because it is converted to float when calling `np.atleast_2d(data)`. Any ideas?

### DIFF
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -601,13 +601,9 @@ class Signal(FrequencyData, TimeData):
             raise ValueError(("Invalid FFT normalization. Has to be "
                               f"{', '.join(self._VALID_FFT_NORMS)}, but "
                               f"found '{fft_norm}'"))
-        # time / normalized frequency data (depending on domain)
-        data = np.atleast_2d(data)
         # check input
-        if data.dtype.kind not in ["u", "i", "f", "c"] or \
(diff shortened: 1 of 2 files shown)


ahms5 (dismissed):

sikersten (dismissed):

mberz (changes requested):
This actually needs to checked in the setter methods of Signal as well.

f-brinkmann:
Good point, I added it to the freq and freq_raw setter so far. I think the time setter is refactored in #396. Should we rather add this as todo there to avoid conflicts? The check is now a static class method so it would be easy to include.

f-brinkmann:
We decided to postpone this until #396 is merged

f-brinkmann:
Closed in favor of #505 which was done for the new Audio classes that can handle complex time data.

PR #436, opened 2023-02-10
Interpolate spectrum on logarithmic frequency axis

f-brinkmann:

`dsp.InterpolateSpectrum` resulted in Signals containing NaN values if interpolating on a log. frequency axis. The problem was caused by passing two identical frequencies to the interpolation object. This is now fixed by a better choice for f=0 Hz, which has to be treated because log(0) = -inf. The fix is required because NaN's are not allowed anymore (#435).

- Replace 0 Hz with small positive value
- add test for output value

### DIFF
--- a/pyfar/dsp/interpolation.py
+++ b/pyfar/dsp/interpolation.py
@@ -895,12 +895,14 @@ class InterpolateSpectrum():
         """
         Return frequencies for creating or quering interpolation objects.
 
-        In case logfrequencies are requested, 0 Hz entries are replaced by
-        the next highest frequency, because the logarithm of 0 does not exist.
+        In case logfrequencies are requested, 0 Hz can not be used, because the
+        logarithm of 0 does not exist. 0 Hz is replaced with a frequency close
+        (but not too close) to 0 Hz to avoid numerical issues during the
+        interpolation.
         """
         if self._fscale == "log":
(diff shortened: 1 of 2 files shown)


ahms5 (dismissed):

sikersten (dismissed):

ahms5 (dismissed):

mberz (approved):
Looks good to me. 
Maybe we can add an example for the log interpolation and a simple test for interpolation of logarithmically spaced frequency data? This could also be done in a separate PR.

f-brinkmann:
The docs contain an example for both interpolation types: https://pyfar.readthedocs.io/en/stable/modules/pyfar.dsp.html#pyfar.dsp.InterpolateSpectrum.

The tests are a bit distributed, but I hope everything is covered
- Interpolators are tested in tests/test_dsp_interpolation.py::test_interpolate_spectrum_interpolation
- and the frequencies used for interpolation are tested in tests/test_dsp_interpolation.py::test_interpolate_spectrum_fscale

The interpolation on a log scale uses the same interpolators as tested in the first test. So I think an additional test is not needed.

mberz:
The docs only cover plotting the same example on a log frequency axis and a linear frequency axis. The interpolator is only called once in the beginning of the example.

PR #443, opened 2023-02-21
Reading frequency domain signals

### LINKED ISSUE #442: Reading frequency domain signals fails
## General

* pyfar version: 0.5.2

## Description

Time domain pyfar Signal objects can be written and read with `io.write` and `io.read` but this fails for frequency domain Signal objects. Example

```python
import pyfar as pf
import deepdiff

signal = pf.signals.impulse(4, 2)
signal.domain = "freq"

pf.io.write("test.far", signal=signal)
signal_read = pf.io.read("test.far")["signal"]
```

The problem is that `Signal._decode` always assumes time domain data. It should be solved by forcing a full initialization in decode, i.e., add
```python
obj_dict['_domain'],
obj_dict['_fft_norm']
```

twennemann:

Closes #442

-  `Signal._decode` always assumed time domain data. So we added
```python
obj_dict['_domain'],
obj_dict['_fft_norm']
```
- Signals initialized `self._n_bins = data.shape[-1]` if `domain =='freq'`, but it was never used.
- parametrized test for read and write a Signal

### DIFF
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -782,8 +782,9 @@ class Signal(FrequencyData, TimeData):
 
     def _encode(self):
         """Return dictionary for the encoding."""
-        self.domain = "time"
-        class_dict = self.copy().__dict__
+        selfcopy = self.copy()
+        selfcopy.domain = "time"
+        class_dict = selfcopy.__dict__
         return class_dict
 
     @classmethod
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -611,7 +611,6 @@ class Signal(FrequencyData, TimeData):
                 raise ValueError(("n_samples can not be larger than "
                                   "2 * data.shape[-1] - 2"))
             self._n_samples = n_samples
-            self._n_bins = data.shape[-1]
             # Init remaining parameters
             FrequencyData.__init__(self, data, self.frequencies, comment)
             delattr(self, '_frequencies')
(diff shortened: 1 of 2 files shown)


f-brinkmann (approved):
thanks!

mberz:
I'm actually not so fond of the idea to save frequency domain data for signals. It's only valid as long as the norm is known and does not change or is deprecated in future versions. I'd prefer to save in the time domain to avoid this.
It's also far easier to read the data from implementations other than pyfar, which may not implement pyfar spectrum norms.

f-brinkmann:
> It's only valid as long as the norm is known and does not change or is deprecated in future versions.

That's true, and would be easy to change. While it would increase the odds of the data being written correctly, the Signal might behave differently as intended when the data was written. I can live with both choices.

pyfar/classes/audio.py
```diff
@@ -781,6 +780,12 @@ def _return_item(self, data):
                       fft_norm=self.fft_norm, comment=self.comment)
         return item
 
+    def _encode(self):
+        """Return dictionary for the encoding."""
+        self.domain = "time"
+        class_dict = self.copy().__dict__
```
sikersten, line 786: I don't know if it is crucial, but it might make sense to change the domain of the copy instead? Something like
```suggestion
        selfcopy = self.copy()
        selfcopy.domain = "time"
        class_dict = selfcopy.__dict__
```

PR #445, opened 2023-02-23
catch reading outdated pyfar objects

### LINKED ISSUE #444: Reading far data written with old pyfar version will fail in the future
## General

* pyfar version: 0.5.2

## Description

After deprecating the old coordinates Class in 0.8.0, new pyfar Version will fail in reading old Coordinate objects. I suggest the following as a trade-off between usability and effort

- write the pyfar version to each .far file
- always read the version first
- everything else should be encapsulated in a `try` statement
- if reading the data fails, an error with a hint to the pyfar version that was used to save the data is raised

f-brinkmann:

Closes #444 

- write and read used pyfar version to .far files
- output pyfar version that was used for writing data if reading the data fails
- add testing


pyfar/io/io.py
```diff
@@ -231,24 +232,51 @@ def read(filename):
...
-                    raise TypeError(
-                        '.far-file contains unknown types.'
-                        'This might occur when writing and reading files with'
-                        'different versions of Pyfar.')
-                collection[name] = obj
+                    collection[name] = obj
+
+            # check version (writing the version was introduced in 0.5.2)
+            if pyfar_version is None:
+                pyfar_version = "<0.5.2"
```
sikersten, line 256:
```suggestion
            # check version (writing the version was introduced in 0.5.3)
            if pyfar_version is None:
                pyfar_version = "<0.5.3"
```

sikersten (approved):
Looks good, thanks for noting and taking care!

pyfar/io/io.py
f-brinkmann: Yes thanks

f-brinkmann:
There are still some tests failing. I think only due to patches that have to be extended for the new cases. But I'll check that after we agreed on a final solution...

sikersten (approved):

ahms5 (approved):

PR #450, opened 2023-03-16
Remove unwanted minor tick label for log plots

### LINKED ISSUE #71: Log formatter  for plots adds unwanted ticks in scientific notation
## Description

`pyfar.plot.ticker.Log*.py` add unwanted ticks depending on the axis range

## What I Did

```
import numpy as np
import pyfar.plot as pp

# this is fine
sin = np.sin(np.linspace(0, 2000*np.pi, 10000))
x = Signal(sin, 44100, fft_norm='amplitude')
ax = pp.line.freq(x)

# this generates an unwanted tick at 300 Hz
ax.set_xlim(50, 425)
```

pingelit:

Closes #71

- Disable the minor tick labels for log axis to remove the unwanted tick labels.

### DIFF
--- a/pyfar/plot/_line.py
+++ b/pyfar/plot/_line.py
@@ -156,6 +156,7 @@ def _phase(signal, deg=False, unwrap=False, freq_scale='log', ax=None,
     # set and format ticks
     if freq_scale == 'log':
         ax.xaxis.set_major_locator(LogLocatorITAToolbox())
+        ax.xaxis.set_minor_formatter(NullFormatter())
     ax.xaxis.set_major_formatter(LogFormatterITAToolbox())
 
     return ax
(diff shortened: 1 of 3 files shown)


pingelit:
So far I've only added this to the `freq` plot, due to some open questions:

- Is this what we want? #71 suggests another approach.
- Should this be the default, even for `freq_scale='linear'`

f-brinkmann:
Thanks for checking, I think it resolves the issue. I did not see any effect on linear axes. Applying it only to log. frequency axis might be the cleaner solution as it would maintain as much Matplotlib default behavior as possible.

mberz (commented):
Just minor things.
I'd also agree to not use this for linearly scaled axes.
We should not close #71 though, or create a new issue, since this only fixes part of the issue. The other problem is that one may end up with a very low number (or on extreme cases none) of major ticks when zooming in on the frequency axis.

f-brinkmann:
Agreed and opened a new issue with a more specific name #451

pingelit:
Just fixed the comments from @mberz.
Before opening the pull request, I want to check if this behavior can be unit tested.

f-brinkmann (approved):

f-brinkmann (approved):
Thanks for implementing!

PR #452, opened 2023-03-29
Add concatenate function

twennemann:

Closes #

- Add a concatenation function, concatenating Signals along specified channels.
- Add broadcasting parameter, which broadcasts the signals into largest cdim and a common shape, ignoring caxis. 

- [x] Please rename to concatenate_channles

Otherwise approved

### DIFF
--- a/pyfar/utils.py
+++ b/pyfar/utils.py
@@ -131,13 +131,7 @@ def broadcast_cdims(signals, cdim=None):
 
 def concatenate(signals, caxis=0, broadcasting=False, comment=""):
     """
-    Merge multiple Signal objects along a given axis. The signals are copied,
-    if needed broadcasted in largest dimenson and to a common cshape and a new
-    object is returned.
-
-    The :py:mod:`cshape <pyfar._concepts.audio_classes>` of the signals are
-    broadcasted following the `numpy broadcasting rules
-    <https://numpy.org/doc/stable/user/basics.broadcasting.html>`_
+    Merge multiple Signal objects along a given caxis.
 
     Parameters
     ----------
@@ -147,12 +141,18 @@ def concatenate(signals, caxis=0, broadcasting=False, comment=""):
         to axis (the first, by default). If this is the case, set
         ``broadcasting=True``.
     caxis : int
-        The axis along which the signals are concatenated.
+        The caxis along which the signals are concatenated. More details and
+        background about caxis is given in the concepts of
+        :py:mod:`Audio classes <pyfar._concepts.audio_classes>`.
         The default is ``0``.
     broadcasting: bool
-        If this is ``True``, the signals will be broadcasted into largest
-        dimension and into a common cshape, except of the caxis along which the
-        signals are concatenated. The default is ``False``.
+        If this is ``True``, the signals will be broadcasted to common
+        cshape, except for the caxis along which the signals are
+        concatenated.
+        The :py:mod:`cshape <pyfar._concepts.audio_classes>` of the signals are
+        broadcasted following the `numpy broadcasting rules
+        <https://numpy.org/doc/stable/user/basics.broadcasting.html>`_
+        The default is ``False``.
     comment: string
         A comment related to the merged `data`. The default is ``""``, which
         initializes an empty string. Comments of the input signals will also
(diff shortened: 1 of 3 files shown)


twennemann:
I tried to adjust the comment of the merged signal so that the comments can be assigned to the data in the merged signal. First, to the comment of the merged signal is passed on which caxis the signals were merged. If there is a comment in one of the input Signals, that comment is appended, as well as the corresponding channels on caxis.
Questions:
- Is it right to call it channels?
- Do we start counting at 0 or at 1? For example `first_signal.cshape[caxis]` is 3, would we say the comment corresponds to channel 1-3 or 0-2?

pyfar/utils.py
```diff
@@ -127,3 +127,92 @@ def broadcast_cdims(signals, cdim=None):
...
+        The default is ``0``.
+    broadcasting: bool
+        If this is ``True``, the signals will be broadcasted into largest
+        dimension and into a common cshape, except of the caxis along which the
+        signals are concatenated. The default is ``False``.
+    comment: string
+        A comment related to the merged `data`. The default is ``""``, which
+        initializes an empty string. Comments of the input signals will also
+        be returned in the new Signal object. These comments are marked with
+        their corresponding signal position in the input tuple.
```
f-brinkmann, line 160: I would suggest to simply the comments handling. Use an empty string if the provided comment. Discard comments from input data. What do the others think?

pyfar/utils.py
```diff
@@ -127,3 +127,92 @@ def broadcast_cdims(signals, cdim=None):
...
+        cshapes = [s.cshape for s in signals]
+        max_cdim = np.max([len(sh) for sh in cshapes])
+        for i, sh in enumerate(cshapes):
+            for _ in range(max_cdim-len(sh)):
+                cshapes[i] = (1,) + cshapes[i]
+        # Finds broadcast cshape without the caxis to ignore
+        cshape = np.broadcast_shapes(*np.delete(cshapes, caxis,
+                                     axis=-1))
+        broad_signals = []
+        for i, s in enumerate(signals):
```
f-brinkmann, line 188: using
```suggestion
        for signal, cshape in zip(signals, cshapes):
```
might be better to read and avoid indexing

pyfar/utils.py
```diff
@@ -127,3 +127,92 @@ def broadcast_cdims(signals, cdim=None):
...
+    if comment != "":
+        comment = comment + '\n'
+    comment = comment + f'Signals concatenated in caxis={caxis}.\n'
+    sig_channel = 0
+    for s in signals:
+        if s.comment != "":
+            comment += f"Channel {sig_channel+1}-"\
+                       f"{sig_channel + s.cshape[caxis]}: "\
+                       + s.comment + '\n'
+        sig_channel += s.cshape[caxis]
```
f-brinkmann, line 212: See above. Should we simplify this?

f-brinkmann (changes requested):

pyfar/utils.py
```diff
@@ -131,15 +131,16 @@ def broadcast_cdims(signals, cdim=None):
...
     ----------
-    signals : tuple of Signal
-        The signals to concatenate. Must either have the same cshape or be
-        broadcastable to the same cshape, except in the dimension corresponding
-        to axis (the first, by default). If this is the case, set
-        ``broadcasting=True``.
+    signals : tuple of Signal, TimeData or FrequencyData
+        The signals to concatenate. All signals must be of the same object type
+        and either have the same cshape or be broadcastable to the same cshape,
+        except in the dimension corresponding to axis (the first, by default).
```
f-brinkmann, line 142: Better to use `caxis` instead of axis.

f-brinkmann (approved):
Thanks for implementing. Approved and awaiting a discussion on the comment handling...

pyfar/utils.py
```diff
@@ -127,3 +127,80 @@ def broadcast_cdims(signals, cdim=None):
...
+    """
+    # check input
+    for signal in signals:
+        if not isinstance(signal, (pf.Signal, pf.TimeData, pf.FrequencyData)):
+            raise TypeError("All input data must be of type pf.Signal, "
+                            "pf.TimeData or pf.FrequencyData.")
+    if not isinstance(broadcasting, bool):
+        raise TypeError("'broadcasting' needs to be False or True.")
+    # check matching meta data of input signals.
+    [signals[0]._assert_matching_meta_data(s) for s in signals]
```
f-brinkmann, line 170: Just a tiny note - I think you can pass all audio objects at the same time
```suggestion
    _assert_matching_meta_data(signals)
```

f-brinkmann, line 170: Just a tiny note - I think you can pass all audio objects at the same time
```suggestion
    _assert_matching_meta_data(signals)
```

f-brinkmann (changes requested):
I just recognized that this would need small changes to work for complex signals. But I would suggest to make this a separate pull.
- the complex variable must be obtained from `_assert_matching_meta_data` which requires #464  
- before concatenating `s._data` the domain must be set to `time` if the input data contains a complex `Signal` object.
- the `complex` flag must be set when returning `Signal` or `TimeData` objects

f-brinkmann (approved):

ahms5 (changes requested):
Thank you for implementing, i just have a few comments on the usage und docstring

pyfar/utils.py
```diff
@@ -127,3 +127,80 @@ def broadcast_cdims(signals, cdim=None):
...
+
+
+def concatenate(signals, caxis=0, broadcasting=False):
+    """
+    Merge multiple Signal, Timedata or Frequencydata objects along a given
+    caxis.
+
+    Parameters
+    ----------
+    signals : tuple of Signal, TimeData or FrequencyData
```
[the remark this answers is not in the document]
f-brinkmann: good to have, but should be another function

f-brinkmann (approved):

PR #460, opened 2023-04-28
Implement filter tests for complex signals

tluebeck:
unit testing for filtering complex signals


f-brinkmann:
We might get away with fewer tests if we test only if the three filter classes FilterFIR, FilterIIR, and FilterSOS can handle complex input. If this is the case, all pyfar filter functions will be good, as far as I understand.

tests/test_filter_classes.py
```diff
@@ -249,6 +273,24 @@ def test_filter_sos_process(impulse):
     npt.assert_allclose(res.time, np.atleast_2d(desired))
 
 
+def test_filter_sos_process_complex(impulse_complex):
+    sos = np.array([[1, 1/2, 0, 1, 0, 0]])
+    filt = fo.FilterSOS(sos, impulse_complex.sampling_rate)
+    coeff = np.array([[1, 1/2, 0], [1, 0, 0]])
```
f-brinkmann, line 279: Why are you initializing twice? According to the docs the shape of the coefficients should be `(1, 1, 6)` did that not work?

f-brinkmann (changes requested):
Thanks for the effort. Only had a tiny question

tests/test_filter_classes.py
tluebeck: It was a copy and paste from the test for the non complex data, to be honest :) ... I removed. Also in the non complex test. Should I also test it with coefficients (1, 1, 6)?

f-brinkmann: I tried it and filt.coefficients have shape `(1,1,6)` in both cases. but maybe inputting the coefficients according to the docs is less confusing, i.e., using `sos = np.array([[[1, 1/2, 0, 1, 0, 0]]])`

f-brinkmann (approved):

tests/test_filter.py
```diff
@@ -79,6 +79,7 @@ def test_chebyshev2(impulse):
         x = pfilt.chebyshev2(None, 2, 40, 1000, 'lowpass')
 
 
+
```
ahms5, line 82:
```suggestion
```

ahms5 (changes requested):
thank you for implementing

PR #462, opened 2023-04-28
Raise errors in energy, power, and rms for complex signals

tluebeck:
Raise errors in energy, power, and rms for complex signals


f-brinkmann (approved):
Thanks for the effort

pyfar/dsp/dsp.py
```diff
@@ -1846,12 +1850,16 @@ def power(signal):
...
-    RMS.
+    RMS. The power of complex-valued signals is not defined.
     """
     # check input
     if not isinstance(signal, pyfar.Signal):
         raise ValueError(f"signal is type '{signal.__class__}'"
                          " but must be of type 'Signal'.")
+
+    if signal.complex:
+        raise ValueError("The power of complex signals is not defined")
```
ahms5, line 1861: same question, not defined or not implemented?

pyfar/dsp/dsp.py
```diff
@@ -1814,6 +1815,9 @@ def energy(signal):
     if not isinstance(signal, pyfar.Signal):
         raise ValueError(f"signal is type '{signal.__class__}'"
                          " but must be of type 'Signal'.")
+    if signal.complex:
+        raise ValueError("The energy of complex signals is not defined")
```
ahms5, line 1819: see below

f-brinkmann:
@tluebeck we decided to change the error messages to '... is not implemented'  instead of '... is not defined'. We might need this at a later point, if we want to compute energy for ambisonics signals.

sikersten:
btw, the Matlab function `rms` operates on the absolute values. Maybe it is indeed a matter of how to define it, which indicates the "not implemented" to be the better choice here

pingelit (approved):

f-brinkmann (approved):
Thanks for the effort. It's looking fine to me.

pingelit (approved):

pyfar/dsp/dsp.py
```diff
@@ -1803,7 +1803,8 @@ def energy(signal):
     Due to the calculation based on the time data, the returned energy is
     independent of the signal's ``fft_norm``.
     :py:func:`~pyfar.dsp.power` and :py:func:`~pyfar.dsp.rms` can be used
-    to compute the power and the rms of a signal.
+    to compute the power and the rms of a signal. The energy of complex-valued
+    signals is not implemented.
```
ahms5, line 1807:
```suggestion
    to compute the power and the rms of a signal. 

```

ahms5, line 1807:
```suggestion
    to compute the power and the rms of a signal.
```

pyfar/dsp/dsp.py
```diff
@@ -1846,14 +1848,15 @@ def power(signal):
     independent of the signal's ``fft_norm``.
     The power equals the squared RMS of a signal. :py:func:`~pyfar.dsp.energy`
     and :py:func:`~pyfar.dsp.rms` can be used to compute the energy and the
-    RMS.
+    RMS. The power of complex-valued signals is not defined.
```
ahms5, line 1851:
```suggestion
    RMS.
```

pyfar/dsp/dsp.py
```diff
@@ -1988,7 +1992,7 @@ def average(signal, mode='linear', caxis=None, weights=None, keepdims=False,
                                         'power',)):
```
ahms5, line 1992:
```suggestion
                                        )):
```
I'm not sure we need this error. I think all these things should work the same way for complex signals. Especially since the first 3 work on frequency data. And ``'power'' is defined for complex time data anyway.

ahms5 (changes requested):
Thanks for the implementation, I found some leftovers from the previous implementation.

pyfar/dsp/dsp.py
```diff
@@ -1988,7 +1990,7 @@ def average(signal, mode='linear', caxis=None, weights=None, keepdims=False,
...
 
         raise ValueError((
-            f"mode '{mode}' is not defined for complex signals."))
+            f"mode '{mode}' is not implemented for complex signals."))
```
ahms5, line 1993: Maybe you can dicscuss it on Friday, I'm not sure, but I think its not nessesary, since all mentioned options are also working for complex signals...

ahms5 (changes requested):

pyfar/dsp/dsp.py
tluebeck: Sorry for that. I already removed that line the previous commit, but something went wrong. Same with the leftovers you found the last time...
You are right, we don't need this errors anymore.

ahms5 (approved):

PR #464, opened 2023-05-11
check and test audio arithmetic for complex-valued data

tluebeck:
adapt _assert_match_for_arithmetic and introduce variable complex


f-brinkmann:
I think we should discuss how we handle mixed input. Currently the following is working and returns a complex signal
```python
import pyfar as pf

a = pf.Signal([1, 2, 3], 44100, complex=True)
b = pf.Signal([1, 2, 3], 44100, complex=False)

c = pf.add((a, b), domain='time')
```

which I think is good. However, the equivalent frequency domain operation
```python
c = pf.add((a, b), 'freq')
```

does not work, because the spectra have different shapes (single vs. both-sided). My thought is that mixed operations should work and always return complex data. This would require casting to double sided spectra for frequency domain operations.

In case we want this, should the casting be done
- somewhere in the arithmetic function,
- be a function of the `pyfar.dsp.fft` module,
- or should we add setter functionality to `Signal.complex`?

f-brinkmann:
We discussed the following in the weekly
- The complex property from TimeData and Signal should become setter functionality
- Setting from `complex = True` to `complex=False` requires a check if the time data has `abs(imag(time)) < tolerance` or if the frequency data is conjugated symmetric around 0 Hz

pyfar/classes/audio.py
```diff
@@ -1594,14 +1608,19 @@ def _get_arithmetic_data(data, domain, cshape, matmul, audio_type):
         Signal. `np.asarray(data)` otherwise.
     """
     if isinstance(data, (Signal, TimeData, FrequencyData)):
+        data_ = data.copy()
+        # check if complex casting of any input signal is necessarry
+        if complex:
+            if isinstance(data_, (Signal, TimeData)) and not data_.complex:
+                data_.complex = True
```
ahms5, line 1615: i think in PR #396 we dont allow a setter

ahms5 (changes requested):
Thank you for implementing, looks fine, i just have one comment

pyfar/classes/audio.py
f-brinkmann: I think we changed that as we noted the need for it. There are now checks that make sure `complex = False` can only be set if the spectrum is symmetric and 0 Hz and Nyquist are real valued.

pyfar/classes/audio.py
```diff
@@ -1555,15 +1565,17 @@ def _assert_match_for_arithmetic(data: tuple, domain: str, division: bool,
...
                 raise ValueError(
                     "Input must be of type Signal, int, float, or complex")
-            if np.asarray(d).dtype.kind == "c" and domain == 'time':
-                raise ValueError(
-                    "Complex input can not be applied in the time domain.")
+            if audio_type == (Signal or TimeData):
+                # if the first data was a Signal or TimeData set
+                # complex flag if the non signal input is complex-valued
+                if np.asarray(d).dtype.kind == "c":
+                    complex = True
```
f-brinkmann, line 1572: I think there are two things that might break here:

- This is in the loop over the input argument `data`, hence `audio_type` can not be used here. In case the first element of `data` is an array, you would enter this else statement and and `auio_type` would be `None`, but it might be changed later if the second element is a Signal.
- If the operation is in the frequency domain, the array data could be complex without the need to cast all other Signals contained in `data` to `complex = True`.

The simplest solution that comes to my mind is

```
if domain == 'time' and np.asarray(d).dtype.kind == "c":
    complex = True
```

Would this work? In any case we should have test cases for that:

- time domain operation (4 tests)
  - real-valued array as first (second) element in `data`
  - complex-valued array as first (second) element in `data`
- frequency domain operation (same 4 tests as above)

pyfar/classes/audio.py
```diff
@@ -1588,14 +1602,19 @@ def _get_arithmetic_data(data, domain, cshape, matmul, audio_type):
         Signal. `np.asarray(data)` otherwise.
     """
     if isinstance(data, (Signal, TimeData, FrequencyData)):
+        data_ = data.copy()
+        # check if complex casting of any input signal is necessarry
```
f-brinkmann, line 1606:
```suggestion
        # check if complex casting of any input signal is necessa
```

pyfar/classes/audio.py
```diff
@@ -1580,6 +1592,8 @@ def _get_arithmetic_data(data, domain, cshape, matmul, audio_type):
         ``True`` if a  matrix multiplication is performed, ``False`` otherwise
     audio_type : type, None
         Type of the audio class of the operation's result.
+    complex : bool
+        Flag which indicates if the operation is performed on complex data
```
f-brinkmann, line 1596:
```suggestion
        Flag which indicates if the operation involves complex valued pyfar audio objects
```

pyfar/classes/audio.py
```diff
@@ -1588,14 +1602,19 @@ def _get_arithmetic_data(data, domain, cshape, matmul, audio_type):
         Signal. `np.asarray(data)` otherwise.
     """
     if isinstance(data, (Signal, TimeData, FrequencyData)):
+        data_ = data.copy()
+        # check if complex casting of any input signal is necessarry
+        if complex:
```
f-brinkmann, line 1607: Could this be simplified to
```suggestion
        if isinstance(data_, (Signal, TimeData)):
           data_.complex = complex
```
We could probably do this to improve readability whithout worring that the complex setter might be called even if it is already set correctly.

f-brinkmann (changes requested):
Thanks for the effort. I only found one part, that might need to be changed.

pyfar/classes/audio.py
```diff
@@ -1588,14 +1601,19 @@ def _get_arithmetic_data(data, domain, cshape, matmul, audio_type):
...
+            if isinstance(data_, (Signal, TimeData)) and not data_.complex:
+                data_.complex = True
```
mberz, line 1608: How about?
```suggestion
        if not isinstance(data_, FrequencyData):
            data_.complex = is_complex
```

mberz (changes requested):
I have a couple of comments.
I'm not so happy about overwriting the in-built `complex` dtype.
Other than that is mostly reducing unnecessary complexity/readability.

f-brinkmann (approved):

mberz (approved):
Seems like this includes a lot of code from the complex_signals branch already.
I think the comments I wanted to make are already included in the other branch.

PR #472, opened 2023-07-07
update sofar version

### LINKED ISSUE #437: read_sofa not working with Path object
## General

* pyfar version: 0.5.2
* Python version: 3.10.9

## Description

Though the docstring of [io.read_sofa](https://pyfar.readthedocs.io/en/stable/modules/pyfar.io.html#pyfar.io.read_sofa) states that the filename could be a string or a pathlib.Path, it only works with strings. The reason is that  _filename.endswith(...)_ is called in [sofa.read_sofa()](https://sofar.readthedocs.io/en/latest/sofar.functions.html#sofar.read_sofa).
This could be fixed similar as in [io.read()](https://pyfar.readthedocs.io/en/stable/modules/pyfar.io.html#pyfar.io.read) with something like `filename = pathlib.Path(filename).with_suffix('.sofa')` either in pyfar or sofar.
A workaround is to call _read_sofa_ with the Path object `filename` and `filename.as_posix()`.

## What I Did

```
import pyfar as pf
from pathlib import Path
filename = Path('file.sofa')
sofa, r_coords, s_coords = pf.io.read_sofa(filename)

AttributeError: 'WindowsPath' object has no attribute 'endswith'
```

f-brinkmann:

Closes #437 

- [x] Update sofar version in pyfar
- [x] Add test also in pyfar

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -19,5 +19,5 @@ insipid-sphinx-theme
 tox-wheel
 autodocsumm
 soundfile>=0.11.0
-sofar>=0.1.2
+sofar>=1.1.1
 nbmake>=0.7.0
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ requirements = [
     'numpy>=1.23.0',
     'scipy>=1.5.0',
     'matplotlib',
-    'sofar>=0.1.2',
+    'sofar>=1.1.1',
     'urllib3',
     'deepdiff',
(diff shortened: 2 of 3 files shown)


pingelit (approved):
Thanks for the changes. Looks good to me!

ahms5 (approved):
thanks, it's a bugfix, isn't it? according to doc Paths should be supported, but they weren't supported before.

PR #477, opened 2023-07-14
Avoid warnings in tests

f-brinkmann:

Avoid numpy warning `DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)` in pyfar testing.


pyfar/dsp/dsp.py
```diff
@@ -1285,7 +1285,7 @@ def find_impulse_response_delay(impulse_response, N=1):
             start_sample = np.nan
             warnings.warn(f"Starting sample not found for channel {ch}")
 
-        start_samples[ch] = start_sample
+        start_samples[ch] = np.squeeze(start_sample)
```
ahms5, line 1288:
```suggestion
        roots = np.squeeze(np.roots(search_region_poly))
        start_samples[ch] = start_sample
```
i would add it here, because here it starts, if I get the logic right, it the result should be squeezed, if all goes right.

ahms5 (commented):
Thanks for implementing, just one comment, otherwise looks good.

ahms5 (approved):

mberz (approved):
This qualifies as a bugfix, to be merged into main instead of develop ;)

PR #478, opened 2023-07-18
refactor find*

### LINKED ISSUE #475: Refactoring of find_nearest_* functions
## Description

Due to changes to the Coordinate class (see #369), the `find_*` functions also need changes.

Changes to functions for searching points in point clouds/ bounded domains

1. We want rename and refactor the method
- [x] `find_nearest_k`
into to return a given number of nearest points according to a specific distance measure (to be checked if that works).
Should return an index and their distance (which are matrices with dimensions (K, n_points).

2. We want to remove 
- [x] `find_slice`

3. We want a second function which allows to search for all points within a given bounded domain (probably only within a sphere of specified radius surrounding a given point for now)
That is, merging:
- [x] `find_nearest_cart`
- [x] `find_nearest_sph`

ahms5:

Closes #475

### description

- as described in #475 
- output of the index is always first dim if available k (find_nearest) or find.cshape (find_within), next is tuple for mutlidimensional self.cshape, then find.cshape (find_nearest) or numer of found points (find_within).
- please see examples in the docs
- example notebook is updated

requieres: 

- [x] #482 for show in examples (doc)
- [x] #483 for flake8


pyfar/classes/coordinates.py
```diff
@@ -1374,16 +1374,347 @@ def show(self, mask=None, **kwargs):
...
             colors[mask] = pf.plot.color('r')
             pf.plot.scatter(self, c=colors.flatten(), **kwargs)
 
+    def find_nearest(self, coords, k=1, distance_measure='euclidean'):
+        """
+        Find the k nearest coordinates points.
+
+        Parameters
+        ----------
+        coords : pf.Coordinates
```
f-brinkmann, line 1389: Should we go for a more descriptive name, e.g., 'find' or 'coordinates'?

pyfar/classes/coordinates.py
```diff
@@ -1374,16 +1374,347 @@ def show(self, mask=None, **kwargs):
...
+
+        # query nearest neighbors
+        points = points.flatten() if coords.csize == 1 else points
+
+        # nearest points
+        distance, index = kdtree.query(points, k=k)
+
+        if distance_measure == 'spherical':
+            # determine validate radius
+            radius = self.radius
```
f-brinkmann, line 1496: I think we should use `np.concatenate((self.radius, coords.radius))` for `radius`

pyfar/classes/coordinates.py
```diff
@@ -1374,16 +1374,347 @@ def show(self, mask=None, **kwargs):
...
+
+    def find_within(
+            self, coords, distance=0., distance_measure='euclidean',
+            atol=1e-15, return_sorted=False):
+        """
+        Find coordinates within certain distance to the query points.
+
+        Parameters
+        ----------
+        coords : pf.Coordinates
```
f-brinkmann, line 1545: See above for comment on parameter name

pyfar/classes/coordinates.py
```diff
@@ -1374,16 +1374,347 @@ def show(self, mask=None, **kwargs):
...
+        if k > 1:
+            distance = np.moveaxis(distance, -1, 0)
+
+        return index, distance
+
+    def find_within(
+            self, coords, distance=0., distance_measure='euclidean',
+            atol=1e-15, return_sorted=False):
+        """
+        Find coordinates within certain distance to the query points.
```
f-brinkmann, line 1541:
```suggestion
        Find coordinates within a certain distance to the query points.
```

pyfar/classes/coordinates.py
```diff
@@ -1374,16 +1374,347 @@ def show(self, mask=None, **kwargs):
...
+            Distance must be >= 0. For just exact matches use ``0``.
+            The default is ``0``.
+        distance_measure : string, optional
+            ``'euclidean'``
+                distance is determined by the euclidean distance.
+                This is default.
+            ``'spherical'``
+                distance is determined by the great-circle distance.
+            ``'angular'``
+                distance is determined by the angles in radiant.
```
f-brinkmann, line 1558: Is this also great circle distance but in radians? Should we also add this option to `find_nearest`?

pyfar/classes/coordinates.py
```diff
@@ -1374,16 +1374,347 @@ def show(self, mask=None, **kwargs):
...
+        return_sorted : bool, optional
+            Sorts returned indicies if True and does not sort them if False.
```
f-brinkmann, line 1560:
```suggestion
            Sorts returned indices if True and does not sort them if False.
```

pyfar/classes/coordinates.py
```diff
@@ -1374,16 +1374,347 @@ def show(self, mask=None, **kwargs):
...
+            If None, does not sort single point queries, but does sort
+            multi-point queries which was the behavior before this
+            option was added.
```
f-brinkmann, line 1563: Since we refactor, we could decide for a consistent default behavior to simplify things. I would opt for `sort=True`

pyfar/classes/coordinates.py
```diff
@@ -1374,16 +1374,347 @@ def show(self, mask=None, **kwargs):
...
+        index : numpy array of ints
+            The indexes of the neighbors. If ``coords.csize>1`` for each
+            dimension, if not, dimension is skipped. Then tuple for
+            the support of multidimensional self, this contains the
+            coordinates with the shape of ``coords.cshape``.
+
+        Notes
+        -----
+        This is a wrapper for ``scipy.spatial.cKDTree``. Other than previous
+        implementations, it supports self.ndim>1 as well.
```
f-brinkmann, line 1576:
```suggestion
        This is a wrapper for ``scipy.spatial.cKDTree``.
```

pyfar/classes/coordinates.py
```diff
@@ -1374,16 +1374,347 @@ def show(self, mask=None, **kwargs):
...
+        return_sorted : bool, optional
+            Sorts returned indicies if True and does not sort them if False.
+            If None, does not sort single point queries, but does sort
+            multi-point queries which was the behavior before this
+            option was added.
+
+        Returns
+        -------
+        index : numpy array of ints
+            The indexes of the neighbors. If ``coords.csize>1`` for each
```
f-brinkmann, line 1568: Can you rephrase this? I did not fully understand it.

pyfar/classes/coordinates.py
```diff
@@ -1374,16 +1374,347 @@ def show(self, mask=None, **kwargs):
...
+
+        Notes
+        -----
+        This is a wrapper for ``scipy.spatial.cKDTree``. Other than previous
+        implementations, it supports self.ndim>1 as well.
+
+        Examples
+        --------
+
+        Find frontal point from a spherical coordinate system
```
f-brinkmann, line 1581: Examples are taken from `find_nearest` and should be changed.

tests/test_deprecations.py
```diff
@@ -246,3 +246,39 @@ def test_signal_len():
         with pytest.raises(TypeError, match=re.escape("had no len()")):
             # remove Signal.__len__ from pyfar 0.8.0!
             len(pf.Signal([1, 2, 3], 44100))
+
+
+def test_deprecations_find_nearest_k():
```
f-brinkmann, line 251: Can you please add the test for checking if the attributes were deprecated everywhere. There are examples above starting with `if version.parse(pf.__version__) >= version.parse('0.8.0'):`. Otherwise we might forget to deprecate later...

f-brinkmann (changes requested):
Thanks for the effort. Looks quite good already. Mainly had comments regarding the docstring an some aspects of the output format

pyfar/classes/coordinates.py
ahms5: yes maybe, lets disucss it. This parameter is directly coming from the scipy.spatial.cKDTree

ahms5: we dicided to sinplify the docstring and remove the None.

pyfar/classes/coordinates.py
```diff
@@ -1357,8 +1357,8 @@ def show(self, mask=None, **kwargs):
         Parameters
         ----------
         mask : boolean numpy array, None, optional
-            Plot points in red if ``mask==True``. The default is ``None``,
-            which the same color for all points.
+            Mask or indexes to plot. Plot points in red if ``mask==True``.
+            The default is ``None``, which the same color for all points.
```
pingelit, line 1361:
```suggestion
            Mask or indexes to highlight. Highlight points in red if ``mask==True``.
            The default is ``None``, which plots all points in the same color.
```
Also, is indexes correct here? For me that would be an array of integers. In turn `mask==True` would not make sense. But I might be overthinking it.

pingelit (changes requested):
Overall a very nice implementation.

Just a few minor doc string changes I noticed. Regarding the `show` method, I don't know if this has to be discussed further or if any changes must be made in the method.

ahms5:
requires #482 now

pyfar/classes/coordinates.py
```diff
@@ -1380,10 +1380,309 @@ def show(self, mask=None, **kwargs):
...
+
+        Parameters
+        ----------
+        find : pf.Coordinates
+            Coordinates to which the nearest neighbors are searched.
+        k : int, optional
+            Number of points to return. k must be > 0. The default is ``1``.
+        distance_measure : string, optional
+            ``'euclidean'``
+                Distance is determined by the euclidean distance i meters
```
f-brinkmann, line 1395:
```suggestion
                Distance is determined by the euclidean distance in meters
```

pyfar/classes/coordinates.py
```diff
@@ -1380,10 +1380,309 @@ def show(self, mask=None, **kwargs):
...
+                between the points. This is default.
+            ``'spherical'``
+                Distance is determined by the great-circle distance in meters
+                between the points.
+            ``'angular'``
+                distance is determined by the angles in radiant.
```
f-brinkmann, line 1401: This I find confusing. Do I get it right that both use the great circle distance, i.e., the smallest distance between two points on a sphere but *spherical* expects it in meters and *angular* in radians? If yes, I think we should discuss better names and find a better description. If not - I'm lost.

pyfar/classes/coordinates.py
```diff
@@ -1380,10 +1380,309 @@ def show(self, mask=None, **kwargs):
...
+
+        Returns
+        -------
+        index : tuple of arrays
+            Indices of the neighbors. Arrays of shape (k, find.cshape) if k>1
+            else (find.cshape).
```
f-brinkmann, line 1407:
```suggestion
            Indices of the neighbors. Arrays of shape ``(k, find.cshape)`` if k>1
            else ``(find.cshape, )``.
```

pyfar/classes/coordinates.py
```diff
@@ -1380,10 +1380,309 @@ def show(self, mask=None, **kwargs):
...
+        distance : numpy array of floats
+            Distance between the points, after the given ``distance_measure``.
+            It's of shape (k, find.cshape).
```
f-brinkmann, line 1410: is this also squeezed to (find.cshape, ) if k=1?

pyfar/classes/coordinates.py
f-brinkmann: See above, might be good to improve naming and documentation

f-brinkmann:
Thanks, the code looks fine to me - I only marked some parts in the docstring for discussion in the next weekly.

f-brinkmann (changes requested):
.

pyfar/classes/coordinates.py
```diff
@@ -1374,16 +1375,313 @@ def show(self, mask=None, **kwargs):
...
+        --------
+
+        Find frontal point from a spherical coordinate system
+
+        .. plot::
+
+            >>> import pyfar as pf
+            >>> coords = pf.samplings.sph_lebedev(sh_order=10)
+            >>> find = pf.Coordinates(1, 0, 0)
+            >>> index, distance = coords.find_nearest(find)
```
pingelit, line 1425:
```suggestion
            >>> to_find = pf.Coordinates(1, 0, 0)
            >>> index, distance = coords.find_nearest(to_find)
```
I feel like this makes it a bit clearer. But it's not that important.

pyfar/classes/coordinates.py
```diff
@@ -1374,16 +1375,313 @@ def show(self, mask=None, **kwargs):
...
+            >>> coords.show(index)
+            >>> distance
+            >>> 0.0
```
pingelit, line 1428: Is `0.0` the output? Then the `>>>` should be removed.

pyfar/classes/coordinates.py
```diff
@@ -1374,16 +1375,313 @@ def show(self, mask=None, **kwargs):
...
+
+        Find multidimensional points in multidimensional coordinates with k=1
+
+        >>> import pyfar as pf
+        >>> import numpy as np
+        >>> coords = pf.Coordinates(np.arange(9).reshape((3, 3)), 0, 1)
+        >>> find = pf.Coordinates(
```
pingelit, line 1435: Same as above, maybe consider `to_find`.

pyfar/classes/coordinates.py
```diff
@@ -1374,16 +1375,313 @@ def show(self, mask=None, **kwargs):
...
+
+        Returns
+        -------
+        index : tuple of array
+            Indices of the containing coordinates. Arrays of shape
+            (find.cshape).
+
+        Notes
+        -----
+        This is a wrapper for ``scipy.spatial.cKDTree``. Other than previous
```
pingelit, line 1572:
```suggestion
        This is a wrapper for ``scipy.spatial.cKDTree``. Compared to previous
```
Minor rewording suggestion.

pingelit: Or did you change this with Fabians comment?

pingelit (changes requested):
Thanks. I agree with Fabians last comments. Other than that, some minor suggestions to improve the doc strings.

PR #481, opened 2023-07-28
add transpose method to _Audio

artpelling:
This PR adds a `transpose` method to all `Signals`, very similar to `np.ndarray.transpose`.

### DIFF
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -114,8 +114,9 @@ class _Audio():
         if axes is None or len(axes) == 0:
             axes = tuple(range(len(self.cshape)))[::-1]
         else:
-            tuple([a-1 if a<0 else a for a in axes])
-        np.empty(np.ones(len(self.cshape), dtype=int)).transpose(axes) # throw exception before deepcopy
+            tuple([a-1 if a < 0 else a for a in axes])
+        # throw exception before deepcopy
+        np.empty(np.ones(len(self.cshape), dtype=int)).transpose(axes)
         transposed = deepcopy(self)
         transposed._data = transposed._data.transpose(*axes, len(self.cshape))
 
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -106,10 +106,9 @@ class _Audio():
         Parameters
         ----------
         axes : `None`, iterable of ints, or n ints
-            leaving the argument empty will reverse the order of `self.caxes`.
+            `None`(default): reverses the order of `self.caxes`.
             iterable of ints: `i` in the `j`-th place of the interable means
-            that `self._data`'s `i`-th axis becomes the transposed datas `j`-th
-            axis. Defaults to `None`.
+            that the `i`-th axis becomes transposed object's `j`-th axis.
             n ints: same as 'iterable of ints'.
         """
         axes = tuple(range(len(self.cshape)))[::-1] if axes is None else tuple([a-1 if a<0 else a for a in axes])
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -112,11 +112,8 @@ class _Audio():
             axis. Defaults to `None`.
             n ints: same as 'iterable of ints'.
         """
-        axes = tuple(range(len(self.cshape)))[::-1] if len(axes) == 0 else tuple(axes)
-        assert all([isinstance(ax, int) for ax in axes]), "axes must be defined as integers"
-        assert len(axes) == len(set(axes)), "axes have to be unique"
-        assert len(axes) == max(axes) + 1 == len(self.cshape), "axes have to match caxes"
-
+        axes = tuple(range(len(self.cshape)))[::-1] if axes is None else tuple([a-1 if a<0 else a for a in axes])
+        np.empty(np.ones(len(self.cshape))).transpose(axes, -1) # throw exception before deepcopy
         transposed = deepcopy(self)
         transposed._data = transposed._data.transpose(*axes, len(self.cshape))
 
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -95,7 +95,7 @@ class _Audio():
                 newshape + (length_last_dimension, ))
         except ValueError:
             if np.prod(newshape) != np.prod(self.cshape):
-                raise ValueError((f"Can not reshape audio object of cshape "
+                raise ValueError((f"Cannot reshape audio object of cshape "
                                   f"{self.cshape} to {newshape}"))
 
         return reshaped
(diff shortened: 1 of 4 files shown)


pyfar/classes/audio.py
```diff
@@ -108,6 +108,28 @@ def reshape(self, newshape):
 
         return reshaped
 
+    def transpose(self, *axes):
```
f-brinkmann, line 111: I would suggest `axes=None` as default as in numpy

pyfar/classes/audio.py
```diff
@@ -108,6 +108,28 @@ def reshape(self, newshape):
...
+        """Return a transposed copy of the audio object.
```
f-brinkmann, line 112: Maybe even more explicit
```suggestion
        """Transpose time/frequency data and return copy of the audio object.
```

pyfar/classes/audio.py
```diff
@@ -108,6 +108,28 @@ def reshape(self, newshape):
...
+
+        Parameters
+        ----------
+        axes : empty, iterable of ints, or n ints
+            leaving the argument empty will reverse the order of `self.caxes`.
+            iterable of ints: `i` in the `j`-th place of the interable means
+            that `self._data`'s `i`-th axis becomes the transposed datas `j`-th
```
f-brinkmann, line 119: I would avoid mentioning `_data` in the docstring because we usually hide this from the user. Maybe soemthing like '... means that the `i`-ths axis of the time/frequency data becomes ...'

pyfar/classes/audio.py
```diff
@@ -108,6 +108,28 @@ def reshape(self, newshape):
...
+            axis.
+            n ints: same as 'iterable of ints'.
+        """
+        axes = tuple(range(len(self.cshape)))[::-1] if len(axes) == 0 else tuple(axes)
+        assert all([isinstance(ax, int) for ax in axes]), "axes must be defined as integers"
```
f-brinkmann, line 124: Nice, did not know the shorthand for throwing a specific error message

pyfar/classes/audio.py
```diff
@@ -108,6 +108,28 @@ def reshape(self, newshape):
...
+        assert len(axes) == len(set(axes)), "axes have to be unique"
+        assert len(axes) == max(axes) + 1 == len(self.cshape), "axes have to match caxes"
```
f-brinkmann, line 126: I think this assert will fail if specifying negative axes, e.g., `axes = (0, -1)`

f-brinkmann (changes requested):
Thanks for the PR. Only a few comments on the docstring and some gneral points to consider/discuss

- If we specify axes for audio objects we have the concept of ignoring the last axis holding the time frequency bins. This means, that axis `-1` in an audio object refers to axis `-2` of the underlying numpy array `_data`. You can account for this by `axes = [a-1 if a<0 else a for a in axes]` and add a description for this in the docstring.
- I noted that the test are failing. Might be because you branched main. Can you branch develop instead?
- Once the code is approved we should add tests to **tests/test_audio_signal.py**, **tests/test_audio_time_data.py**, and **tests/test_audio_frequency_data.py**

pyfar/classes/audio.py
artpelling: Ah thanks. I totally missed that negative axes are admissible in `np.transpose`.

artpelling:
I've implemented your suggestions. Out of lazyness, I removed all assertions and simply transpose a mock array with identical `ndim` before deep copying instead of reverse engineering all rules. This will throw numpy's exceptions. Let me know if you like it or not.

> * Once the code is approved we should add tests to **tests/test_audio_signal.py**, **tests/test_audio_time_data.py**, and **tests/test_audio_frequency_data.py**

Will do that once the code is agreed on.

pyfar/classes/audio.py
artpelling: I did that but then realized that this does not reproduce the behaviour of numpy. `np.transpose` uses `axes=None` per default, however, `np.ndarray.transpose` uses `*axes` which allows passing of tuples and multiple ints.

I have reverted these changes, left `axes` empty per default and also accept `None`.

ahms5 (commented):
Thanks for the PR, very elegant, looks fine to me. for me just tests are remaining. flake8 should be fixed with #483

artpelling:
@f-brinkmann This might be ready now.

f-brinkmann (changes requested):
Thanks, looks very good already. I only had minor comments on the docstring and testing left. Can you also add a brief test for the shorthand signal.T?

pyfar/classes/audio.py
```diff
@@ -95,11 +95,40 @@ def reshape(self, newshape):
...
+            iterable of ints: `i` in the `j`-th place of the interable means
+            that the `i`-th axis becomes transposed object's `j`-th axis.
+            n ints: same as 'iterable of ints'.
+        """
+        if hasattr(axes, '__iter__'):
+            axes = axes[0] if len(axes) == 1 else axes
+        if axes is None or len(axes) == 0:
+            axes = tuple(range(len(self.cshape)))[::-1]
+        else:
+            tuple([a-1 if a < 0 else a for a in axes])
```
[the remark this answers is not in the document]
artpelling: I think it is fine, this will lead to errors in both 
```python
np.empty(np.ones(len(self.cshape), dtype=int)).transpose(axes)
``` 
as well as
```python
transposed._data = transposed._data.transpose(*axes, len(self.cshape))
```

tests/test_audio_frequency_data.py
```diff
@@ -104,10 +104,22 @@ def test_reshape_exceptions():
         data_out = data_in.reshape([3, 2])
 
     # test assertion for wrong dimension
-    with pytest.raises(ValueError, match='Can not reshape audio object'):
+    with pytest.raises(ValueError, match='Cannot reshape audio object'):
         data_out = data_in.reshape((3, 4))
 
 
+def test_transpose():
```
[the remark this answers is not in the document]
artpelling: I feel like a lot of the tests could be reduced that way. Essentially the files `test_audio_signal.py`, `test_audio_frequency_data.py` and `test_audio_time_data.py`could be merged.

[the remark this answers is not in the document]
f-brinkmann: I never checked - you could make an issue so we don't forget to discuss it on the weekly.

f-brinkmann (approved):

f-brinkmann (dismissed):
There seems to be an error in the tests now, but otherwise approved. Thanks for the effort.

tests/test_audio_frequency_data.py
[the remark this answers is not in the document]
artpelling: Done. #516

f-brinkmann (approved):

ahms5 (approved):

PR #482, opened 2023-08-07
allow indices for Coordinates.show

ahms5:
first Coordinates.show just allow a mask, since we dont support masks for the find functions (#478) anymore we allow indexes as well.

- [x] requieres #483 for flake8


pingelit (approved):
Thanks for clarifying this.

f-brinkmann (approved):
Thanks for implementing!

PR #484, opened 2023-08-10
[fractional_time_shift] add missing arg to docstr

artpelling:
Adds a missing parameter to the docstring. My suggestion would be to also accept `'seconds'` as a unit.

### DIFF
--- a/pyfar/dsp/interpolation.py
+++ b/pyfar/dsp/interpolation.py
@@ -264,7 +264,7 @@ def fractional_time_shift(signal, shift, unit="samples", order=30,
         `Numpy broadcasting
         <https://numpy.org/doc/stable/user/basics.broadcasting.html>`_)
     unit : str, optional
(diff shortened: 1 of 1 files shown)


ahms5 (commented):
Thank you for implementing, good idea to add 'seconds'. maybe we need to also add a test for seconds.

f-brinkmann (changes requested):
For discussion: we use `'s'` for seconds in various places If we change this, I think we sould be consistent everywhere.

artpelling:
> For discussion: we use `'s'` for seconds in various places If we change this, I think we sould be consistent everywhere.

yeah I realized the same thing. I like 'seconds' as it is more clear but maybe not in this PR :)

f-brinkmann (approved):
Agreed, can you open an issue for that?

ahms5:
tests are fixed now in develop, you can merge dev into your branch

PR #485, opened 2023-08-11
draft for SamplingSphere

ahms5:
main problem is deprection of sh_order in Coordinates and refereing to spharpy.SamplingSphere  which is not ready yet is not so nice.
- propose class for SamplingSphere instead of ref to spharpy (the problem is that spharpy.samplingSphere does not inherit from pyfar.Coordinates, it has its own implementation.) after this is established we could shift pyfar.samplings and pyfar.SamplingSphere to spharpy directly
- fix deprection warnings in pf.samplings

please comment :)


f-brinkmann:
@ahms5 should we close this, since it will move to spharpy?

PR #494, opened 2023-09-07
Migrate from `flake8` to `ruff`

artpelling:
Here is a first suggestion for migration to `ruff`.
I've copied the configuration (modules, line length, ignores etc) from other projects where I use it.

Obviously, this is quite opinionated and should be discussed and tweaked so that everyone is happy.
I hope it is a good starting point, enforcing some things but not being too strict.

On top of that I've added PEP `D205`, `N8021` and `N803` to the ignores, because they were coming up a lot.
About 3000 fixable linting mistakes were autofixed by `ruff`, I have to reviewed them in detail.

- [ ] Check if flake related stuff is removed in #407


ahms5:
Hi, thanks for the implementation. Since we renamed the circle ci job into ``ruff``, you can replace the ``flake`` by ``ruff`` in the whole file in pyfar/.circleci/config.yml, then circleci should be fixed.

artpelling:
> Hi, thanks for the implementation. Since we renamed the circle ci job into `ruff`, you can replace the `flake` by `ruff` in the whole file in pyfar/.circleci/config.yml, then circleci should be fixed.

Thanks. I've fixed circleci but then had a weird error about circular imports. I've reverted the autofixes that ruff did and enabled the modules step by step. This way, the diff should be easier to understand and will hoepfully help you to decide which modules you would like to activate.

Enabling the `isort` module (automatic sorting of import statements) then leads to a circular import error. My impression is that this is not a bug in `ruff` but an underlying issue with how imports are done in `pyfar`...

ahms5:
Thank you, we just discussed it in the Meeting and we suggest that in this PR we configure ``ruff`` like ``flake8``, so that all tests pass. This would be sth like:
```
ignore = []
line-length = 79
select = ["E", "F", "W"]
```
In additional PRs, we can add more rules and restrictions, such as importing, documentation, etc.

artpelling:
> Thank you, we just discussed it in the Meeting and we suggest that in this PR we configure `ruff` like `flake8`, so that all tests pass. This would be sth like:
> 
> ```
> ignore = []
> line-length = 79
> select = ["E", "F", "W"]
> ```
> 
> In additional PRs, we can add more rules and restrictions, such as importing, documentation, etc.

Sorry it took me a while! I've just cherry picked the important commits and made a basic config that should mimic `flake8`.

.circleci/config.yml
```diff
@@ -65,8 +65,8 @@ jobs:
           # app-dir: ~/project/package-directory/  # If you're requirements.txt isn't in the root directory.
           pip-dependency-file: requirements_dev.txt
       - run:
-          name: Flake8
-          command: flake8 pyfar tests
+          name: Ruff
+          command: ruff pyfar
```
ahms5, line 69:
```suggestion
          command: ruff pyfar tests
```

ahms5 (changes requested):
Thanks for the implementation! I'm really looking forward to having ruff instead of flake8 :)
This PR also includes tox again, which we removed in #480 because nobody was using it and we had circleci to check all pyfar versions. Other than that, we need to merge the development into this brunch, which includes e.g. python 3.12. Otherwise looks good to me.

artpelling:
> Thanks for the implementation! I'm really looking forward to having ruff instead of flake8 :) This PR also includes tox again, which we removed in #480 because nobody was using it and we had circleci to check all pyfar versions. Other than that, we need to merge the development into this brunch, which includes e.g. python 3.12. Otherwise looks good to me.

Thanks. Sorry, `tox` must've slipped in again while cherry picking. I've dropped the `tox` dependency. While at it, I've also updated the `ruff` dependency to the latest version.

ahms5 (commented):
Thank you very much! Is there a specific reason why you specify the ruff version? we usually don't specify it unless a particular version doesn't work.
If we merge the dev into this, to get rid of conflix is ready to merge from my side!

artpelling:
> Thank you very much! Is there a specific reason why you specify the ruff version? we usually don't specify it unless a particular version doesn't work. If we merge the dev into this, to get rid of conflix is ready to merge from my side!

I've rebased on `develop`. I pinned the version because CI depends on it. `ruff` is quite fast moving with multiple releases every month. To ensure identical behaviour locally and on github CI it would be good to fix it. Since it does not affect the behaviour of any of the code nor any other dependencies, fixing the version here should not lead to any installation issues AFAICT.

ahms5 (approved):

f-brinkmann (dismissed):
Thanks for the effort. Fine for me
- There are no more mentions of flake in pyfar
- There are not mentions of flake in the new [all-pyfar-wide guidelines](https://github.com/pyfar/gallery/pull/51)

artpelling:
@ahms5 CircleCI is broken. I tried to fix conflicts but am not sure what is done/intended in CI. Probably just remove `test_examples` from the requirements but I don't really know what I am doing.

.circleci/config.yml
```diff
@@ -206,7 +206,8 @@ workflows:
                 - "3.9"
           requires:
             - build_and_test
-            - flake
+            - ruff
+            - test_examples
```
ahms5, line 210:
```suggestion
```
i think this should fix circle ci

mberz (commented):
Thanks for the PR. Looks good to me.

I'd squash the commit history when merging to make it more compact (compressing commits with reverted changes etc.)

mberz (approved):

PR #499, opened 2023-09-29
Fix pyfar not running in read only containers

### LINKED ISSUE #495: Cannot import pyfar when working with write protected images (Singularity)
## General

* pyfar version: 0.5.3
* Python version: 3.9
* Operating System: Ubuntu (Singularity Container)

## Description

I tried to import pyfar from a Python instance running inside a singularity container. However, since I do not have any permission to create a directory inside a write protected image, an error was raised related to the following line:

https://github.com/pyfar/pyfar/blob/a556fb93a6a3e0ea1d196726a89dad30095a422b/pyfar/signals/files.py#L27-L31
There should be some way to manually specify the file path. Alternatively, I believe it would be better to only create the directory if needed.

f-brinkmann:

Closes #495

- Only create folder, for downloading file if actually downloading a file


ahms5 (approved):

mberz:
How does creating a folder circumvent write access problems in read-only containers? If you don't have write access, how can you create a folder?

f-brinkmann:
It will allow `import pyfar` and using most of the functionality but not downloading files. Before the fix, the folder was created upon the pyfar import and made it impossible to use pyfar at all.

mberz:
Ah, so it's more of a workaround and not a fix, got it. 
I was thinking about this for other parts of the code already (the samplings for example). It may be beneficial to download these things as part of the setup when installing the package. This also circumvents the problem of not being able to use certain functionality when offline.

f-brinkmann:
> I'd at least would like to see some information of what went wrong instead of getting an arbitrary file system access error though.

Good point. @artpelling can you post the error message, that you are getting so I can specifically catch it?

adku1173:
Well, here is an example with singularity:

Build a standard singularity container with 

```sudo singularity build pyfar.sif singularity.def```

from singularity.def file with content:
```
Bootstrap: docker
From: python:3.9

%post
pip install pyfar
```

running the container with:
```
singularity exec pyfar.sif python -c "import pyfar"
```

results in:

```
  File "<string>", line 1, in <module>
  File "/usr/local/lib/python3.9/site-packages/pyfar/__init__.py", line 21, in <module>
    from . import signals
  File "/usr/local/lib/python3.9/site-packages/pyfar/signals/__init__.py", line 19, in <module>
    from . import files
  File "/usr/local/lib/python3.9/site-packages/pyfar/signals/files.py", line 30, in <module>
    os.mkdir(file_dir)
OSError: [Errno 30] Read-only file system: '/usr/local/lib/python3.9/site-packages/pyfar/signals/files'
```

pingelit (approved):

f-brinkmann:
@adku1173 you should now get a more verbose error message and importing pyfar should work. Do you have time to test if it works as intended? I think this woud require a local install from the sources by running `pip install pyfar .` in the pyfar folder

PR #500, opened 2023-09-29
Add `rad2deg` and `deg2rad`

### LINKED ISSUE #432: Handle degree values after Coordinates refactoring
## Description

The new coordinates class will always return and expect angles in radians. We decided to add functions `deg2rad` and `rad2deg` that get N-dimensional numpy arrays and has an option to convert only specific dimensions/axis.

f-brinkmann:

Closes #432

- [x] add `deg2rad` and `rad2deg`
- [x] discuss import structure

will add testing if we agree on the general behavior and functionality

### DIFF
--- a/pyfar/classes/coordinates.py
+++ b/pyfar/classes/coordinates.py
@@ -2559,39 +2559,91 @@ def cyl2cart(azimuth, height, radius):
     return x, y, z
 
 
-def rad2deg(rad, skip=2):
+def rad2deg(coordinates, domain='spherical'):
     """
-    Convert radians to degree
+    Convert coordinates in radians to degree
 
     Parameters
     ----------
-    rad : numpy array
-        N-dimensional coordinates array of shape `(..., M)`. For example four
-        coordinate points of a spherical sampling grid with the ``M=3``
-        coordinates `azimuth`, `elevation`, and `radius` would be stored in an
-        array of shape ``(4, 3)``.
-    skip : int, list, optional
-        Integer or list of integers of coordinates that are not converted. The
-        default ``2`` assumes that the radius of spherical or cylindrical
-        coordinate points should not be converted and is located at
-        ``rad[..., 2]``.
+    coordinates : array like
+        N-dimensional array of shape `(..., 3)`.
+    domain : str, optional
+        Specifies what data are contained in `coordinates`
+
+        ``'spherical'``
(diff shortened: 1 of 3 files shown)


ahms5 (commented):
isnt it more intuiotive to mark the axis which should be converted?

f-brinkmann:
> isnt it more intuiotive to mark the axis which should be converted?

It's not really an axis. For example, in a 2D array it would be the columns. But I was strugelling with a good description. Was this unclear? If we want to reverse the logic, I would suggest `convert=slice(2)` as the default. But it might be harder to understand to people who are new to Python.

f-brinkmann:
We decided to restrict the function to the pyfar case and use a parameter `domain` with possible values `'spherical'` and `'cylindircal'`

ahms5 (approved):
look fine, thanks for implementing. 
Just one small thing, instead of ``kind='deg2rad'`` we could directly write the multiplication factor, this would save the ``if kind=='bla'`` in ``_convert_angles()``
 but is more cosmetics

pyfar/classes/coordinates.py
```diff
@@ -2559,6 +2559,91 @@ def cyl2cart(azimuth, height, radius):
...
+    -------
+    coordinates : numpy array
+        The converted coordinates of the same shape as the input data.
+    """
+    return _convert_angles(coordinates, domain, 180/np.pi)
+
+
+def deg2rad(coordinates, domain='spherical'):
+    """
+    Convert coordinates in degree to radians
```
ahms5, line 2594:
```suggestion
    Convert a copy of coordinates in degree to radians
```

pyfar/classes/coordinates.py
```diff
@@ -2559,6 +2559,91 @@ def cyl2cart(azimuth, height, radius):
     return x, y, z
 
 
+def rad2deg(coordinates, domain='spherical'):
+    """
+    Convert coordinates in radians to degree
```
ahms5, line 2564:
```suggestion
    Convert a copy of coordinates in radians to degree
```

pingelit (approved):
Thanks for the implementation. I'm not quite sure if we require the `deg2rad` method. But that might just be me missing a point.

ahms5 (approved):

ahms5:
my require #507

PR #505, opened 2023-10-12
Enhance/force numeric data in audio objects

### LINKED ISSUE #431: Calculations for Signal containing NaN elements return only NaN.

f-brinkmann:

Closes #431, requires #436 (merged) and #453 (merged)

- add tests for type and values of input data in `_Audio` class
- apply tests in `__init__` and `time`/`freq` setters of all audio classes
- adapt old tests
- add new tests

I did not find a way to check for boolean data, because it is converted to float when calling np.atleast_2d(data). Any ideas?

### DIFF
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -184,6 +184,29 @@ class _Audio():
         self._assert_matching_meta_data(value)
         self._data[key] = value._data
 
+    @staticmethod
+    def _check_input_type_is_numeric(data: np.ndarray):
+        """
+        Check if input data is numeric and raise TypeError if not.
+        """
+        # check if data type is numeric
+        if data.dtype.kind not in ["u", "i", "f", "c"]:
+            raise TypeError((f"The input data is {data.dtype} must be int, "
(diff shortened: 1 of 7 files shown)


ahms5 (commented):
It is just for ``Signal``, not for ``FrequencyData`` right?

f-brinkmann:
> It is just for `Signal`, not for `FrequencyData` right?

Jein. Numeric data types are enforced for all audio objects, i.e., you can not store arrays of string in `TimeData.time` but numeric values are enforced only for `Signal`, i.e., you can store NaN's in `TimeData.time` and `FrequencyData.freq` but not in `Signal.time` or `Signal.freq`.

ahms5 (approved):

pingelit (approved):

PR #507, opened 2023-10-13
add deprecation for 'deg'

ahms5:


pyfar/classes/coordinates.py
```diff
@@ -799,6 +818,10 @@ def _set_cyl(self, azimuth, z, rho, convention='top', unit='rad'):
 
         # convert to radians
         if unit == 'deg':
+            warnings.warn((
+                "'deg' parameter will be deprecated in pyfar 0.8.0 in favor "
+                "of the 'pf.deg2rad' and 'pf.rad2deg"),
```
f-brinkmann, line 823: I would suggest this
```suggestion
                "of pyfar.deg2rad and pyfar.rad2deg"),
```

pyfar/classes/coordinates.py
```diff
@@ -887,6 +914,10 @@ def _get_cyl(self, convention='top', unit='rad'):
 
         # convert to degrees
         if unit == 'deg':
+            warnings.warn((
+                "'deg' parameter will be deprecated in pyfar 0.8.0 in favor "
+                "of the 'pf.deg2rad' and 'pf.rad2deg"),
```
f-brinkmann, line 919: see above

pyfar/classes/coordinates.py
```diff
@@ -42,7 +42,8 @@ def __init__(
         domain use :py:func:`from_spherical_colatitude`,
         :py:func:`from_spherical_elevation`, :py:func:`from_spherical_front`,
         :py:func:`from_spherical_side`, or :py:func:`from_cylindrical`
-        instead.
+        instead. For conversions from or into degree
+        use :py:func:`pf.deg2rad` and :py:func:`pf.rad2deg`.
```
f-brinkmann, line 46: Did you check how this looks in the docs? We usually use, which display only the function name:
```suggestion
        use :py:func:`~pf.deg2rad` and :py:func:`~pf.rad2deg`.
```

f-brinkmann (dismissed):
Thanks, I think this is a good idea. Only had to minor comments that I don't need to review again :). The docs don't build, but that can probably be fixed easily.

pyfar/classes/coordinates.py
ahms5: not sure, it did not change anything in the doc, so I left it. But I fixed the links in the doc

f-brinkmann (approved):

PR #510, opened 2023-10-18
Complex audio arithmetics

tluebeck:

tiny bugfix in _get_arithmetic_data
* exchange `isinstance ` with check for type. 
add tests for multiplication in frequency domain
* multiplication in time domain didn't evoke the error

### DIFF
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -1579,18 +1579,18 @@ def _get_arithmetic_data(data, domain, cshape, matmul, audio_type,
         Signal. `np.asarray(data)` otherwise.
     """
     if isinstance(data, (Signal, TimeData, FrequencyData)):
-        data_ = data.copy()
+        data = data.copy()
         # check if complex casting of any input signal is necessary
-        if not type(data_) is FrequencyData:
-            data_.complex = contains_complex
+        if not type(data) is FrequencyData:
+            data.complex = contains_complex
         # get signal in correct domain
         if domain == "time":
-            data_out = data_.time
+            data_out = data.time
         elif domain == "freq":
-            if type(data_) is Signal:
-                data_out = data_.freq_raw
+            if type(data) is Signal:
+                data_out = data.freq_raw
             else:
-                data_out = data_.freq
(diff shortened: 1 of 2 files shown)


tests/test_audio_signal_arithmetic.py
```diff
@@ -408,6 +408,27 @@ def test_complex_real_multiplication():
                         atol=1e-15)
 
 
+def test_complex_real_multiplication_freq():
+    # only test one case - everything else is tested below
+    x = Signal([1, 0, 0], 44100)
+    y = Signal([0, 1, 0], 44100, is_complex=True)
+    z = pf.multiply((x, y), 'freq')
+
+    # check result
```
f-brinkmann, line 417: Can you add `assert x.complex == False` as a check to make clear that the input data is not changes

tests/test_audio_signal_arithmetic.py
```diff
@@ -408,6 +408,27 @@ def test_complex_real_multiplication():
...
+    x.complex = True
+    ref = x.freq * y.freq
+    npt.assert_allclose(z.freq, ref, atol=1e-15)
+
+    x = Signal([1, 0, 0], 44100, is_complex=True)
+    y = Signal([0, 1, 0], 44100)
+    z = pf.multiply((x, y), 'freq')
+
+    # check result
+    y.complex = True
```
f-brinkmann, line 427: Can you add `assert y.complex == False` as a check to make clear that the input data is not changes

pyfar/classes/audio.py
```diff
@@ -1581,16 +1581,13 @@ def _get_arithmetic_data(data, domain, cshape, matmul, audio_type,
     if isinstance(data, (Signal, TimeData, FrequencyData)):
         data_ = data.copy()
```
f-brinkmann, line 1582: I think we could copy in place, i.e., `data = data.copy()` to avoid doubling the data.

f-brinkmann (changes requested):
Thanks for finding and fixing. I only had to minor suggestions

f-brinkmann (approved):

PR #513, opened 2023-10-23
add check for fft_norm in complex setter

tluebeck:

- check the fft normalization in complex setter
- throw error if complex flag is to be set and fft norm is `rms`


f-brinkmann (changes requested):
I think you also have to test in the stter for fft_norm. This should raise an error but does not

f-brinkmann (approved):

ahms5 (approved):

PR #524, opened 2023-12-12
hotfix find_within

### LINKED ISSUE #523: `find_within` bug with `distance_measure='spherical_radians'`
## General

* pyfar version: 0.6.1
* Python version: 3.9.1
* Operating System: MacOS

## Description

`find_within` bug with `distance_measure='spherical_radians'`
 
## What I Did

I want to use pf to find all point on the northern hemisphere, I used scripts below. It returns a crash.

```
find = pf.Coordinates(0, 0, 1)
spatial_mask = coords.find_within(
    find,
    distance=np.pi/2,
    distance_measure='spherical_radians',
    atol=0.1)
coords.show(spatial_mask[0])
```

I located bug here, change `==` to `in` help me solve the problem.

```
        if distance_measure == ['spherical_radians', 'spherical_meter']:
            # determine validate radius
            radius = self.radius
            delta_radius = np.max(radius) - np.min(radius)
            if delta_radius > 1e-15:
                raise ValueError(
                    "find_nearest_sph only works if all points have the same \
                    radius. Differences are larger than 1e-15")
            radius = np.max(radius)

```

furthermore, I'm interseted in the delta_radius, 1e-15 is very small so that float value is hard to satify this threshold. Could we consider lowering the threshold?

ahms5:

Closes #523

- fix bug and replace ``==`` to ``in``
- add test for spherical distance measure
- add doc and test for atol
- add new option ``radius_tol`` to change the tolerance for radius + add test

### DIFF
--- a/pyfar/classes/coordinates.py
+++ b/pyfar/classes/coordinates.py
@@ -1601,7 +1601,7 @@ class Coordinates():
             a sphere, so the radius must be constant. This parameter defines
             the maximum allowed difference within the radii. Note that too
             large tolerances can lead to wrong results. The default is 2*eps,
-            see ``atol'' for more details.
+            see ``atol`` for more details.
 
         Returns
         -------
--- a/pyfar/classes/coordinates.py
+++ b/pyfar/classes/coordinates.py
@@ -1568,7 +1568,7 @@ class Coordinates():
 
     def find_within(
             self, find, distance=0., distance_measure='euclidean',
-            atol=1e-15, radius_tol=1e-15, return_sorted=True):
+            atol=None, return_sorted=True, radius_tol=None):
         """
         Find coordinates within a certain distance to the query points.
 
@@ -1590,14 +1590,18 @@ class Coordinates():
             ``'spherical_meter'``
                 distance is determined by the great-circle distance
                 expressed in meters.
-        atol : float
-            Absolute tolerance for distance. The default is 1e-15.
-        radius_tol : float
-            Maximum allowed difference within radii for spherical distance
-            measures. The default is 1e-15.
+        atol : float, None
+            Absolute tolerance for distance. The default is 2*eps, where eps
+            is the resolution of the data type of the coordinates.
         return_sorted : bool, optional
             Sorts returned indices if True and does not sort them if False.
             The default is True.
+        radius_tol : float, None
+            For all spherical distance measures, the coordinates must be on
+            a sphere, so the radius must be constant. This parameter defines
+            the maximum allowed difference within the radii. Note that too
+            large tolerances can lead to wrong results. The default is 2*eps,
+            see ``atol'' for more details.
 
         Returns
         -------
@@ -1635,6 +1639,10 @@ class Coordinates():
         """
 
         # check the input
+        if radius_tol is None:
+            radius_tol = 2 * np.finfo(self.x.dtype).resolution
+        if atol is None:
+            atol = 2 * np.finfo(self.x.dtype).resolution
         if float(distance) < 0:
             raise ValueError("distance must be a non negative number.")
         if not isinstance(atol, float) or atol < 0:
--- a/pyfar/classes/coordinates.py
+++ b/pyfar/classes/coordinates.py
@@ -1568,7 +1568,7 @@ class Coordinates():
 
     def find_within(
             self, find, distance=0., distance_measure='euclidean',
-            atol=1e-15, return_sorted=True):
+            atol=1e-15, radius_tol=1e-15, return_sorted=True):
         """
         Find coordinates within a certain distance to the query points.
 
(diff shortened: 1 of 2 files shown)


pyfar/classes/coordinates.py
```diff
@@ -1568,7 +1568,7 @@ def find_nearest(self, find, k=1, distance_measure='euclidean'):
 
     def find_within(
             self, find, distance=0., distance_measure='euclidean',
-            atol=1e-15, return_sorted=True):
+            atol=1e-15, radius_tol=1e-15, return_sorted=True):
```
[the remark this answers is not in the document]
mberz: I'll have another look later. There's no opportunity for me to get at a computer.

mberz (changes requested):

pyfar/classes/coordinates.py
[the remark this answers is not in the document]
f-brinkmann: I would suggest to move the new parameter to the end.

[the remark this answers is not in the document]
f-brinkmann: One general question. If I remember correctly, we are using cartesian coordinates in the background to find data, even if providin spherical distance measures. Does this mean thatfor large values of `radius_tol`, we could find points that are acutally further away than specified by `distance` or not find points, that are closer then `distance``? If yes, I would suggest to mention it in the docstring.

f-brinkmann (changes requested):
Thanks for implementing. I had one question regarding the behavior of the function.

pyfar/classes/coordinates.py
```diff
@@ -1590,6 +1590,11 @@ def find_within(
             ``'spherical_meter'``
                 distance is determined by the great-circle distance
                 expressed in meters.
+        atol : float
+            Absolute tolerance for distance. The default is 1e-15.
+        radius_tol : float
+            Maximum allowed difference within radii for spherical distance
+            measures. The default is 1e-15.
```
pingelit, line 1597: From this explanation, I don't understand what happens. Maybe a general explanation regarding tolerance could help. ... Or I#m just missing something.

pingelit (changes requested):

pyfar/classes/coordinates.py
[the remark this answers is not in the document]
ahms5: I added it in the docs

pyfar/classes/coordinates.py
ahms5: fixed

pyfar/classes/coordinates.py
```diff
@@ -1590,14 +1590,18 @@ def find_within(
...
                 distance is determined by the great-circle distance
                 expressed in meters.
-        atol : float
-            Absolute tolerance for distance. The default is 1e-15.
-        radius_tol : float
-            Maximum allowed difference within radii for spherical distance
-            measures. The default is 1e-15.
+        atol : float, None
+            Absolute tolerance for distance. The default is 2*eps, where eps
+            is the resolution of the data type of the coordinates.
```
f-brinkmann, line 1595: If were not using eps, I would suggest to be a bit more clear in the docsring.
```suggestion
            Absolute tolerance for distance. The default ``None`` uses a tolerance of two times the decimal resolution, which is determined from the data type of the coordinate points using ``numpy.finfo``.
```

pyfar/classes/coordinates.py
```diff
@@ -1590,14 +1590,18 @@ def find_within(
...
         return_sorted : bool, optional
             Sorts returned indices if True and does not sort them if False.
             The default is True.
+        radius_tol : float, None
+            For all spherical distance measures, the coordinates must be on
+            a sphere, so the radius must be constant. This parameter defines
+            the maximum allowed difference within the radii. Note that too
+            large tolerances can lead to wrong results. The default is 2*eps,
```
f-brinkmann, line 1603: Maybe be more explicit and write:

Note that increasing the tolearance decreases the accuracy of the search, i.e., points that are within the search distance might not be found or points outside the search distance may be returned.

pyfar/classes/coordinates.py
```diff
@@ -1590,14 +1590,18 @@ def find_within(
...
+            see ``atol`` for more details.
```
f-brinkmann, line 1604: I would suggest to be more explicit as above and not refer to `atol`

f-brinkmann (changes requested):
I made some suggestions that might improve the docstring and learned the difference between finfo.resolution and finfo.eps - thanks

pingelit (commented):
Thanks for adapting the documentation. 

I agree with @f-brinkmann; being more specific would help even more. When it's changed I will approve.

pingelit (approved):
Thanks!

f-brinkmann (approved):

PR #532, opened 2024-01-26
Fix bug in concatenate channels

### LINKED ISSUE #531: `pyfar.utils.concatenate_channels` can change amplitude of input data and raise a value error
## General

* pyfar version: 0.6.x

## Description
The function concatenates `obj._data`. This contains the current data, which is either `obj.time` or `obj.freq_raw` and uses this to create a new object after concatening the data. If the `obj` is in the frequency domain the concatenation worked on `freq_raw` but the new object expects `freq`, i.e., the data including FFT scaling.

```python
input = pf.signals.sine(100, 100, phase=np.pi/2)
print(input.time[0, 0])
# >>> 1

input.domain = 'freq'
concat = pf.utils.concatenate_channels((input, input))
print(concat.time[0, 0])
# >>> 91.3
```
In addition, it fails when passing inputs of mixed domains, e.g., when adding to the above example

```python
input2 = pf.signals.sine(100, 100, phase=np.pi/2)
concat = pf.utils.concatenate_channels((input, input2))
# >>> ValueError...
```

f-brinkmann:

Closes #531

### DIFF
--- a/pyfar/utils.py
+++ b/pyfar/utils.py
@@ -169,14 +169,6 @@ def concatenate_channels(signals, caxis=0, broadcasting=False):
     # check matching meta data of input signals.
     [signals[0]._assert_matching_meta_data(s) for s in signals]
 
-    # force time domain for Signal objects:
-    # - shapes of data would not match in mixed domain concatenation
-    # - fft_norm may cause wrong time domain amplitudes in frequency domain
-    #   concatenation
-    if isinstance(signal, pf.Signal):
-        for signal in signals:
-            signal.domain = 'time'
-
     # broadcast signals into largest dimension and common cshapes
     if broadcasting is True:
         # broadcast signals into common cshape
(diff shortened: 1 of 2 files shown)


pingelit (approved):
Looks good to me. There might be a more elegant solution to #531 but this solve the problem from what I can tell.

sikersten (commented):
Thanks for noting the bug! What do you think of solving it explicitely by moving the if case to l. 196, e.g.:
`if isinstance(signal, pf.FrequencyData):
    data = np.concatenate([s.freq for s in signals], axis=axis)
else:
    data = np.concatenate([s.time for s in signals], axis=axis)`

f-brinkmann:
> Thanks for noting the bug! What do you think of solving it explicitely by moving the if case to l. 196, e.g.:

Did this as suggested

PR #534, opened 2024-02-01
Fix #533

### LINKED ISSUE #533: type checking in `Signal.sampling_rate`
The `sampling_rate` kwarg in the `Signal` constructor is not thoroughly typechecked or parsed which leads to errors down the line that are hard to backtrace. IMO `sampling_rate` should be cast as an integer. Of course having `sampling_rate` as a 2d-array is not a great design choice but happens sometimes, e.g. when reading data from `.mat` files

 
MWE:
``` python
import numpy as np
import pyfar as pf

n = np.random.rand(100)
fs = np.array([[100]])
sig = pf.Signal(n, sampling_rate=fs)
pf.plot.group_delay(sig) # this fails
```

### LINKED ISSUE #553: Retaining the phase for FFT normalizations `'power'` and `'psd'` is obsolete
## General

* pyfar version: 0.6.4

## Description

[pyfar.dsp.normalization](https://pyfar.readthedocs.io/en/stable/_modules/pyfar/dsp/fft.html#normalization) retaines the phase information for the normalizations `'power'` and `'psd'`, though there are regularly defined as real-valued only.
I think this is a leftover from before the `signal.freq_raw` property was introduced in 0.4.0 (PR #274), which stores the spectrum without normalization (incl. phase). We should consider removing phase information to have coherent definitions.

## What I Did

```
import pyfar as pf
import numpy as np

signal = pf.signals.noise(44100)
signal.fft_norm = 'psd'
pf.plot.freq_phase(signal)
```
image

artpelling:
Closes #533

### DIFF
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -631,7 +631,7 @@ class Signal(FrequencyData, TimeData):
         """
         # unpack array
         if hasattr(sampling_rate, '__iter__'):
-            assert len(sampling_rate) == 0
+            assert len(sampling_rate) == 1
             sampling_rate = sampling_rate[0]
 
         # initialize signal specific parameters
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -631,7 +631,9 @@ class Signal(FrequencyData, TimeData):
         """
         # unpack array
         if hasattr(sampling_rate, '__iter__'):
-            assert len(sampling_rate) == 1
+            assert len(sampling_rate) != 0
+            if len(sampling_rate) != 1:
+                raise ValueError("Multirate signals are not supported.")
             sampling_rate = sampling_rate[0]
 
         # initialize signal specific parameters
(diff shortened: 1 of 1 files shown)


mberz (changes requested):
Thanks for implementing.
I'm a bit unsure though about the implicit and arbitrary selection of a single value from an array here. I think I'd prefer if this is a bit more verbose, otherwise users might assume that multi-rate signals are supported. 
Could your maybe add a warning?
Also I was thinking if this could additionally check if the array is longer than a single entry and raise an exception in this case.

f-brinkmann:
> Could your maybe add a warning? Also I was thinking if this could additionally check if the array is longer than a single entry and raise an exception in this case.

Maybe it's sufficient to raise an exception if there are more then one entry and convert silently otherwise?

mberz:
I'd also be okay with that.

pingelit:
Thanks for the effort, I agree with the suggestion by f-brinkmann to to raise an exception if there are more than one entry.

mberz (approved):

PR #539, opened 2024-02-03
Refactor tests for coordinates to make deprecation easier

ahms5:

- restructure test, now we can just delete ``test_samplings.py`` and ``test_coordinates.py``, if we deprecate the old stuff
- deprecation warings are ignored in these files. Actaully all warnings are ignored, because PyfarDeprecationWarning is inherid from Warning, so we cannot filter the warnings.
-


sikersten (commented):
Thanks for refactoring! Most of it looks fine to me.

- I would suggest to rename the file to be deleted to something like "test_coordinates_old.py", so the new file keeps the original name "test_coordinates.py".
- I'm not sure to understand why the PyfarDeprecationWarning should be ignored. Isn't it its purpose to show deprecated code is being used?

ahms5:
I renamed it, so the diff is totally currupted. maybe we should do it at the end

PR #541, opened 2024-02-08
Update fft.py

MelsJagt:
Corrected the normalization factor for the case fft_norm=='psd'. 

The elements of the window function should be squared before taking the sum of all elements. See Equation 13 in Ahrens et al. 2020 for reference.

### DIFF
--- a/pyfar/dsp/fft.py
+++ b/pyfar/dsp/fft.py
@@ -240,7 +240,7 @@ def normalization(spec, n_samples, sampling_rate, fft_norm='none',
             norm /= (n_samples * sampling_rate)
         else:
             # Equation 13 in Ahrens et al. 2020
-            norm /= (np.sum(window)**2 * sampling_rate)
+            norm /= (np.sum(window**2) * sampling_rate)
         # the phase is kept for being able to switch between normalizations
         # altoug the power spectrum does usually not have phase information,
         # i.e., spec = np.abs(spec)**2
--- a/pyfar/dsp/fft.py
+++ b/pyfar/dsp/fft.py
@@ -240,7 +240,7 @@ def normalization(spec, n_samples, sampling_rate, fft_norm='none',
             norm /= (n_samples * sampling_rate)
         else:
             # Equation 13 in Ahrens et al. 2020
-            norm /= (np.sum(window**2) * sampling_rate)
+            norm /= (np.sum(np.asarray(window)**2) * sampling_rate)
         # the phase is kept for being able to switch between normalizations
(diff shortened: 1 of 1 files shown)


pyfar/dsp/fft.py
```diff
@@ -240,7 +240,7 @@ def normalization(spec, n_samples, sampling_rate, fft_norm='none',
             norm /= (n_samples * sampling_rate)
         else:
             # Equation 13 in Ahrens et al. 2020
-            norm /= (np.sum(window)**2 * sampling_rate)
+            norm /= (np.sum(window**2) * sampling_rate)
```
f-brinkmann, line 243: I think the tests are currently failing because we are passing a window as a list. Changing to 
```suggestion
            norm /= (np.sum(np.asarray(window)**2) * sampling_rate)
```
would be more robust. After that we can see if the tests need to be adapted. Did you check if the inverse normalization works as expected?

f-brinkmann (changes requested):
Thanks for noting and starting the PR. I saw the tests are failing and made a suggestion to fix it.

pyfar/dsp/fft.py
MelsJagt: Indeed, it think the issue relates to the fact that `window` should be of type `np.ndarray`  (I forgot to check this).

Also I did not properly check the inverse normaliation yet.

f-brinkmann:
Thanks for the update. I'll check the tests - I think the inverse normalization should be tested. I'll let you know the next days. Apart from that everything looks fine to me

f-brinkmann:
There was a test, but it looked a bit fishy to me, because the window we used for testing contained only ones. Can you exchange the following test in tests/test_fft.py and ad an `import pytest` at the top or should we take it from here?

```python
@pytest.mark.parametrize('window', ([.5, 1, 1, .5], np.array([.5, 1, 1, .5])))
@pytest.mark.parametrize(
    'fft_norm', ('unitary', 'amplitude', 'rms', 'power', 'psd'))
def test_normalization_with_window(window, fft_norm):
    """
    Test if the window cancels out if applying the normalization and
    inverse normalization and if it works with lists and numpy arrays.
    """

    spec = fft.normalization(
        np.array([.5, 1, .5]), 4, 44100, fft_norm, window=window)
    spec = fft.normalization(
        spec, 4, 44100, fft_norm, inverse=True, window=window)
    npt.assert_allclose(spec, np.array([.5, 1, .5]), atol=1e-15)
```

f-brinkmann (approved):

ahms5 (approved):

PR #543, opened 2024-02-12
concatenate bins

### LINKED ISSUE #469: merge_data
add `merge_data` for FrequencyData and TimeData. concatenate and sort in freq or time domain.

ahms5:

Closes #469

-
-
-

### DIFF
--- a/pyfar/utils.py
+++ b/pyfar/utils.py
@@ -212,3 +212,47 @@ def concatenate_channels(signals, caxis=0, broadcasting=False):
         return pf.TimeData(data, signals[0].times)
     else:
         return pf.FrequencyData(data, signals[0].frequencies)
+
+
+def concatenate_bins(signals):
+    """
+    Merge multiple Signal, Timedata or Frequencydata objects along the time or
+    frequency axe.
+
+    Parameters
+    ----------
+    signals : tuple of Signal, TimeData or FrequencyData
+        The signals to concatenate. All signals must be of the same object type
+        and either have the same cshape or be broadcastable to the same cshape,
+        except in the dimension corresponding to caxis (the first, by default).
+        If this is the case, set ``broadcasting=True``.
+
+    Returns
+    -------
+    merged : Signal, TimeData, FrequencyData
+        The merged signal object.
+    """
+    # check input
+    for signal in signals:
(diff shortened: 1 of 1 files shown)


mberz (changes requested):
Hey, thanks for implementing this. 
I have a few conceptual comments/questions. 
1. Would it make sense to combine this with a function which also supports merging TimeData?
2. Since the function does sort the frequency data, it's more of a merge, rather than concatenation of the data. I'd rename to sth. like `merge_data` or similar.

ahms5:
Good question, but then sorting the bins does not make much scence, doesn't it? or we also add concatenate, which just concatenate the data, and merge does the sorting... an other option would be to add the sorting as an option... not sure what is the best way to go for.

sure we can rename it. lets finalize your first question first.

pyfar/utils.py
```diff
@@ -212,3 +212,45 @@ def concatenate_channels(signals, caxis=0, broadcasting=False):
...
+    merged : FrequencyData
+        The merged signal object.
+    """
+    # check input
+    if not isinstance(signals, (tuple, list)):
+        raise TypeError(
+            "Input must be a tuple or list of pf.FrequencyData objects.")
+
+    for signal in signals:
+        if not isinstance(signal, (pf.FrequencyData)):
```
f-brinkmann, line 239: This is `True` for `pyfar.Signal` due to the inheritance. I think `type(signal) != pf.FrequencyData` is what you want.

pyfar/utils.py
```diff
@@ -212,3 +212,45 @@ def concatenate_channels(signals, caxis=0, broadcasting=False):
         return pf.TimeData(data, signals[0].times)
     else:
         return pf.FrequencyData(data, signals[0].frequencies)
+
+
+def concatenate_bins(signals):
```
f-brinkmann, line 217: Since we now have functions to broadcast cshapes of audio objects it would be nice to include this by default to make this more flexible.

f-brinkmann (changes requested):
Also in favor for discussing suggestions from @mberz and added one more idea to improve flexibility.

pyfar/utils.py
pingelit: This should also be tested then.

mberz:
I've linked #469 to this PR, @ahms5 please check if this is correct.

PR #546, opened 2024-02-25
Complex group delay

### LINKED ISSUE #384: Roadmap to complex signals
Problem: We want to allow signals with complex time data 

Required changes in audio classes:
- [x] TimeData: add `complex` argument in init and make it a class property
- [x] Signal.domain: check `complex` and chose FFT accordingly                           
- [x] Signal.freq: Call normalization according to `complex` property
- [x] FFT module: wrap fft and ifft and fft_frequencies in analogy to the real parts
- [x] Frequencies in complex signals will go from -fs/2 to fs/2
- [x] check arithmetic operations on audio class, with one complex and one real-valued signal.

Changes in plot module (good first issues)
- [x] Plot functions time domain: add parameter `???` to plot `'real'`, `'imag'`, or `'abs'` values
- [x] Plot functions frequency domain: add parameter `side` to plot `'left'` or `'right'` sided spectrum
- [x] Add plot interaction for both parameters

Add tests in dsp module (good first issues)
- [x] average
- [x] convolve
- [x] decibel
- [x] deconvolve
- [x] fractional_time_shift
- [x] phase
- [x] pad_zeros
- [x] regularized_spectrum_inversion
- [x] resample (depends on scipy implementation)
- [x] spectrogram
- [x] time_shift
- [x] time_window
- [x] filter classes

Throw errors in dsp module - we decided to implement energy, power, and rms for complex signals
- [x] energy
- [x] power
- [x] rms
 
Possibily change functionality
- [x] find_impulse_response_delay:
- [x] find_impulse_response_start: detect separately for real and imag and return earliest or use absolute value (test with gammatone filters and decide later, probably absolute value because this is the envelope)
- [x] group_delay (check because the estimation is FFT based)
- [x] normalize (some modes like rms not supported)
- [x] smooth_fractional_octaves
- [x] update concatenate function (see comment in #452)

Open question (throw errors for now)
- [x] linear_phase
- [x] minimum_phase
- [x] zero_phase

To discuss:
- arithmetic operation that casts a real to complex valued signal, with invalid ftt_norm. Cast to none or error?

To do:
- check docstrings for all above methods

tluebeck:

Closes #384

- edit group delay to work with complex signals

### DIFF
--- a/tests/test_dsp.py
+++ b/tests/test_dsp.py
@@ -93,6 +93,34 @@ def test_group_delay_single_channel(impulse_group_delay):
         grp, impulse_group_delay[1].flatten(), rtol=1e-10, atol=1e-10)
 
 
+def test_complex_group_delay_single_channel(impulse_complex_group_delay):
+    """Test the function returning the group delay of a signal,
+    single channel."""
+    signal = impulse_complex_group_delay[0]
+
+    with pytest.raises(ValueError, match="Invalid method"):
+        dsp.group_delay(signal, method='invalid')
+
+    with pytest.raises(ValueError, match="not supported"):
+        dsp.group_delay(signal, method='fft', frequencies=[1, 2, 3])
+
+    grp = dsp.group_delay(signal, method='scipy')
+    assert grp.shape == (signal.n_bins, )
+    npt.assert_allclose(
+        grp, impulse_complex_group_delay[1].flatten(), rtol=1e-10, atol=1e-10)
+
+    grp = dsp.group_delay(signal, method='fft')
+    assert grp.shape == (signal.n_bins, )
+    npt.assert_allclose(
+        grp, impulse_complex_group_delay[1].flatten(), rtol=1e-10, atol=1e-10)
+
+    grp = dsp.group_delay(
+        signal, method='fft')
+    assert grp.shape == (signal.n_bins, )
+    npt.assert_allclose(
+        grp, impulse_complex_group_delay[1].flatten(), rtol=1e-10, atol=1e-10)
+
+
 def test_group_delay_two_channel(impulse_group_delay_two_channel):
     """Test the function returning the group delay of a signal,
     two channels."""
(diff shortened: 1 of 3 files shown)


tests/test_dsp.py
```diff
@@ -93,6 +93,34 @@ def test_group_delay_single_channel(impulse_group_delay):
         grp, impulse_group_delay[1].flatten(), rtol=1e-10, atol=1e-10)
 
 
+def test_complex_group_delay_single_channel(impulse_complex_group_delay):
+    """Test the function returning the group delay of a signal,
+    single channel."""
+    signal = impulse_complex_group_delay[0]
+
+    with pytest.raises(ValueError, match="Invalid method"):
```
f-brinkmann, line 101: I guess the ValueErrors are already tested elsewhere and could be removed here

tests/test_dsp.py
```diff
@@ -93,6 +93,34 @@ def test_group_delay_single_channel(impulse_group_delay):
...
+    assert grp.shape == (signal.n_bins, )
+    npt.assert_allclose(
+        grp, impulse_complex_group_delay[1].flatten(), rtol=1e-10, atol=1e-10)
+
+    grp = dsp.group_delay(signal, method='fft')
+    assert grp.shape == (signal.n_bins, )
+    npt.assert_allclose(
+        grp, impulse_complex_group_delay[1].flatten(), rtol=1e-10, atol=1e-10)
+
+    grp = dsp.group_delay(
```
f-brinkmann, line 117: This looks like the same test case as above. If true, it could be deleted

f-brinkmann (changes requested):
Thanks for implementing. Just had two minor comments

tests/test_dsp.py
tluebeck: Probably you're right. I just tested the same as for non complex valued signals to be sure we have the same behavior. But I kill these lines...

tests/test_dsp.py
tluebeck: Yes, but here it's tested for the 'fft' method. The test above is for 'scipy' method. I think its good to test, that both methods do the same for complex signals.

f-brinkmann: The comment might have moved, I was referring to the code below, where there are two test for `grp = dsp.group_delay(signal, method='fft')`

ahms5 (approved):

f-brinkmann (approved):

PR #547, opened 2024-02-26
Test complex concatenate

tluebeck:

Closes #

- test concatenation of complex signals


pyfar/utils.py
```diff
@@ -207,8 +212,10 @@ def concatenate_channels(signals, caxis=0, broadcasting=False):
         return pf.Signal(data, signals[0].sampling_rate,
                          n_samples=signals[0].n_samples,
                          domain=signals[0].domain,
```
f-brinkmann, line 214: This is correct but maybe a little implicit: The domain will always be `time` because we querried s.time in line 209. How about being more explicit

```suggestion
                         domain='time',
```

f-brinkmann (approved):
Thanks for implementing - looks fine to me. Just had one optional comment

ahms5 (approved):

PR #549, opened 2024-02-27
remove obsolete information from print Coordinates

ahms5:

change output from:
```
1D Coordinates object with 40 points of cshape (40,)
domain: cart, convention: right, unit: met
coordinates: x in meters, y in meters, z in meters
Does not contain sampling weights
```
to:
```
1D Coordinates object with 40 points of cshape (40,)
Does not contain sampling weights
```

### DIFF
--- a/pyfar/classes/coordinates.py
+++ b/pyfar/classes/coordinates.py
@@ -2651,17 +2651,8 @@ class Coordinates():
         else:
             obj = "Empty Coordinates object"
 
-        # coordinate convention
-        conv = "domain: {}, convention: {}, unit: {}".format(
-            self._system['domain'], self._system['convention'],
-            self._system['unit'])
-
-        # coordinates and units
-        coords = ["{} in {}".format(c, u) for c, u in
-                  zip(self._system['coordinates'], self._system['units'])]
-
         # join information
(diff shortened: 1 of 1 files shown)


f-brinkmann (approved):
Thanks, this looks clearner now. But I'm not sure if we consider it a bug and merge into main or merge into dev for later release.

sikersten (approved):

sikersten:
> Thanks, this looks clearner now. But I'm not sure if we consider it a bug and merge into main or merge into dev for later release.

+1 for bugfix in develop instead of hotfix in main

pingelit (approved):

pingelit:
Yes, I also think this is a bugfix.

ahms5:
a bugfix is a bug on ``development`` and will be fixed in a patch release, a hotfix is a bug on ``main`` and should be fixes by a patch release. (even it's not that critical). 
Every version, in this case ``0.6.x``, in itself should be as free from bugs as possible, so I would like to handle it as a hotfix.

mberz:
Thanks for implementing, I'm all for more compact reprs
Since everyone accepted, I'll merge into main.

PR #555, opened 2024-03-05
Return the axes from the `Coordinates.show`

pingelit:

`Coordinates.show` does not actually return the axes object.

- actually return the axes object from `pyfar.scatter`


pingelit:
I misread #554, closing.

PR #562, opened 2024-03-15
Replace latest with stable when linking to the docs

mberz:
We're often linking to the latest version of the docs, which points to the latest commit on the main branch.
This is replaced with the stable version, which is the latest tagged commit.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -187,7 +187,7 @@ Mocks are similar to stubs but used for **behavioral verification**. For example
 Writing the Documentation
 -------------------------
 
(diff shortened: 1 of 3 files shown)


pingelit (approved):

f-brinkmann (approved):

sikersten:
Most of the changes might get obsolete, when we agree on the changes in #563.
Is there a reason to link to ..io/.../stable, in case where the webpage can just be accessed with ....io?

ahms5:
edit: the changes @sikersten mentioned are in https://github.com/pyfar/pyfar/pull/569

PR #564, opened 2024-03-23
Remove restriction for matplotlib version

### LINKED ISSUE #501: Account for deprecation of `matplotlib.tight_layout` module
## General

* pyfar version: 0.5.4
* Matplotlib version: >=3.8

## Description

Matplotlib is deprecating the `matplotlib.tight_layout`. It contains the function `get_subplotspec_list` that we use to check if `tight_layout` should be applied to a figure. Without this check, using `tight_layout` may result in figures with overlapping elements. This happens for example for a 2,1 subplot layout with a shared colorbar on the side (see example below).

## Possible solution

We could use [constrained](https://matplotlib.org/stable/users/explain/axes/constrainedlayout_guide.html) layout as the new default (see example below). It used to be experimental but seems to be stable now. It works in most cases with few exceptions only, and it seems to be further improved.

## Implementation

Constrained layout can be enabled via the plotstyle by setting `figure.constrained_layout.use : True`. I would suggest to add two new plotstyles *light_constrained.mplstyle* and *dark_constrained.mplsytle*. The former would be the new default in favor of *light.mplstyle*. This would save a bunch of code in pyfar and maintain full flexibility. If using *light.mplstyle* manual layout or `tight_layout` could be applied to a figure before or after plotting with pyfar. I tested it offline and it seems to work.

```python
import numpy as np

for layout in ['tight', 'constrained']:
    np.random.seed(19680801)
    fig, axs = plt.subplots(2, 1, figsize=(2, 2), layout=layout)
    img = axs[0].pcolormesh(np.random.rand(10, 20))
    axs[1].pcolormesh(np.random.rand(10, 20))
    fig.colorbar(img, ax=axs, orientation='vertical', fraction=.1)
    plt.show()

    fig.savefig(f'test_{layout}.png')
    plt.clf()
```

Tight Layout
test_tight

Constrained Layout
test_constrained

f-brinkmann:

Closes #501 

- Remove version restriction for Matplotlib

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -11,7 +11,7 @@ pytest-runner
 pytest-cov
 numpy>=1.23.0
 scipy>=1.5.0
-matplotlib<=3.7
+matplotlib
 urllib3
 deepdiff
 autodocsumm
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ with open('HISTORY.rst') as history_file:
 requirements = [
     'numpy>=1.23.0',
     'scipy>=1.5.0',
-    'matplotlib<=3.7',
+    'matplotlib',
     'sofar>=0.1.2',
     'urllib3',
     'deepdiff',


mberz:
The code changes making the constraint obsolete were merged in PR https://github.com/pyfar/pyfar/pull/406

sikersten (approved):

ahms5 (approved):

PR #570, opened 2024-04-05
remove unused github workflows

ahms5:
we are not using the workflow, so I think we can remove it.


f-brinkmann (approved):

f-brinkmann:
We could delete the projects completely after merging this.

sikersten (approved):

PR #591, opened 2024-04-20
Documentation/remove concepts

f-brinkmann:
### Which issue(s) are closed by this pull request?

closes #551, part of https://github.com/pyfar/gallery/issues/20

### Changes proposed in this pull request:

- [x] Move information from Concepts>Coordinates to Classes>Coordinates and update references
- [x] Remove all other concepts pages and link to pyfar gallery instead
- [x] replace links to gallery with intersphinx links
- [x] add introductory section to pyfar.utils linking to cshape, caxis, and cdim

PR #597, opened 2024-04-26
Fix math and description of coordinates

f-brinkmann:

Docstring for coordinates properties are now identical with the table introduced in #591. Note that this includes a couple of small fixes where coordinate names were confused after copy/past mistakes and correcting the coordinate ranges from < to \leq if required.

### DIFF
--- a/pyfar/classes/coordinates.py
+++ b/pyfar/classes/coordinates.py
@@ -1155,7 +1155,7 @@ class Coordinates():
     @property
     def rho(self):
         r"""
-        Distance perpendicular to the the z-axis of the right handed Cartesian
+        Radial distance to the the z-axis of the right handed Cartesian
         coordinate system (:math:`0` < rho < :math:`\infty`)."""
         return self.cylindrical[..., 2]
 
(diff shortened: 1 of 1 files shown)


pyfar/classes/coordinates.py
```diff
@@ -1155,8 +1155,8 @@ def z(self, value):
     @property
     def rho(self):
         r"""
-        Radial distance to the the z-axis of the right handed
-        Cartesian coordinate system in meters (0 \leq radius < \infty)."""
+        Distance perpendicular to the the z-axis of the right handed Cartesian
```
mberz, line 1158:
```suggestion
        Radial distance to the the z-axis of the right handed Cartesian
```
This was correct before. A distance is not perpendicular, only the vector (with length corresponding to the distance) is perpendicular. ;) Please change back.

mberz (changes requested):

ahms5 (changes requested):
thank you for taking care, we might need to add Marcos Comments also in the table in the other pr. and for some reason its not renderd correctly: https://pyfar--597.org.readthedocs.build/en/597/classes/pyfar.coordinates.html#pyfar.classes.coordinates.Coordinates.rho

f-brinkmann:
@ahms5 I corrected it in the other pull as well. But the link you gave above seems to render correct for me...

PR #598, opened 2024-04-29
Test find impulse response start

tluebeck:

Closes #

-
-
-


ahms5 (changes requested):
Thanky you, looks fine, but maybe we should add a hint in the function doc what will happen if the signal is complex.

PR #601, opened 2024-05-02
Test complex frac time shift

tluebeck:

Closes #

-
-
-


f-brinkmann (changes requested):
You are using an integer dealy for testing in which case only `signal = pf.dsp.time_shift(signal, delay_int, mode)` is used. You could use a fractional dealy and test if the group_delay behaves as expected within a heuristically defined frequency range outside of which the delay is non-ideal.

f-brinkmann (approved):
Thanks - looks good to me now

ahms5 (approved):
Looks fine, but I'm also not sure about the context

PR #612, opened 2024-05-22
ruff: manual docstring rules

ahms5:
- add the checks for the docsstring, see ``pyproject.toml``
- full list can be found here: https://docs.astral.sh/ruff/rules/#pydocstyle-d

I have added this rules because accoring to my understanding they should be included anyway. 
We could also think about adding 
D100 - D107 to make sure that every public class, function, module, etc has a documentation. for now we dont have it, we could also exclude the tests in this case
- [x] add D415, D417

update since we follow numpy docstyle we can choose the numpy setting in ruff see pr #685 :
https://docs.astral.sh/ruff/faq/#does-ruff-support-numpy-or-google-style-docstrings


f-brinkmann (dismissed):
I'm fine with the changes caused by this and I would also support adding D100 - D106. In our case D107 does not seem so useful because we usually document the __init__ as part of the class Docstring.

ahms5:
Thanks for your feedback. I've created a new pr for D100-D107 in #614

pyproject.toml
```diff
@@ -5,10 +5,24 @@ exclude = [
     "setup.py",
 ]
 line-length = 79
-lint.ignore = []
+lint.ignore = [
```
pingelit, line 8: Maybe remove this change?

pingelit (approved):
Thanks, this look very nice. Just one very minor comment.

pyproject.toml
```diff
@@ -5,10 +5,24 @@ exclude = [
...
+]
```
ahms5, line 9:
```suggestion
lint.ignore = []
```

pyproject.toml
ahms5: done

pingelit (approved):

f-brinkmann (dismissed):
There are 3 ruff errors left, that seem to be easy to fix. Approved otherwise - thanks for the effort

ahms5:
merging dev into this pr required more fixes, please review again @f-brinkmann and others.

pyfar/classes/audio.py
```diff
@@ -373,6 +373,7 @@ def time(self, value):
 
     @property
     def complex(self):
+        """Return or set the flag indicating if the data is complex."""
```
f-brinkmann, line 376: Maybe more explicit: `"""Return or set the flag indicating if the time data is complex."""`. The property does not exist for Frequency data.

pyfar/classes/audio.py
```diff
@@ -965,6 +966,7 @@ def sampling_rate(self, value):
...
```
f-brinkmann, line 969: See comment above.

f-brinkmann (approved):
Thanks a lot - this looks like it was quite a bunch of work. I only found one added docstring that might be slightly improved.

f-brinkmann (approved):

pyfar/io/_codec.py
```diff
@@ -340,11 +340,12 @@ def _encode_ndarray(ndarray):
...
     ----------
-    ndarray: numpy.array.
+    ndarray: numpy.array
+        The numpy array that should be encoded.
 
     Returns
     -------
-    bytes.
+    bytes : bytes
         They bytes that where written by `numpy.save` into a memfile.
```
hoyer-a, line 349:
```suggestion
        The bytes that were written by `numpy.save` into a memfile.
```

pyfar/plot/_interaction.py
```diff
@@ -570,7 +570,7 @@ def toggle_side(self):
 
     @property
     def plot_type(self):
-        """Return current the plot type"""
+        """Return current the plot type."""
```
hoyer-a, line 573:
```suggestion
        """Return the current plot type."""
```

hoyer-a (commented):
Just two small suggestions. Thanks for the effort!

hoyer-a (approved):

f-brinkmann:
Yes, I think so

PR #613, opened 2024-05-22
add pytest preps via ruff

ahms5:
- add support of pytest preps
- list can be find here https://docs.astral.sh/ruff/rules/#flake8-pytest-style-pt
- i ignored some, please see lint.ignore


f-brinkmann (approved):
The only thing you actively changes is *pyproject.toml*, right? I would be fine with enforcing the changes caused by that. It looks a little cleaner to me in most cases.

ahms5:
basically correct, but I had to fix some ruff warnings by hand.

pyproject.toml
```diff
@@ -5,10 +5,15 @@ exclude = [
     "setup.py",
 ]
 line-length = 79
-lint.ignore = []
+lint.ignore = [
+    "PT011", # use match in pytest.raises
```
pingelit, line 9: Why not include this?

pingelit (dismissed):
Thanks for adding this, I would also add `PT011` but this is probably another big chink of work.

pyproject.toml
ahms5: Surely it would make sense. I think I was just too lazy to add it manually at any time.

ahms5:
let's wait for the discussion in https://github.com/pyfar/pyfar/issues/647

f-brinkmann (dismissed):

ahms5:
closes in favor of #692

PR #614, opened 2024-05-27
ruff: check docstring format

ahms5:
- add D100 - D107 to ruff and add remaining docs
- here are the things we need to fix https://circleci.com/gh/pyfar/pyfar/8710
- ChatGPT and I have added documentation where its required. Please also crosscheck the content.


f-brinkmann (changes requested):
Thanks for the effort. I generarly like having this. In some cases you added documentation that shows up in the build docs:

- classes/orientations.py
- dsp/dsp.py
- utils.py

In those cases we should put in a little more effort. But maybe wait until we have the opinon from others. In case of utils.py a proper docstring is introduced in #591

pyfar/classes/orientations.py
```diff
@@ -1,3 +1,4 @@
+"""This module contains the Orientations class. """
```
ahms5, line 1: shows up in doc

pyfar/dsp/__init__.py
```diff
@@ -1,3 +1,5 @@
+"""Digital signal processing functions."""
```
ahms5, line 1: shows up in doc

pyfar/io/__init__.py
```diff
@@ -1,3 +1,7 @@
+"""
```
ahms5, line 1: shows up in doc

ahms5 (commented):
marked all parts of the document that are part of the document.

pyfar/classes/orientations.py
ahms5: maybe we should shift the doc from the class to the toplevel

f-brinkmann (approved):

PR #615, opened 2024-05-27
Plot complex signals

tluebeck:

Closes #

-
-
-


pyfar/plot/_line.py
```diff
@@ -77,6 +90,21 @@ def _freq(signal, dB=True, log_prefix=None, log_reference=1, freq_scale='log',
...
+        elif side == 'left':
+            dc_idx = data.shape[-1] // 2
+            data = data[..., dc_idx-1:]
+            frequencies = np.atleast_1d(
+                dsp.fft.rfftfreq(signal.n_samples, signal.sampling_rate))
+        else:
+            ValueError('Invalid side parameter, pass left or right ...')
+    else:
+        frequencies = signal.frequencies
+
```
f-brinkmann, line 107: We could discuss to make this more explicit

- use 'auto' as the default value for `side`
- mention in the docs, that side can be 'left' or 'right' in case of complex signals (default 'right')
- throw an error if side is not 'auto' in case of `FrequencyData` and real valued `Signal` objects.

Probably, we could have a private function for this to avoid rewriting code. It could get the data (magnitude, phase, group delay) and type(signal).

pyfar/plot/_line.py
```diff
@@ -28,6 +28,19 @@ def _time(signal, dB=False, log_prefix=20, log_reference=1, unit="s",
...
+    elif show_real_imag_abs == 'imag':
+        data = np.imag(data)
+        y_label = 'Amplitude (imaginary)'
+    elif show_real_imag_abs == 'abs':
+        data = np.abs(data)
+        y_label = 'Amplitude (absolute)'
+    else:
+        raise ValueError('`show_real_imag_abs` has to be `real`, `imag`, or '
+                         f'`abs`, but is {show_real_imag_abs}.')
     # auto detect the time unit
```
f-brinkmann, line 44: - Could probably go into a private function to be reused.
- And you made it very clear, that we need a better name for the parameter :)

f-brinkmann (commented):
Thanks for the draft. I left two comments for discussion

pyfar/plot/_line.py
tluebeck: I shifted everything into a private method in _utils. Do you think we really need the mask as discussed in today's meeting. I think everything is implicitly covered by the is_complex check, isn't it?

pyfar/plot/line.py
```diff
@@ -289,7 +289,8 @@ def group_delay(signal, unit="s", freq_scale='log', ax=None, style='light',
 
 def time_freq(signal, dB_time=False, dB_freq=True, log_prefix_time=20,
               log_prefix_freq=None, log_reference=1, freq_scale='log',
-              unit="s", ax=None, style='light', **kwargs):
+              unit="s", ax=None, show_real_imag_abs='real',
+              side='right', style='light', **kwargs):
```
ahms5, line 293:
```suggestion
              unit="s", ax=None, style='light',
              show_real_imag_abs='real', side='right', **kwargs):
```
add it at the end, due to compability issues

tests/test_plot.py
```diff
@@ -74,6 +74,56 @@ def test_line_plots(function, handsome_signal, handsome_signal_v2):
...
+    'real', 'imag', 'abs'])
+def test_complex_time_plots(show_real_imag_abs_flag,
+                            handsome_complex_signal,
+                            handsome_complex_signal_v2):
+    """Test all line plots with default arguments and hold functionality."""
+    print(f"Testing: {plot.time.__name__}, show_real_imag_abs "
+          f"= {show_real_imag_abs_flag}")
+
+    # initial plot
+    filename = plot.time.__name__ + '_default'
```
ahms5, line 87:
```suggestion
    filename = f'{plot.time.__name__}_{show_real_imag_abs_flag}_default'
```
otherwise the baseline will be overwirtten anytime

tests/test_plot.py
```diff
@@ -74,6 +74,56 @@ def test_line_plots(function, handsome_signal, handsome_signal_v2):
...
+@pytest.mark.parametrize('side_flag', [
+                         'left', 'right'])
+def test_complex_freq_plots(function, side_flag,
+                            handsome_complex_signal,
+                            handsome_complex_signal_v2):
+    """Test all line plots with default arguments and hold functionality."""
+    print(f"Testing: {function.__name__}")
+
+    # initial plot
+    filename = function.__name__ + '_default'
```
ahms5, line 114:
```suggestion
    filename = f'{function.__name__}_{side_flag}_default'
```

tests/test_plot.py
```diff
@@ -74,6 +74,56 @@ def test_line_plots(function, handsome_signal, handsome_signal_v2):
...
+    create_figure()
+    function(handsome_complex_signal, side=side_flag)
+    save_and_compare(create_baseline, baseline_path, output_path, filename,
+                     file_type, compare_output)
+
+    # test hold functionality
+    filename = function.__name__ + '_hold'
```
ahms5, line 121:
```suggestion
    filename = f'{function.__name__}_{side_flag}_hold'
```

tests/test_plot.py
```diff
@@ -74,6 +74,56 @@ def test_line_plots(function, handsome_signal, handsome_signal_v2):
...
+    # initial plot
+    filename = plot.time.__name__ + '_default'
+    create_figure()
+    plot.time(handsome_complex_signal,
+              show_real_imag_abs=show_real_imag_abs_flag)
+    save_and_compare(create_baseline, baseline_path, output_path, filename,
+                     file_type, compare_output)
+
+    # test hold functionality
+    filename = plot.time.__name__ + '_hold'
```
ahms5, line 95:
```suggestion
    filename = f'{plot.time.__name__}_{show_real_imag_abs_flag}_hold'
```

ahms5 (changes requested):
Thank you! just small comments and we also need to add the new baseline plots for your functions to this pr.

f-brinkmann:
I agree with ahms points and would suggest to surrender the search of a good parameter name and simply use `mode` for denoting if the real, abs, or imag should be showin in time plots - maybe we agree on this (or something else) later today in the weekly.

PR #621, opened 2024-06-07
Introduce radius_tol to Coordiantes.find_nearest()

### LINKED ISSUE #616: Add radius tolerance to `Coordinates.find_nearest`
## General

* pyfar version: 0.6.5

## Description

`Coordinates.find_nearest` throws an error if using spherical distance measures in case the difference between radii is minimally larger than zero. We should introduce an `radius_tol` parameter as in `Coodinates.find_within`.

hoyer-a:
Closes #616 

Added a radius tolerance ``radius_tol`` parameter to ``Coordinates.find_nearest`` in case the differences between the radii are larger than zero when using spherical distance measures.

### DIFF
--- a/pyfar/classes/coordinates.py
+++ b/pyfar/classes/coordinates.py
@@ -1420,7 +1420,8 @@ class Coordinates():
 
         return ax
 
-    def find_nearest(self, find, k=1, distance_measure='euclidean'):
+    def find_nearest(self, find, k=1, distance_measure='euclidean',
+                     radius_tol=None):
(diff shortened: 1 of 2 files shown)


f-brinkmann (approved):
Thanks, looks good to me!

ahms5 (changes requested):
i think its a new feature, and we could put it into the comming feature release, so we can merge into dev. otherwise approved

ahms5 (approved):

PR #629, opened 2024-06-15
hotfix numpy>=2.0

### LINKED ISSUE #630: Import of `orientation` fails since release of Numpy v2.0
## General

* pyfar version: 0.6.6
* Python version: 3.12.2
* Operating System: MacOS Sonoma 14.5 + Ubuntu 22.04.4

## Description

Import of `orientations` module fails with Numpy v2.0. (released on 16.06.2024; see [Numpy v2.0.0 release](https://github.com/numpy/numpy/releases/tag/v2.0.0))

Line where exception is raised:
https://github.com/pyfar/pyfar/blob/e566619b976e589cf4a5b091d0902725fea9bb86/pyfar/classes/orientations.py#L10

log:
```
  File "<stdin>", line 1 in <module>
  File [some/path]/numpy/__init__.py": line 410, in __getattr__
    raise AttributeError("module {!r} has no attribute "
AttributeError: module 'numpy' has no attribute 'VisibleDeprecationWarning'
```

## What I Did

This can be fixed by importing `VisibleDeprecationWarning` from `numpy.exceptions` instead of importing `numpy`.

Alternatively, numpy version could be pinned in the requirement file to `<2.0` until fixes are made.

## Notes:
This suggests there could be other parts of the repo which are not compliant with Numpy v2.0.

ahms5:
- np.VisibleDeprecationWarning -> np.exceptions.VisibleDeprecationWarning https://numpy.org/devdocs/release/2.0.0-notes.html#numpy-2-0-python-api-removals
- tuple was returned as (np.int64(4), np.int64(5)) instead of (4, 5) it is fixed now.
- python 3.8 is not supported for numpy>=1.25, with this changes pyfar is still compatible to numpy 1.24..
- it was tests for numpy 2.0 nad numpy < 2.0 for all supported python version in pyfar

- installation of pyfar will not work with numpy == 2.0 for now, since the import of the ``VisibleDeprecationWarning`` will fail
fixes  #630

### DIFF
--- a/pyfar/classes/orientations.py
+++ b/pyfar/classes/orientations.py
@@ -3,11 +3,15 @@ import numpy as np
 import warnings
 
 import pyfar as pf
+if np.__version__ < '2.0.0':
+    from numpy import VisibleDeprecationWarning
+else:
+    from numpy.exceptions import VisibleDeprecationWarning
 
 # this warning needs to be caught and appears if numpy array are generated
 # from nested lists containing lists of unequal lengths, e.g.,
 #  [[1, 0, 0], [1, 0]]
-warnings.filterwarnings("error", category=np.exceptions.VisibleDeprecationWarning)
+warnings.filterwarnings("error", category=VisibleDeprecationWarning)
 
 
 class Orientations(Rotation):
@@ -92,7 +96,7 @@ class Orientations(Rotation):
         try:
             views = np.atleast_2d(views).astype(np.float64)
             ups = np.atleast_2d(ups).astype(np.float64)
-        except np.exceptions.VisibleDeprecationWarning:
+        except VisibleDeprecationWarning:
             raise ValueError("Expected `views` and `ups` to have shape (N, 3)")
 
         # check views and ups
--- a/pyfar/classes/orientations.py
+++ b/pyfar/classes/orientations.py
@@ -7,7 +7,7 @@ import pyfar as pf
 # this warning needs to be caught and appears if numpy array are generated
 # from nested lists containing lists of unequal lengths, e.g.,
 #  [[1, 0, 0], [1, 0]]
-warnings.filterwarnings("error", category=np.VisibleDeprecationWarning)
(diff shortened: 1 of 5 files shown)


f-brinkmann (approved):
Thanks for fixing!

f-brinkmann:
Maybe squash the commits before we merge?

pyfar/classes/orientations.py
```diff
@@ -3,11 +3,15 @@
 import warnings
 
 import pyfar as pf
+if np.__version__ < '2.0.0':
+    from numpy import VisibleDeprecationWarning
+else:
+    from numpy.exceptions import VisibleDeprecationWarning
```
mberz, line 9: This class has been moved to the exceptions module in numpy 1.25 (https://numpy.org/doc/stable/release/1.25.0-notes.html#numpy-now-has-an-np-exceptions-namespace), currently we're supporting versions >= 1.23.
So another option to avoid this would be raise the minimum supported numpy version to 1.25.
Downside is that numpy 1.25 also requires at minimum Python 3.9.

mberz (commented):
There's one thing that comes to mind on the numpy exception submodule:

This class has been moved to the exceptions module in numpy 1.25 (https://numpy.org/doc/stable/release/1.25.0-notes.html#numpy-now-has-an-np-exceptions-namespace), currently we're supporting versions >= 1.23.
So another option to avoid this would be raise the minimum supported numpy version to 1.25.
Downside is that numpy 1.25 also requires at minimum Python 3.9.

Maybe it's something better suited for the pyfar 0.7.0 release to avoid dropping support for Python < 3.9 in a patch release, so open for discussion.

f-brinkmann:
@mberz, @ahms5 maybe release this as is as a patch and discuss, if we want to restrict to numpy>=1.25 and python>=3.9 during the next weekly?

mberz:
Yes, that would be my suggestion as well. I'd prefer to not make the version jump in the minimum required numpy and hence Python version within a patch release.
@ahms5 So it's only the missing space 🤣

PR #631, opened 2024-06-17
Harmonize class documentation

### LINKED ISSUE #506: Improve docs and import structure
## General

* pyfar version: 0.5.4

## Description

- [x] Unify documentation of classes, especially of magic functions like `__init__`

f-brinkmann:
Some classes still used an explicit documentation of their magic methods, which I think is not good practive.

closes #506 

- Move documentation of `Orientation.__init__` to the class docstring and make interactive links
- Move documentation from Audio objects (Signal/TimeData/FrequencyData) `__init__` to the class docstring.
- Move Information from Audio objects `__getitem__`, `__setitem__`, and `__iter__` to the module dosctring.

In the latter case, we might consider adding an example notebook to the gallery instead. The information in the docs on this is rather sparse. But I did not want to add stuff in this pull.

### DIFF
--- a/docs/classes/pyfar.audio.rst
+++ b/docs/classes/pyfar.audio.rst
@@ -4,5 +4,4 @@ Audio (Signal, TimeData, FrequencyData)
 .. automodule:: pyfar.classes.audio
    :members:
    :undoc-members:
-   :special-members: [, __init__, __getitem__, __setitem__, __iter__, ]
    :inherited-members:
--- a/docs/classes/pyfar.orientations.rst
+++ b/docs/classes/pyfar.orientations.rst
@@ -4,5 +4,4 @@ Orientations
 .. automodule:: pyfar.classes.orientations
    :members:
    :undoc-members:
-   :special-members: __init__
    :show-inheritance:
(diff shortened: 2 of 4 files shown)


ahms5 (commented):
Do you have any reference why it is bad? I just found [references](https://docs.astral.sh/ruff/rules/undocumented-magic-method/) why you should do it, but they also dont really tell why. And i found [sth ](https://github.com/PyCQA/pydocstyle/issues/60#issuecomment-31206791) that says one should add docstrings for init functions.
(maybe you have discussed it at the weeky and i was not participating)

f-brinkmann:
Maybe the formulation was a little unlucky. I am very much in favour of documenting the behaviour of the magic methods but without showing, e.g., `__rmul__` and `__init__` explicitly in the docs and you don't use the magigc methods with `Signal.__init__()`. Moreover, those names are likely riddles to many new users. I think with the newer classes, we always document the init behaviour as part of the docstring and other methods like arithmetic operations in the module docstring or examples in the gallery.

This is not speaking against adding docstrings to magic methods, which I think is good practice to aid development.

sikersten (approved):

ahms5 (changes requested):
Thank you, maybe one point to discuss next time

f-brinkmann:
Maybe this is a good point for general discussion. If we keep it here, there are magic functions in other classes we should add to the docs for consistenty...

ahms5 (approved):

PR #640, opened 2024-06-26
fix deprecation warnings from matplotlib and numpy

ahms5:
fix deprecation warnings on main to avoid sth like #638 in future
requires #638

we could think about a test on circle ci which breaks if we get deprecation warnings:
```
pytest tests -W error::DeprecationWarning
```


f-brinkmann (approved):
Thanks for the effort.

PR #644, opened 2024-07-05
add deprecation test on cirlce ci

ahms5:
check on circle ci for deprecationswarnings in tests, as discussed in #640

requires #670


.circleci/config.yml
```diff
@@ -150,6 +173,14 @@ workflows:
           requires:
             - build_and_test
 
+      - test_deprecation_warnings:
+          matrix:
+            parameters:
+              version:
+                - "3.9"
```
mberz, line 180: I think this should be the latest aka most recent supported version. Otherwise issues are detected when it's already too late.

.circleci/config.yml
```diff
@@ -199,6 +230,20 @@ workflows:
             tags:
               only: /^v[0-9]+(\.[0-9]+)*$/
 
+      - test_deprecation_warnings:
+          matrix:
+            parameters:
+              version:
+                - "3.9"
```
mberz, line 237: I think this should be the latest aka most recent supported version. Otherwise issues are detected when it's already too late.

mberz (changes requested):

f-brinkmann (approved):
Approved, given the changes suggested by Marco are done

.circleci/config.yml
ahms5: I have changed all to 3.12

ahms5:
seems like the docker container python:latest is not the latest python version
https://hub.docker.com/layers/circleci/python/latest/images/sha256-573319853615e623acf89d325e3f3e2281891b5283f7c4d23fca8da84085ebdd?context=explore

ahms5:
So its not possible to use the latest docker container, @mberz do you have an other idea?

ahms5:
@mberz  maybe we should go for the github action, here we can test python version which are not released yet. on cirlce ci its not possible. maybe they also have a latest tag

mberz:
I guess the GitHub action did not work, given your last commit message? I would just set the version number manually then and stick with circleci for now.

Issue #627, opened 2024-06-14
Filtering `Signal` alters its `cshape`


artpelling:
## General

* pyfar version: 0.6.6
* Python version: 3.10.12
* Operating System: Ubuntu LTS 22.04.4
* Did you install pyfar via pip: yes

## Description
Filtering a `Signal` changes its `cshape`, if one dimension is `1`. I don't want my shapes being touched. Worse, it leads to incorrect broadcasting down the line.

## What I Did

MWE
``` python
#!/usr/bin/env python3

import numpy as np
import pyfar


sig = pyfar.Signal(np.zeros((4,1,10)), 1)
out = pyfar.dsp.filter.butterworth(sig, 2, 0.1)

print(sig)
print(out)
print(sig-out)
```
Output
``` bash
time domain energy Signal:
(4, 1) channels with 10 samples @ 1 Hz sampling rate and none FFT normalization

time domain energy Signal:
(4,) channels with 10 samples @ 1 Hz sampling rate and none FFT normalization

freq domain energy Signal:
(4, 4) channels with 10 samples @ 1 Hz sampling rate and none FFT normalization
```

PR #654, opened 2024-07-16
modify `shape` of return value

artpelling:
- remove `np.squeeze` in order to assure that the shape of the return type matches that of the signal

Working with pyfar, I've had my code fail at runtime because the shapes are inconsistent, especially in the `dsp` module. My code works for `n x m` channels, but when setting `n=1` or `m=1` sometimes dimensions are squeezed away which is a headache. Somewhat related: #627

### DIFF
--- a/pyfar/dsp/dsp.py
+++ b/pyfar/dsp/dsp.py
@@ -1430,7 +1430,7 @@ def find_impulse_response_start(
                     f'No values below threshold found found for channel {ch}',
                     'defaulting to 0')
 
-    return np.squeeze(start_sample)
+    return start_sample
 
 
 @rename_arg({"freq_range": "frequency_range"},


f-brinkmann:
I think this does not closely relate to #627? I think in this case, we were following numpy behavior, which often is to squeeze singular dimensions from returned arrays. But I see how it is annoying in this case. Maybe this is worth a general discussion.

artpelling:
> I think this does not closely relate to #627? I think in this case, we were following numpy behavior, which often is to squeeze singular dimensions from returned arrays. But I see how it is annoying in this case. Maybe this is worth a general discussion.

It is not directly related, but another instance of `dsp` changing shapes. I linked it, because they could be discussed together maybe.

ahms5:
I would also go for an ``keepdims`` parameter.

artpelling:
I am not convinced, yet..
If I understand correctly, you suggest to add `keepdims` parameter to all/certain `dsp` functions. I think that would lose a lot of convenience and also add more potential for incompatible shapes when pipelining multiple `dsp` functions.
I strongly feel that the `cshapes` of all `Signals` should always be preserved in order to assure compatibility and lower the potential for bugs like in #627.

I understand that an unnecessary dimension can be annoying and I also have cases where I prefer to work without it.
In these cases, I just squeeze the `Signals` on creation to avoid this. @f-brinkmann, @ahms5, what do you think about adding this `keepdims` functionality to the `Signal` constructor instead?

ahms5:
Yes you are right, I also think it should keep its cshapes. mutliplication of numpy arrays does the same. the problem is just the backwards compability. 
Another option would be to also add a separate function ``Signal.squeeze()`` that does this if the user needs it.
I think we will discuss this on Friday.

f-brinkmann:
This is now part of #659 and will be merged soon

PR #655, opened 2024-07-18
update signal iterator

tluebeck:

Closes #

-
-
-

### DIFF
--- a/tests/test_audio_signal_iterator.py
+++ b/tests/test_audio_signal_iterator.py
@@ -4,24 +4,11 @@ import numpy.testing as npt
 from pyfar import Signal
 
 
-def test_simple_iter():
+@pytest.mark.parametrize('is_complex', [False, True])
+def test_simple_iter(is_complex):
     data = np.ones((3, 1024)) * np.atleast_2d(np.arange(3)).T
     sig = Signal(data, 1)
-
-    idx = 0
-    for s in sig:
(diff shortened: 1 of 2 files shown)


tests/test_audio_signal_iterator.py
```diff
@@ -17,6 +17,20 @@ def test_simple_iter():
         npt.assert_array_equal(s._data, np.atleast_2d(data[idx, :]))
 
 
+def test_complex_simple_iter():
```
f-brinkmann, line 20:
```suggestion
@pytest.mark.parametrize('is_complex', [False, True])
def test_complex_simple_iter():
```

tests/test_audio_signal_iterator.py
```diff
@@ -17,6 +17,20 @@ def test_simple_iter():
...
+    data = np.ones((3, 1024)) * np.atleast_2d(np.arange(3)).T
+    sig = Signal(data, 1)
+    sig.complex = True
```
f-brinkmann, line 23:
```suggestion
    sig.complex = is_complex
```

f-brinkmann (changes requested):
Thanks for catching. I would suggest to parametrize the existing test to minimize duplicate code

tests/test_audio_signal_iterator.py
tluebeck: good point thank you :)

PR #656, opened 2024-07-21
Add complex 2d plots

tluebeck:

Closes #

-
-
-


tests/test_plot.py
```diff
@@ -315,6 +315,18 @@ def test_2d_plots(function, handsome_signal_2d):
                      file_type, compare_output)
 
 
+@pytest.mark.parametrize('function', [
+    (plot.time_2d), (plot.freq_2d), (plot.phase_2d), (plot.group_delay_2d),
+    (plot.time_freq_2d), (plot.freq_phase_2d), (plot.freq_group_delay_2d)])
+def test_2d_plots_complex(function, handsome_signal_complex_2d):
+    """Test all 2d plots with default arguments."""
+    filename = function.__name__
```
ahms5, line 323:
```suggestion
    filename = f'complex_2d_{function.__name__}'
```
I think otherwise it will cmp vs the wrong files.

ahms5 (changes requested):
Thanks, it think after adding baseline it should be fine.

pyfar/plot/line.py
```diff
@@ -193,6 +193,8 @@ def phase(signal, deg=False, unwrap=False, freq_scale='log', ax=None,
         ``style = {'axes.facecolor':'black'}``. Pass an empty dictionary
         ``style = {}`` to use the currently active plotstyle. The default is
         ``light``.
+    side : str
+        ``left`` or ``right`` to specify which side of a spectrum is plotted.
```
f-brinkmann, line 197: I would suggest to give a little more information and also mention the default here and everywhere else:

```suggestion
    side : str, optional
        ``'right'`` to plot the right-sided spectrum containing the positive
        frequencies, or ``'left'``to plot the left-sided spectrum containing
        the negative frequencies (only possible for complex Signals). The
        default is ``'right'``.
```

pyfar/plot/line.py
```diff
@@ -358,6 +362,12 @@ def time_freq(signal, dB_time=False, dB_freq=True, log_prefix_time=20,
         ``style = {'axes.facecolor':'black'}``. Pass an empty dictionary
         ``style = {}`` to use the currently active plotstyle. The default is
         ``light``.
+    mode : str
+        ``real``, ``imag``, or ``abs`` to specify if the real part, imaginary
+        part or absolute value of a complex-valued time domain signal is
+        plotted. The default is ``real``.
```
f-brinkmann, line 368: I would mention that imag and abs can not be used for non-complex Signals here and everywhere else:

```suggestion
    mode : str
        ``'real'``, ``'imag'``, or ``'abs'`` to specify if the real part, imaginary
        part or absolute value of the time data is plotted. ``'imag'`` and ``'abs'```
        can only be used for complex Signals. The default is ``'real'``.
```

f-brinkmann (changes requested):
+1 for changing the file names of the baseline plots as suggested by ahms. Apart from that I only suggest to be more verbose in the docs - thanks a lot!

ahms5 (approved):
looks good to me, thank you, we just need to set ``create_baseline`` to ``False`` again.

ahms5 (approved):

f-brinkmann (approved):

PR #659, opened 2024-08-01
Fix squeeze in dsp functions to keep cshape of input signal

hoyer-a:
Adjusted the dsp-functions so signals won't be squeezed and the original cshape is kept.

separate PR for dsp.spectrogram #666


hoyer-a:
Harmonize functions so the cshape of the input signal is always kept even for signals of cshape = (1, )

tests/test_dsp.py
```diff
@@ -111,6 +111,18 @@ def test_group_delay_custom_frequencies(impulse_group_delay):
         grp, impulse_group_delay[1][0, frequency_idx], atol=1e-10)
 
 
+@pytest.mark.parametrize("shape", [(4, 1), (1, 4)])
+def test_group_delay_cshape(shape):
+    """Test if group delay function keeps cshape of signals of shape `m x n`
+     with `m = 1` or `n = 1`."""
+    delay = np.random.randint(0, 200, (shape))
+    impulse = pf.signals.impulse(256, delay, np.ones(shape))
```
ahms5, line 119:
```suggestion
    impulse = pf.signals.impulse(256, 64, np.ones(shape))
```
better would be not to use random inside tests.

tests/test_dsp.py
```diff
@@ -111,6 +111,18 @@ def test_group_delay_custom_frequencies(impulse_group_delay):
...
+
+    grp = dsp.group_delay(impulse)
+
+    assert grp.shape[:2] == impulse.cshape
```
ahms5, line 123:
```suggestion
    assert grp.shape[:-1] == impulse.cshape
```
is safer

tests/test_dsp.py
```diff
@@ -111,6 +111,18 @@ def test_group_delay_custom_frequencies(impulse_group_delay):
         grp, impulse_group_delay[1][0, frequency_idx], atol=1e-10)
 
 
+@pytest.mark.parametrize("shape", [(4, 1), (1, 4)])
```
ahms5, line 114:
```suggestion
@pytest.mark.parametrize("shape", [(4, 1), (1, 4), (1, ), (1, 1)])
```
i think input (1,) should also return (1,..). maybe we need to discuss. in the doc its mentioned : "The array is flattened if a single channel signal was passed to the function."

tests/test_dsp.py
```diff
@@ -654,6 +666,16 @@ def test_impulse_response_delay_multidim():
     npt.assert_allclose(start_sample_est, start_sample, atol=1e-2)
 
 
+@pytest.mark.parametrize("shape", [(4, 1), (1, 4)])
```
ahms5, line 669:
```suggestion
@pytest.mark.parametrize("shape", [(4, 1), (1, 4), (1, ), (1, 1)])
```

tests/test_dsp.py
```diff
@@ -756,6 +778,16 @@ def test_impulse_response_start_multidim():
     npt.assert_allclose(start_sample_est, start_samples - 1)
 
 
+@pytest.mark.parametrize("shape", [(4, 1), (1, 4)])
```
ahms5, line 781:
```suggestion
@pytest.mark.parametrize("shape", [(4, 1), (1, 4), (1, ), (1, 1)])
```

ahms5 (changes requested):
thank you for taking care. I think we need to decide if we want cshape (1,) will be squeezed towards (256) for group delays. I would suggest to also leave it as (1, 256). See below

tests/test_dsp.py
hoyer-a: Yeah, i remember that we talked about keeping the input shape in any case for all the dsp functions.
I am on it!

pyfar/dsp/dsp.py
```diff
@@ -116,9 +115,10 @@ def group_delay(signal, frequencies=None, method='fft'):
         raise ValueError(
             "Invalid method, needs to be either 'scipy' or 'fft'.")
 
-    # flatten in numpy fashion if a single channel is returned
-    if signal.cshape == (1, ):
-        group_delay = np.squeeze(group_delay)
+    # squeeze last dimension. If only one frequency is given, shape (1, ) is
+    # returned instead of (1, 1)
+    if frequencies is not None and frequencies.size == 1:
+        group_delay = np.squeeze(group_delay, axis=-1)
```
ahms5, line 121:
```suggestion
```
not sure why we squeeze the last dimension.

pyfar/dsp/dsp.py
```diff
@@ -67,8 +67,7 @@ def group_delay(signal, frequencies=None, method='fft'):
     Returns
     -------
     group_delay : numpy array
-        Frequency dependent group delay in samples. The array is flattened if
-        a single channel signal was passed to the function.
+        Frequency dependent group delay in samples.
```
ahms5, line 70:
```suggestion
        Frequency dependent group delay in samples of shape ``(*signal.cshape, #frequencies)``.
```
maybe something like this? I would not squeeze the dimension of the frequency if its just one.

ahms5 (changes requested):
Thank you, just a small comment about squeezing.

pyfar/dsp/dsp.py
hoyer-a: Probably not really nescessary, but some tests failed without it and i thought it would be nicer like that.

With the squeeze we get rid of the singular frequency dimension, similar to when we squeeze singular filter channels in the filter classes. Otherwise a single channel with a single frequency given would be returned as an array of shape (1, 1).

hoyer-a: I can remove it. 

I think some tests failed without it and i thought it would be similar to when we squeeze singular filter dimensions in the filter classes.

f-brinkmann: We decided not to squeeze and adopt the tests.

tests/test_dsp.py
```diff
@@ -106,9 +106,20 @@ def test_group_delay_custom_frequencies(impulse_group_delay):
...
+
+@pytest.mark.parametrize("shape", [(4, 1), (1, 4), (1, ), (1, 1)])
+def test_group_delay_cshape(shape):
+    """Test if group delay function keeps cshape of signals of shape `m x n`
+     with `m = 1` or `n = 1`."""
+    impulse = pf.signals.impulse(256, 0, np.ones(shape))
+
+    grp = dsp.group_delay(impulse)
+
+    assert grp.shape == (*impulse.cshape, grp.shape[-1])
```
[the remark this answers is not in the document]
hoyer-a: I changed it to `impulse.n_bins)` since the last dimension of group delay is frequencies.

ahms5 (approved):

tests/test_dsp.py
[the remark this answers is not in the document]
ahms5: even better

PR #664, opened 2024-08-13
allow pf.signals.impulse(3, 0, np.ones((1,1))

ahms5:
- pf.signals.impulse(3, 0, np.ones((1,1)) was not working
- bug is fixed and tests are added

### DIFF
--- a/pyfar/signals/deterministic.py
+++ b/pyfar/signals/deterministic.py
@@ -359,7 +359,7 @@ def _match_shape(*args):
     shape = (1, )
     for arg in args:
         arg = np.asarray(arg)
-        if arg.size > size:
+        if (arg.size > size) or (arg.ndim > 1):
             size = arg.size
             shape = arg.shape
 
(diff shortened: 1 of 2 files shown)


f-brinkmann (changes requested):
Thanks for catching, but I think the solution needs to be more complex. In the current case `size` and `shape` are overwritten each time `arg.ndim > 1` and hence `pf.signals.impulse(10, np.ones((2, 2)), np.ones((1, 1)))` fails, because the last arguments determines the shape.

Would it make sense to first check for the largest ndim and check for the largest size per dimension in a second loop?

ahms5:
update:
- using ``np.broadcast_shapes`` directly
- add tests for ``pyfar.signals.sine`` and reviesed doc
- add tests for ``pyfar.signals.impulse`` and reviesed doc

f-brinkmann (approved):

hoyer-a (approved):

PR #668, opened 2024-08-21
implement _return_item for complex signals

tluebeck:

Closes #

-
-
-

### DIFF
--- a/tests/test_audio_signal.py
+++ b/tests/test_audio_signal.py
@@ -395,6 +395,7 @@ def test_magic_getitem_complex():
     """Test slicing operations by the magic function __getitem__."""
     time = np.arange(2 * 3 * 4, dtype=complex).reshape((2, 3, 4))
     signal = Signal(time, 44100, domain='time', is_complex=True)
+    assert signal[0].complex
     npt.assert_allclose(signal[0]._data, time[0])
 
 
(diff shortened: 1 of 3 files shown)


tests/test_audio_signal.py
```diff
@@ -391,6 +391,13 @@ def test_magic_getitem():
     npt.assert_allclose(signal[0]._data, time[0])
 
 
+def test_magic_getitem_complex():
+    """Test slicing operations by the magic function __getitem__."""
+    time = np.arange(2 * 3 * 4, dtype=complex).reshape((2, 3, 4))
+    signal = Signal(time, 44100, domain='time', is_complex=True)
+    npt.assert_allclose(signal[0]._data, time[0])
```
ahms5, line 398: we could also add a test for TimeData

ahms5 (changes requested):
uhh we are coming closer. Thank you! just one small comment

tests/test_audio_signal.py
f-brinkmann: +1 and I would assert `signal[0].complex` because

```
import numpy as np
import numpy.testing as npt

npt.assert_allclose(np.array([1, 1, 1], dtype=complex), [1, 1, 1])
```

does not assert if the Signal is complex

ahms5 (approved):

PR #670, opened 2024-08-21
fix deprecation warnings

ahms5:
### fixes 2 deprecation warnings which came up in #644
```
FAILED tests/test_coordinates_new.py::test__array__getter[x6-10--1] - DeprecationWarning: __array__ implementation doesn't accept a copy keyword, so passing copy=False failed. __array__ must implement 'dtype' and 'copy' keyword arguments.
FAILED tests/test_signals.py::test_sine_different_cshapes[None-None-frequency_shape1] - DeprecationWarning: Passing None into shape arguments as an alias for () is deprecated.
```


hoyer-a (approved):
Looks good to me!

PR #666, opened 2024-08-17
fix squeeze in dsp.spectrogram to keep cshape

hoyer-a:
separate pull to fix squeeze in dsp.spectrogram to keep cshape of input signal

merges into #659

PR #671, opened 2024-08-22
Plot complex spectogram

tluebeck:

Closes #

-
-
-


ahms5 (approved):

f-brinkmann (changes requested):
Thanks for the effort. Found a typo and an error message that might not be needed

pyfar/plot/_utils.py
```diff
@@ -457,6 +457,34 @@ def _assert_and_match_data_to_side(data, signal, side):
...
+        mask = frequencies <= 0
+    elif side == 'right':
+        mask = frequencies >= 0
+    else:
+        raise ValueError('Invalid `side` parameter, pass either `left` or '
+                         '`right`.')
+
+    # get corresponding data
+    frequencies = frequencies[mask]
+    spectrogram = spectrogram[mask, ...]
```
f-brinkmann, line 474: I think this needs to be changed for multi-dimensional signals:

If `signal.cdim == 1` this works, but if for example `signal.cdim == 2` the frequencies are in the third dimension. Is this related to pull requests for maintaining the cshape?

f-brinkmann (changes requested):
I found one point, where I was not sure...

pyfar/plot/_utils.py
tluebeck: The [..., mask, :] thing works for cdim > 1 :)

f-brinkmann: And should also work for cdim == 1 after merging #666, right?

tluebeck: yes, that should work

f-brinkmann (approved):

PR #674, opened 2024-08-23
Disable getting specific samples or frequencies from a signal through slicing

f-brinkmann:
- [x] As discussed during the weekly, I forced the `__getitem__` methods of the audio classes to throw an error.
- [x] Update tests accordingly
- [x] Opened issue in gallery to make sure we update the notebooks once pyfar 0.6.9 is released:  https://github.com/pyfar/gallery/issues/79

### DIFF
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -166,7 +166,11 @@ class _Audio():
 
 
         """
+        length = self._data.shape[-1]
         data = self._data[key]
+        if length != np.atleast_1d(data).shape[-1]:
+            raise KeyError((f'Indexed dimensions must not exceed the channel '
+                            f'dimension (cdim), which is {len(self.cshape)}'))
         return self._return_item(data)
 
     def __setitem__(self, key, value):
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -166,11 +166,23 @@ class _Audio():
 
 
         """
-        length = self._data.shape[-1]
-        data = self._data[key]
-        if length != np.atleast_1d(data).shape[-1]:
-            raise KeyError((f'Indexed dimensions must not exceed the channel '
-                            f'dimension (cdim), which is {len(self.cshape)}'))
+
+        # add empty slice at the end to always get all data contained in last
+        # dimension (samples or frequency bins)
+        if hasattr(key, '__iter__'):
+            key = (*key, slice(None))
+
+        # try indexing and raise verbose errors if it fails
+        try:
+            data = self._data[key]
+        except IndexError as Error:
+            if 'out of bounds' in str(Error):
+                raise Error
+            else:
+                raise IndexError((
+                    f'Indexed dimensions must not exceed the channel '
+                    f'dimension (cdim), which is {len(self.cshape)}'))
+
         return self._return_item(data)
 
     def __setitem__(self, key, value):
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -176,12 +176,12 @@ class _Audio():
         try:
             data = self._data[key]
         except IndexError as Error:
-            if 'out of bounds' in str(Error):
-                raise Error
-            else:
+            if 'too many indices for array' in str(Error):
                 raise IndexError((
                     f'Indexed dimensions must not exceed the channel '
                     f'dimension (cdim), which is {len(self.cshape)}'))
+            else:
+                raise Error
 
         return self._return_item(data)
 
--- a/tests/test_audio_frequency_data.py
+++ b/tests/test_audio_frequency_data.py
@@ -150,6 +150,20 @@ def test_magic_getitem_slice():
     npt.assert_allclose(FrequencyData(data[0], freqs)._data, freq[0]._data)
 
 
+def test_magic_getitem_error():
+    """
+    Test if indexing that would return a subset of the frequency bins raises a
+    key error.
+    """
+    freq = pf.FrequencyData([[0, 0, 0], [1, 1, 1]], [0, 1, 3])
+    # manually indexing too many dimensions
+    with pytest.raises(KeyError, match='Indexed dimensions must not exceed'):
+        freq[0, 1]
+    # indexing too many dimensions with ellipsis operator
+    with pytest.raises(KeyError, match='Indexed dimensions must not exceed'):
+        freq[..., 1]
+
+
 def test_magic_setitem():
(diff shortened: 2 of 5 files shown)


pyfar/classes/audio.py
```diff
@@ -166,7 +166,11 @@ def __getitem__(self, key):
 
 
         """
+        length = self._data.shape[-1]
```
ahms5, line 169:
```suggestion
        length = self._data.shape[-1]
        if hasattr(key, '__iter__'):
            key = (*key, slice(None, None, None))
```

ahms5 (commented):
maybe this will work. As far as i see there is no error by default if the there are to many slince arguments and ``:`` is at the end.

ph-schaefer:
Regarding the usage of indicing samples/bins in the gallery and tests:
I guess the clean way, would be to extract the data using obj.time[...] or obj.freq[...]. If it is still required to have an audio object return, one could think about a utils function that extracts samples or bins.

pyfar/classes/audio.py
f-brinkmann: I had a working solution already, but I had to update the tests, which is now done.

ahms5: it does not fix the problem:
```
import pyfar as pf
signal = pf.Signal([[[0, 0, 0, 0, 0], [1, 1, 1, 1, 1]]], 1)
signal.cshape
signal[..., 1]
```
``signal[..., 1]``still does not work

ahms5 (changes requested):

pyfar/classes/audio.py
ahms5: therefore I added the ``slice(None, None, None)``, then signal[..., 1] would return the correct data signal[:, 1]

f-brinkmann: Ah, now I got what you ment. Should be fine now.

ahms5 (approved):

pyfar/classes/audio.py
```diff
@@ -166,7 +166,23 @@ def __getitem__(self, key):
...
-        data = self._data[key]
+
+        # add empty slice at the end to always get all data contained in last
+        # dimension (samples or frequency bins)
+        if hasattr(key, '__iter__'):
+            key = (*key, slice(None))
+
+        # try indexing and raise verbose errors if it fails
+        try:
+            data = self._data[key]
```
ph-schaefer, line 177: Unfortunately, this still does not work for all cases:
When accessing an object with more than one channel dimension, and accessing a single slice of a channel, this dimension is squeezed. I wrote a unit test (see `test_audio_frequency_data.py`), which invokes this error.

tests/test_audio_frequency_data.py
```diff
@@ -150,6 +150,20 @@ def test_magic_getitem_slice():
...
+    key error.
+    """
+    freq = pf.FrequencyData([[0, 0, 0], [1, 1, 1]], [0, 1, 3])
+    # manually indexing too many dimensions
+    with pytest.raises(IndexError, match='Indexed dimensions must not exceed'):
+        freq[0, 1]
+    # indexing too many dimensions with ellipsis operator
+    with pytest.raises(IndexError, match='Indexed dimensions must not exceed'):
+        freq[0, 0, ..., 1]
+
```
ph-schaefer, line 165: Additional unit test:
```suggestion

@pytest.mark.parametrize("shape", [(1,2), (1,2,3), (1,2,3,4)])
def test_magic_getitem_cannot_slice_freq(shape):
    frequencies = range(shape[-1])
    data = pf.FrequencyData(np.ones(shape), frequencies)

    last_cdim_slice = data[..., 0].freq
    equivalent_freq_slice = data.freq[...,0:1,:]
    last_datadim_slice = data.freq[..., 0] #= slice at specific frequency bin

    target_cdim_slice_shape = list(shape)
    target_cdim_slice_shape[-2] = 1

    assert(last_cdim_slice.shape != last_datadim_slice.shape)
    assert(last_cdim_slice.shape == equivalent_freq_slice.shape)
    assert(last_cdim_slice.shape == tuple(target_cdim_slice_shape))
```

ph-schaefer (changes requested):
Thanks for this implementation.
Unfortunately, there are still some cases, which are not handled correctly (see detailed comments below)

tests/test_audio_frequency_data.py
ahms5: I think the test is wrong: ``0:1`` is not equal to ``0`` thats why its failing. if you fix line 171 its everthing is working as expected. If you slice with ``0`` the dimension is removed.
```suggestion
@pytest.mark.parametrize("shape", [(1,2), (1,2,3), (1,2,3,4)])
def test_magic_getitem_cannot_slice_freq(shape):
    frequencies = range(shape[-1])
    data = pf.FrequencyData(np.ones(shape), frequencies)
    last_cdim_slice = data[..., 0:1].freq
    equivalent_freq_slice = data.freq[...,0:1,:]
    last_datadim_slice = data.freq[..., 0] #= slice at specific frequency bin
    target_cdim_slice_shape = list(shape)
    target_cdim_slice_shape[-2] = 1
    assert(last_cdim_slice.shape != last_datadim_slice.shape)
    assert(last_cdim_slice.shape == equivalent_freq_slice.shape)
    assert(last_cdim_slice.shape == tuple(target_cdim_slice_shape))
```

f-brinkmann: I just checked that as well. This also pertains to the `target_cdim_sice_shape`.  This would mean that the cshape can change due to slicing. If we consider this expected behaviour were fine. If not were in trouble...

ph-schaefer: Sorry, my bad. I misinterpreted parts of the discussions last week.
Nevertheless, I think keeping the unit test is good, since it tests data with higher number of channel dimensions.

f-brinkmann: I'm alread testing it here for 2D signals:
```python
def test_magic_getitem_ellipsis():
    """Test slicing operations by the magic function __getitem__."""
    signal = pf.Signal([[[1, 1, 1], [2, 2, 2]]], 44100)
    npt.assert_allclose(signal[..., 0].time, np.atleast_2d([1, 1, 1]))
    assert signal[..., 0].time.shape == (1, 3)
```

f-brinkmann: We do have tests for 2D signals in there, maybe thats already enough.

PR #680, opened 2024-09-02
add arithmetics for coordinates

### LINKED ISSUE #632: Arithmetic operations with pf.Coordinates
It would be nice to be able to use arithmetic operations with pf.Coordinates, similar to audio objects. 
ideas would be:
- [x] addition of 2 Coordinates objects
- [x] substraction of 2 Coordinates objects
- [x] dot product
- [x] cross product

ahms5:

Closes #632

- [x] addition of 2 Coordinates objects
- [x] substraction of 2 Coordinates objects
- [x] dot product
- [x] cross product
- [x] multiply and devide with numbers
- [x] add doc at the beginning

### DIFF
--- a/pyfar/classes/coordinates.py
+++ b/pyfar/classes/coordinates.py
@@ -10,8 +10,12 @@ Different coordinate systems are frequently used in acoustics research and
 handling sampling points and different systems can be cumbersome. The
 Coordinates class was designed with this in mind. It stores coordinates in
 cartesian coordinates internally and can convert to all coordinate systems
-listed below. Additionally, the the class can  query and plot coordinates
-points. Functions for converting coordinates not stored in a Coordinates object
+listed below. Additionally, the the class can query and plot coordinates
+points. Addition and subtraction are supported with numbers and Coordinates
+objects, while multiplication and division are supported with numbers only.
+All arithmetic operations are performed element-wise on Cartesian coordinates
+using the appropriate operator.
+Functions for converting coordinates not stored in a Coordinates object
 are available for convenience. However, it is strongly recommended to use the
 Coordinates class for all conversions.
 
--- a/pyfar/classes/coordinates.py
+++ b/pyfar/classes/coordinates.py
@@ -2881,6 +2881,14 @@ def dot(a, b):
     -------
     result : np.ndarray
         array with the dot product of the two objects
+
+    Examples
+    --------
+    >>> import pyfar as pf
+    >>> a = pf.Coordinates(1, 0, 0)
+    >>> b = pf.Coordinates(1, 0, 0)
+    >>> pf.dot(a, b)
+    array([1.])
     """
 
     if not isinstance(a, Coordinates) or not isinstance(b, Coordinates):
@@ -2910,6 +2918,15 @@ def cross(a, b):
     -------
     result : pf.Coordinates
         new Coordinates object with the cross product of the two objects
+
+    Examples
+    --------
+    >>> import pyfar as pf
+    >>> a = pf.Coordinates(1, 0, 0)
+    >>> b = pf.Coordinates(0, 1, 0)
+    >>> result = pf.cross(a, b)
+    >>> result.cartesian
+    array([0., 0., 1.])
     """
 
     if not isinstance(a, Coordinates) or not isinstance(b, Coordinates):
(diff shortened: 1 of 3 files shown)


pyfar/classes/coordinates.py
```diff
@@ -2650,6 +2650,11 @@ def _set_points(self, x, y, z):
         y = np.broadcast_to(y, shapes)
         z = np.broadcast_to(z, shapes)
 
+        # set writeable
```
f-brinkmann, line 2653: Maybe be more informative in the comment and add why this is needed.

pyfar/classes/coordinates.py
```diff
@@ -2807,12 +2807,146 @@ def __eq__(self, other):
...
+    for obj in [first, second]:
+        if isinstance(obj, Coordinates):
+            data.append(obj.cartesian)
+            num_objects += 1
+        elif isinstance(obj, (int, float)):
+            data.append(np.array(obj))
+        else:
+            op = 'Addition' if operation == 'add' else 'Subtraction'
+            raise TypeError(
+                f"{op} is only possible with Coordinates or number.")
```
f-brinkmann, line 2926: This could also be reached for operations 'mul' and 'div' - why are they not considered for throwing the error? In addition, the docstring states that arrays can be used, but currently I think this would raise the TypeError as well. Maybe we can discuss if there are use cases for array input, or if any array data must be passed as coordintates.

pyfar/classes/coordinates.py
```diff
@@ -2807,12 +2807,146 @@ def __eq__(self, other):
...
+
+    def __rtruediv__(self, other):
+        """Divide two Coordinates objects."""
+        return _arithmetics(other, self, 'div')
+
+    def __rdiv__(self, other):
+        """Divide two Coordinates objects."""
+        return _arithmetics(other, self, 'div')
+
+    def dot(self, other):
```
f-brinkmann, line 2850: This parameter name might be more illustrative. To me `'other'` sounds a little like dev-speak:
```suggestion
    def dot(self, coordinates):
```

pyfar/classes/coordinates.py
```diff
@@ -2807,12 +2807,146 @@ def __eq__(self, other):
...
+        """Dot product two Coordinates objects."""
```
f-brinkmann, line 2851: It would be nice if the docstrings for `dot` and `cross` would

- mention that the data must be broadcastable
- give the formula to make it clear that this is operation is (and can only) done on the cartesian coordinates.

pyfar/classes/coordinates.py
```diff
@@ -2807,12 +2807,146 @@ def __eq__(self, other):
...
+
+        if not isinstance(other, Coordinates):
+            raise TypeError(
+                "Dot product is only possible with Coordinates objects.")
+
+        return self.x * other.x + self.y * other.y + self.z * other.z
```
f-brinkmann, line 2857: For discussion: should we encapsule all arithmetic operations and broadcasting in `try` statements to raise more informative errors?

tests/test_coordiantes_arithmetic.py
```diff
@@ -0,0 +1,143 @@
...
+    assert isinstance(new, pf.Coordinates)
+    npt.assert_array_equal(new.cartesian, desired.cartesian)
+
+
+
+@pytest.mark.parametrize(
+    "other",
+    (
+        1,
+        1.0,
```
f-brinkmann, line 55: I would suggest non-unitary operators for testing. Multiplication and division by 1 does not change the numbers, which makes it hard to test if the operation was actually applied.

f-brinkmann (changes requested):
Thanks for the effort. This is a nice feature. I have one general question for discussion: For audio objects, we decided to not use class-methods for the operations. I would tend to do the same for coordinates and make `dot` and `cross` available as `pyfar.dot` and `pyfar.cross`.

pyfar/classes/coordinates.py
```diff
@@ -2756,11 +2761,6 @@ def __getitem__(self, index):
 
         return new
```
ahms5, line 2763:
```suggestion
   def __array__(self, copy=True, dtype=None):
        """Instances of Coordinates behave like `numpy.ndarray`, array_like."""
        # copy to avoid changing the coordinate system of the original object
        return np.array(self.cartesian, copy=copy, dtype=dtype)


```

hoyer-a (approved):
Thanks, looks good to me!

f-brinkmann (approved):

hoyer-a (approved):

PR #636, opened 2024-06-25
Feature/transmission matrix

ph-schaefer:
## Feature: New class representing transmission matrices

A transmission matrix (T-matrix) is used to model a two-port or quadrupole, which can represent electrical circuits, mechanical systems (e.g. loudspeakers), acoustic transmission lines, and so on. Systems can be cascaded by multiplying matrices. Functions are provided to calculate the input and output impedance as well as transfer functions.

### Extension towards acoustic transmission lines and other applications
My personal goal is to create a class / module that allows to handle and cascade different type of transmission line segments, which are internally represented by T-matrices. Generally, other applications are possible (e.g. loudspeaker modeling), but would be beyone the scope of my implementation.
- Should such a representation be part of pyfar or a new package? -> not pyfar
- Could be a new pyfar module? -> for now personal module

### Numpy 1.24 (Python 3.8) and logical indicing
Numpy >= 1.25 does not support Python 3.8 anymore (see [here](https://endoflife.date/numpy)). I had some issues with numpy 1.24 (latest Version for Python 3.8) regarding logical indicing. I found a work around but generally logical indicing seems to a little more tricky in older versions.
For higher Python Versions, numpy 2.x was used and my original code worked. I did not check higher numpy 1.XX versions.

### Notes
- A change of the internal `_arithmetic()` method of the `_Audio` class was required

PR #682, opened 2024-09-04
Replace scipy's legacy spectrogram with new implementaton

### LINKED ISSUE #667: `dsp.spectrogram` relies on scipy's `spectrogram` which is considered legacy
## General

* pyfar version: 0.6.8

## Description

`dsp.spectrogram()` relies on [scipy.signal.spectrogram](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.spectrogram.html) which is considered legacy. 

The use of [scipy.signal.ShortTimeFFT](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.ShortTimeFFT.html#scipy.signal.ShortTimeFFT) is recommended.

hoyer-a:
Replace [scipy.signal.spectrogram](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.spectrogram.html#scipy.signal.spectrogram) with [scipy.signal.ShortTimeFFT](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.ShortTimeFFT.html#scipy.signal.ShortTimeFFT) and new [spectrogram-method](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.ShortTimeFFT.spectrogram.html#scipy.signal.ShortTimeFFT.spectrogram).

Based on: https://docs.scipy.org/doc/scipy/tutorial/signal.html#short-time-fourier-transform

Closes #667


ahms5 (dismissed):
if we decide to kick python 3.8 out, I think its time.
https://docs.scipy.org/doc/scipy-1.13.0/release/1.12.0-notes.html

f-brinkmann (changes requested):
Can you please check if the Python 3.8 tests still work if you restrinct scipy in setup.py to the version where `ShortTimeFFT` was introduced? If not I would also be in favor of dropping Python 3.8 support.

f-brinkmann:
@tluebeck, once this is merged, it will probably give a conflict with the develop signal branch

ahms5:
> Can you please check if the Python 3.8 tests still work if you restrinct scipy in setup.py to the version where `ShortTimeFFT` was introduced? If not I would also be in favor of dropping Python 3.8 support.

maybe that wouldn't even work, (and if, we should leave it anyway) since they have stopped supporting 3.8 with the same version.

f-brinkmann:
> maybe that wouldn't even work, (and if, we should leave it anyway) since they have stopped supporting 3.8 with the same version.

Ah, ok. In this case I would vote for dropping 3.8 but merging this into develop for release with 0.7.0. Might be a bit unexpected to deprecate a Python version in a patch release.

sikersten:
+1 for dropping 3.8. with 0.7.0 release. I think this might also affect #636, @ph-schaefer

f-brinkmann:
We decided to deprecate Python 3.8 in pyfar 0.7.0. This will be a separate pull.

ahms5 (approved):
requires #683

f-brinkmann (approved):

PR #684, opened 2024-09-07
Show paths as imported in pyfar

f-brinkmann:

We discussed in the weekly that it would be better if the paths in the docs refelct the import structure of pyfar, which is not always the case, e.g. `pyfar.Signal` shows up in the docs as `pyfar.classes.audio.Signal`.

The paths are correct now in the new documentation and there are only minor changes to the layout

- The tables listing all classes and functions look slightly different because we are now using `autosummary` to generate them manually. Before, they were generated by `automodule`
- I removed the table listing classes for the Orientations class, because it is the only class on that page.

### DIFF
--- a/docs/classes/pyfar.coordinates.rst
+++ b/docs/classes/pyfar.coordinates.rst
@@ -13,12 +13,12 @@ Coordinates
 
 .. autosummary::
 
-   pyfar.classes.coordinates.cart2cyl
-   pyfar.classes.coordinates.cart2sph
-   pyfar.classes.coordinates.cyl2cart
+   cart2cyl
+   cart2sph
+   cyl2cart
    pyfar.deg2rad
    pyfar.rad2deg
-   pyfar.classes.coordinates.sph2cart
+   sph2cart
 
 .. autoclass:: pyfar.Coordinates
    :members:
--- a/docs/classes/pyfar.coordinates.rst
+++ b/docs/classes/pyfar.coordinates.rst
@@ -2,6 +2,37 @@ Coordinates
 -----------
 
 .. automodule:: pyfar.classes.coordinates
+
+**Classes:**
+
+.. autosummary::
+
+   pyfar.Coordinates
+
+**Functions:**
+
+.. autosummary::
+
+   pyfar.classes.coordinates.cart2cyl
+   pyfar.classes.coordinates.cart2sph
+   pyfar.classes.coordinates.cyl2cart
+   pyfar.deg2rad
+   pyfar.rad2deg
+   pyfar.classes.coordinates.sph2cart
+
+.. autoclass:: pyfar.Coordinates
    :members:
    :undoc-members:
-   :show-inheritance:
+   :inherited-members:
+
+.. autofunction:: pyfar.classes.coordinates.cart2cyl
+
+.. autofunction:: pyfar.classes.coordinates.cart2sph
+
+.. autofunction:: pyfar.classes.coordinates.cyl2cart
+
+.. autofunction:: pyfar.deg2rad
+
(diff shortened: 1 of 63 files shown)


sikersten:
Thanks for looking that up! I think the explicit listing is not a huge issue, since these cases are handled explicitely in __init__.py anyway, right?
Would it be possible to add an extra .rst file for including `.. automodule:: pyfar` and `:exclude-members:`, so this is not done implicitely in pyfar.audio.rst?

docs/classes/pyfar.coordinates.rst
```diff
@@ -2,6 +2,37 @@ Coordinates
...
+   pyfar.Coordinates
+
+**Functions:**
+
+.. autosummary::
+
+   pyfar.classes.coordinates.cart2cyl
+   pyfar.classes.coordinates.cart2sph
+   pyfar.classes.coordinates.cyl2cart
+   pyfar.deg2rad
```
ahms5, line 19: maybe change the order, first rad2deg/deg2rad then the others?

docs/conf.py
```diff
@@ -19,6 +19,7 @@
 
 extensions = [
     'sphinx.ext.autodoc',
+    'sphinx.ext.autosummary',
```
mberz, line 22:
```suggestion
```
See 3 lines below ;)

ahms5 (changes requested):
nice solution, thank you a lot! Maybe we can now change also the paths to the other classes to the sort url: eg. :py:class:`~pyfar.classes.audio.Signal` -> :py:class:`~pyfar.Signal`. I have checked it, the old version is still working, but new new one is better, i think

mberz (changes requested):
Autosummary was already configured in the current Sphinx conf.

Otherwise I really appreciate this change.

docs/conf.py
f-brinkmann: The mention below is an option for autodoc and I thought I needed to include it here, when I use it ecplicitly. But it seems to work without as well so I removed it. Maybe its imported as part of autodoc in this case.

docs/classes/pyfar.coordinates.rst
f-brinkmann: I thought it might be nicer to strictly stay with the alphabetic order. Maybe we can generally discuss if we want to import those as top level functions...

ahms5: also fine for me. should we rename the references now? like mentioned in the comment of the review?

ahms5 (approved):
#701 should fix the error

f-brinkmann:
@ahms5 sorry, just fixed it and saw you pull later. I'm closing it in favor of this 🙈

docs/conf.py
mberz: This is the current conf.py. It's right there. ;)

```Python
extensions = [
    'sphinx.ext.autodoc',
    'sphinx.ext.viewcode',
    'sphinx.ext.napoleon',
    'sphinx.ext.autosummary', # <--- ;)
    'matplotlib.sphinxext.plot_directive',
    'sphinx.ext.mathjax',
    'sphinx.ext.intersphinx',
    'autodocsumm',
    'sphinx_design',
    'sphinx_favicon',
    'sphinx_reredirects',
    'sphinx_mdinclude',
]
```

ahms5 (approved):
exept for ruff

sikersten (approved):
Thanks for the effort! Looks very nice.
I think the contribution guide need to be adjusted at one position referencing ``pyfar.classes.audio.Signal``, but his is part of the Gallery, right?

f-brinkmann:
Yes that should be part of the gallery

Issue #647, opened 2024-07-05
Ruff configuration


mberz:
Currently ruff is set-up to replicate the flake8 guidelines, which could be extended.
I'm creating this issue to keep track of the different configuration options and respective branches.

In progress:
- [ ] docstring rules #612 and #685
- [ ] pytest rules #692 
- [x] "FIX": Code should not contain FIXME, TODO, etc. #687 
- [x] "ERA": No commented out code #686
- [x] "NPY": Check all numpy related deprecations #688 
- [x] "COM": trailing comma rules. Can be very helpful to avoid accidental tuple creation and other problems. - COM819?  #689 
- [x] "A": Avoid builtin function and type shadowing. #690
- [x] "I002": [missing-required-import](https://docs.astral.sh/ruff/rules/missing-required-import/)  #691
- [x] "TID252": Use absolute over relative imports.  #691
- [x] "ARG": Remove unused function/method arguments. #697
- [ ] "B": flake8-bugbear extension. Contains a number of guidelines (seem to be quite opinionated, but at least some look very reasonable) https://docs.astral.sh/ruff/rules/#flake8-bugbear-b  - without B006, B008 #698
- [x] "C4": flake8-comprehensions #699 

Some of them are very easy to apply and already mostly fulfilled in pyfar. For example, builtin shadowing was something we tried to avoid (although not 100% successful as it seems :) )
Some may even be auto-fixed., others require more changes and additional care. 
Feel free to suggest additional rules or rules (or respective subsets) which should not be enforced.

Update: We agreed on all options except for "B" and will discuss options from @ahms5 at a later point but want to have "C90" and "C4" and "PT011" mentioned in #613 (as a separate pull)

ahms5:
note on in progress:
- [ ] docstring rules https://github.com/pyfar/pyfar/pull/612
```
    "D100", # Missing docstring in public module
    "D101", # Missing docstring in public class
    "D102", # Missing docstring in public method
    "D103", # Missing docstring in public function
    "D104", # Missing docstring in public package
    "D105", # Missing docstring in magic method
    "D106", # Missing docstring in public nested class
```
- [ ] also docstring https://github.com/pyfar/pyfar/pull/614
```
    "D206", # Docstring should be indented with spaces, not tabs
    "D207", # Docstring is under-indented
    "D208", # Docstring is over-indented
    "D407", # Missing section underline
    "D408", # Section underline should be in the line following the section's name
    "D409", # Section underline should match the length of its name
[...]

PR #689, opened 2024-09-08
Ruff: COM, trailing comma rules

ahms5:

See #647

### DIFF
--- a/pyfar/classes/coordinates.py
+++ b/pyfar/classes/coordinates.py
@@ -1667,9 +1667,9 @@ class Coordinates():
                 index_multi = np.moveaxis(index, -1, 0)
                 index = np.empty((k), dtype=tuple)
                 for kk in range(k):
-                    index[kk] = tuple([index_multi[kk]], )
+                    index[kk] = tuple([index_multi[kk]] )
             else:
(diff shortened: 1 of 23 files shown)


sikersten (approved):

pyfar/classes/coordinates.py
```diff
@@ -1667,9 +1667,9 @@ def find_nearest(self, find, k=1, distance_measure='euclidean',
                 index_multi = np.moveaxis(index, -1, 0)
                 index = np.empty((k), dtype=tuple)
                 for kk in range(k):
-                    index[kk] = tuple([index_multi[kk]], )
+                    index[kk] = tuple([index_multi[kk]] )
```
hoyer-a, line 1670: Is there a reason to keep the whitespace before `)`?
Would think that violates another codestyle-rule.

ahms5: good point, I removed the space

hoyer-a (changes requested):
Thanks! 

Found two more whitespaces before `]`. Probably also a potential problem?

hoyer-a (dismissed):

mberz (approved):

PR #694, opened 2024-09-11
Remove unwanted warning in Coordinates initialization

f-brinkmann:

Creating a Coordinates object always raises a warning, even in featres scheduled for deprecation are *not* used.

Change the way the warning for the deprecated `sh` property of the class is raised.

### DIFF
--- a/pyfar/classes/coordinates.py
+++ b/pyfar/classes/coordinates.py
@@ -260,12 +260,6 @@ class Coordinates():
         self.sh_order = sh_order
         self._comment = comment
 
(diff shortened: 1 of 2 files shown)


ahms5 (approved):
if the problem with flake8 is fixed, (odd behaviour)

f-brinkmann:
I also wondered about that. Am I allowed to fix it in this pull?

pingelit (approved):
Same as Anne. No clue why flake8 has an issue here.

PR #699, opened 2024-09-15
Ruff: C4, comprehensions

ahms5:
see #647


pyfar/io/_codec.py
```diff
@@ -154,7 +154,7 @@ def _decode(obj, zipfile):
     if isinstance(obj, dict):
         for key in obj.keys():
             _inner_decode(obj, key, zipfile)
-    elif any([isinstance(obj, x) for x in [list, tuple, set, frozenset]]):
+    elif type(obj) in [list, tuple, set, frozenset]:
```
f-brinkmann, line 157: Is using `type` instead of `isinstance` intended? If not `isinstance(obj, (list, tuple, set, frozenset))` should also work.

f-brinkmann (dismissed):
Thanks for the effort. Only had one question where the code could be changed, but otherwise approved.

f-brinkmann (approved):

hoyer-a (approved):

PR #707, opened 2024-09-23
hot: fix compress flag in pyfar.io.write

### LINKED ISSUE #706: io.write(..., compress=True) increases file size
## General

* pyfar version: 0.6.9
* Python version: 3.11.9
* Operating System: Ubuntu 23.10

## Description
Hi guys, 

I noticed that instead of compressing the data, `pf.io.write()` produces much bigger files when using `compress=True`. 

I tested this with a sine signal. The behavior is consistent for long and short signals and single-channel as well as multi-channel audio. 

When using noise, the difference decreases but the compressed file is still larger than the uncompressed one.

## What I Did

```
import pyfar as pf

signal = pf.signals.sine(frequency=1000, n_samples=44100, amplitude=1, sampling_rate=44100)

pf.io.write('signal_uncompressed', signal=signal)
pf.io.write('signal_compressed', signal=signal, compress=True)
```
`signal_uncompressed.far` is 383.4 kB, while `signal_compressed.far` is 706.7 kB.

ahms5:

Closes #706

- fix usage of compress parameter in ``pyfar.io.write``
- add test whether the compressed file is smaller then the uncompressed one.

### DIFF
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -451,6 +451,15 @@ def test_write_read_multiplePyfarObjects(
     assert dict_of_builtins.items() <= actual.items()
 
 
+def test_write_read_compression(sine, tmpdir):
+    filename_compressed = os.path.join(tmpdir, 'sine_compressed.far')
+    io.write(filename_compressed, signal=sine, compress=True)
+    filename_uncompressed = os.path.join(tmpdir, 'sine_uncompressed.far')
+    io.write(filename_uncompressed, signal=sine, compress=False)
+    assert os.path.getsize(filename_uncompressed) > os.path.getsize(
+        filename_compressed)
+
+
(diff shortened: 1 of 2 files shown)


sikersten (approved):
Oh wow, interesting bug! Thanks for fixing!

tests/test_io.py
```diff
@@ -451,6 +451,15 @@ def test_write_read_multiplePyfarObjects(
     assert dict_of_builtins.items() <= actual.items()
 
 
+def test_write_read_compression(sine, tmpdir):
```
ahms5, line 454:
```suggestion
def test_write_read_compression(sine, tmpdir):
    """Test whether compressed files are larger than uncompressed files."""
```

f-brinkmann (approved):
Should we make it part of 0.7.0 if we release this week?

PR #708, opened 2024-09-24
add verbose to read_sofa

ahms5:
verbose was always True since it was not possible to change it when sf.read_sofa is called interanally.
Maybe we can make it in this releae ^^


f-brinkmann (approved):
Thanks, good point to add this to avoid console prints if desired.

sikersten (approved):

Issue #635, opened 2024-06-21
`dsp.regularized_spectrum_inversion` requires `frequency_range` parameter even when `regu_final` is given


sikersten:
## General

* pyfar version: 0.6.7

## Description

The `dsp.regularized_spectrum_inversion` function requires the `frequency_range` parameter even when `regu_final` is given, though it is unused in this case.

## What I Did

```
import pyfar as pf
import numpy as np

signal = pf.signals.impulse(10)
regu_final = np.random.rand(10)
pf.dsp.regularized_spectrum_inversion(signal, regu_final=regu_final)
```
TypeError: regularized_spectrum_inversion() missing 1 required positional argument: 'frequency_range'

f-brinkmann:
How about taking this as an opportunity to start #440. Since this is a little larger and also interesting project, this might be a nice next project for @hoyer-a. We could discuss during the next weekly...

ahms5:
done in #716 , wil be released in v0.8.0

PR #716, opened 2024-10-17
Regularization class

### LINKED ISSUE #440: Restructure `pf.dsp.regularized_spectrum_inversion`
## General

* pyfar version: 0.5.2

## Description

We discussed that function in its current form is a bit too complicated. It has a special case of a regularization function implement through `freq_range`, etc. and a general case through `regu_final`. We decided to refactor the function on the long run:

- move the regularization function generated by `freq_range` et al. to a separate function/class
- always require to pass a regularization function to `pf.dsp.regularized_spectrum_inversion`
- regularization function must be passed as `FrequencyData`object

hoyer-a:
New class for regularization and spectrum inversion to replace `dsp.regularized_spectrum_inversion()`.
Each method of defining a regularization function has it's own class method, so more regularization types can easily be added.

Closes #440 #635 

- move spectrum inversion to `Regularization.invert()` and deprecate dsp function
- add possibility to pass target function for inversion. If provided the convolution of the inverse filter with the system approaches the target function.
- simplified regularization from frequency range by replacing regu_inside and regu_outside by a single regularization factor. To this end regu_inside is set to 0, and the `normalize` parameter is removed. Results using the new parameters are close to identical but the interface is cleaner (see example below).
- regularization from a passed signal, containing the regularization factors as `FrequencyData` (_regu_final_ in dsp-function)

### Previous vs. new implementation

The regularization function is now squared, which was not done previously but seems to be more common. This causes slight differences, with respect to the old method:

```python
import pyfar as pf
%matplotlib inline

sweep = pf.signals.exponential_sweep_freq(2**16, [50, 16e3], 1000, 10e3)

inversion_1 = pf.dsp.regularized_spectrum_inversion(sweep, [50, 16e3])

Inversion = pf.dsp.RegularizedSpectrumInversion.from_frequency_range(
    [50, 16e3])
inversion_2 = Inversion.invert(sweep, normalize_regularization='max')

ax = pf.plot.freq(inversion_1, label='previous')
pf.plot.freq(inversion_2, linestyle=':', label='new')
ax.legend()
```

output

### tbc
- Add more regularization types (e.g. maximum gain, automatic regularization parameter)

### For discussion:
- [x] Rename class to `RegularizedSpectrumInversion`, `from_signal` to `from_magnitude_spectrum`, and `get_regularization` to `regularization`
- [x] Use magnitude and phase of target function
- [x] Make the normalization of the regularization function a parameter
- [x] Check warnings for using `len(Signal)` when using `if signal`
- [ ] Examples Notebook


pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,337 @@
...
+        normalize_regularization : str, optional
+            Normalize the regularization factors. Default is ``None``.
+
+            ``'energy'``
+                Normalize regularization factors to the quadratic maximum
+                of a given signal.
+        """
+        instance = cls(regu_type = "frequency range")
+
+        if len(frequency_range) < 2:
```
f-brinkmann, line 144:
```suggestion
        if len(frequency_range) != 2:
```

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,337 @@
...
+            inverse.freq = np.conj(data) * regu_final / (np.conj(data) * data)
+        else:
+            inverse.freq = np.conj(data) / (np.conj(data) * data + regu_final)
+
+        return inverse
+
+
+    def regularization(self, signal):
+        """
+        Get the frequency dependent regularization factors.
```
f-brinkmann, line 233:
```suggestion
        Get the frequency dependent regularization function :math:`\epsilon(f)`.
```

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,337 @@
...
+                self._frequency_range[1],
+                np.min([self._frequency_range[1]*np.sqrt(2),
+                        signal.sampling_rate/2])])
+
+            # crossfade regularization factors at frequency limits
+            regu_final = _cross_fade(regu_final, regu_outside, idx_xfade_upper)
+
+        if self._target is None:
+            # control amount of regularization using beta
+            regu_final *= self._beta
```
f-brinkmann, line 318: I would suggest to apply beta only during the inversion. It should not be in the returned regularization function.

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,337 @@
...
+            regu = self._get_regularization_from_frequency_range(signal)
+        elif self._regu_type == "magnitude spectrum":
+            regu = self._get_regularization_from_regularization_signal(signal)
+
+        # Normalize regularization factors to energy of signal
+        if self._normalize == 'energy':
+            regu.freq *= np.max(np.abs(signal.freq)**2)
+
+        if self._target is not None:
+            regu = self._get_regularization_from_target(signal, regu)
```
f-brinkmann, line 271: I would move this to self.invert

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,337 @@
...
+
+        # Call private method to get regularization factors
+        if self._regu_type == "frequency range":
+            regu = self._get_regularization_from_frequency_range(signal)
+        elif self._regu_type == "magnitude spectrum":
+            regu = self._get_regularization_from_regularization_signal(signal)
+
+        # Normalize regularization factors to energy of signal
+        if self._normalize == 'energy':
+            regu.freq *= np.max(np.abs(signal.freq)**2)
```
f-brinkmann, line 268: Two thoughts:

- The way this was used before, max(regu.freq)=1 was assumed. We now need to scale regu.freq by f(signal) / f(regu), where f is some funtion. 
- I would suggest to call the normalization 'maximum_energy' and introduce a second normalization termed 'energy' which uses f = np.abs()**2 without taking the maximum.

pyfar/dsp/regularization.py
f-brinkmann: It should also be mentioned that the normalization is applied.

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,337 @@
...
+    signal's spectrum or by the limit of a system's bandwidth.
+
+    The frequency dependent regularization :math:`\epsilon(f)` can be defined
+    using one of the class methods.
+    The regularization is then applied when calculating the inverse of a
+    signal as [1]_, [2]_:
+
+    .. math::
+
+        S^{-1}(f) = \frac{S^*(f)}{S^*(f)S(f) + \beta * \epsilon(f)},
```
f-brinkmann, line 24: I think we should write AND apply epsilon**2 as noted in Norcross Eq. 2.6. Not sure why we did not square in the earlier implementation - maybe because of the way beta was applied.

f-brinkmann (changes requested):
Thanks for the implementation. In the first round, I commented on some structural changes that I would suggest. I'll do a more detailed review including tests in the next round.

ahms5:
looks like an example in the doc is failing

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,313 @@
...
+        """
+        instance = cls(regu_type = "frequency range")
+
+        if len(frequency_range) != 2:
+            raise ValueError(
+                "The frequency range needs to specify lower and upper limits.")
+
+        instance._frequency_range = np.asarray(frequency_range)
+
+        return instance
```
f-brinkmann, line 112: I dont think we need a return value here. Its changed in place, i.e., inside the class object anyhow

hoyer-a: Mh, we're also instancing the class-object in `from_frequency_range()`, without the return we don't get any object to work with.

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,313 @@
...
+                :math:`Normalization Factor = \frac{\max(|S(f)|)}{\max(|\epsilon(f)|)}`
+            ``'mean'``
+                Normalize the regularization :math:`\epsilon` to the mean
+                magnitude of a given signal :math:`S(f)`.
+
+                :math:`Normalization Factor = \frac{\text{mean}(|S(f)|)}{\text{mean}(|\epsilon(f)|)}`
+
+        Returns
+        -------
+        normalization_factor : float
```
ahms5, line 255:
```suggestion
        beta : float
```

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,313 @@
...
+    signal as [1]_:
+
+    .. math::
+
+        S^{-1}(f) = \frac{S^*(f)}{S^*(f)S(f) + \beta * |\epsilon(f)|^2} * D(f)
+
+    with :math:`S(f)` being the input signal's Fourier transform,
+    :math:`S^*(f)` the complex conjugate, and :math:`\beta` a parameter to
+    control the influence of the regularization function on the inversion.
+    :math:`D(f)` denotes an optional arbitrary target function.
```
ahms5, line 29: I would skip applying the target function here. it is making it more complex then it is, and the Class is called Inversion.

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,313 @@
...
+        if not isinstance(regularization, pf.Signal):
+            raise ValueError(
+                "Regularization must be a pyfar.Signal object.")
+
+        instance = cls(regu_type = "magnitude spectrum")
+        instance._regu = regularization
+
+        return instance
+
+    def invert(self, signal, beta=1, target=None,
```
ahms5, line 135: invert should invert the signal and not multily it by a target

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,313 @@
...
+
+        Parameters
+        ----------
+        signal : pf.Signal
+            Signal to be inverted.
+        beta : float, optional
+            Beta parameter to control the amount of regularization.
+        target : pf.Signal, optional
+            Target function for the regularization.
+        normalize_regularization : str
```
ahms5, line 148:
```suggestion
        normalize_regularization : str, optinal
```

and we need to add, by defauld no normalization

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,313 @@
...
+        regularization.
+
+        Parameters
+        ----------
+        signal : pf.Signal
+            Signal to be inverted.
+        beta : float, optional
+            Beta parameter to control the amount of regularization.
+        target : pf.Signal, optional
+            Target function for the regularization.
```
ahms5, line 147:
```suggestion
```
see above

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,313 @@
...
+        # Call private method to get regularization factors
+        if self._regu_type == "frequency range":
+            regu = self._get_regularization_from_frequency_range(signal)
+        elif self._regu_type == "magnitude spectrum":
+            regu = self._get_regularization_from_regularization_signal(signal)
+
+        return regu
+
+
+    def normalization_factor(self, signal, normalize_regularization):
```
ahms5, line 230: maybe this method should be private

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,313 @@
...
+                             np.abs(regu.freq)**2)
+
+        # Apply target function
+        if target is not None:
+            inverse.freq *= target.freq
+
+        return inverse
+
+
+    def regularization(self, signal):
```
ahms5, line 203: maybe this should be private

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,313 @@
...
+        if regu_type == "":
+            raise RuntimeError("Regularization objects must be created "
+                             "using a classmethod.")
+
+    def __repr__(self):
+        return f"Regularization object with regularization "\
+            f"of type '{self._regu_type}'."
+
+    @classmethod
+    def from_frequency_range(cls, frequency_range):
```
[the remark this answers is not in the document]
ahms5: and how about the old parameters ``regu_outside``and ``regu_inside``?

ahms5 (changes requested):

pyfar/dsp/regularization.py
f-brinkmann: I like the verbose parameter name. But how about mentioning the 'beta' in the docstring as math symbol?

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,362 @@
...
+    """ # noqa: E501
+
+    def __init__(self, regu_type: str="") -> None:
+
+        self._regu_type = regu_type
+
+        # throw error if object is instanced without classmethod
+        if regu_type == "":
+            raise RuntimeError("Regularization objects must be created "
+                             "using a classmethod.")
```
f-brinkmann, line 126: I think the indent is wrong and there should be another set of () around the string.

pyfar/dsp/regularization.py
f-brinkmann: ah, sorry - make sense!

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,362 @@
...
+            n_samples = max(signal.n_samples, target.n_samples)
+            signal = pf.dsp.pad_zeros(signal, (n_samples-signal.n_samples))
+            target = pf.dsp.pad_zeros(target, (n_samples-target.n_samples))
+
+        # get regularization factors
+        regu = self.regularization(signal)
+
+        # Apply optional normalization
+        if normalize_regularization is not None:
+            normalization_factor_ = \
```
f-brinkmann, line 234: why the '_' at the end?

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,362 @@
...
+                self.normalization_factor(signal, normalize_regularization)
+            regu.freq *= normalization_factor_
+
+        data = signal.freq
+
+        # calculate inverse filter
+        inverse = signal.copy()
+        inverse.freq = \
+            np.conj(data) / (np.conj(data) * data + beta *
```
f-brinkmann, line 243: I think this is a little faster...
```suggestion
            np.conj(data) / (np.abs(data)**2 + beta *
```

f-brinkmann (changes requested):
Thanks, looks quite good to me. Maybe we can discuss the final names during the next weekly...

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,313 @@
...
+        Returns
+        -------
+        inverse : pf.Signal
+            Resulting signal after regularized inversion.
+        """ # noqa: E501
+        if target is not None and not isinstance(target, pf.Signal):
+            raise ValueError(
+                "Target function must be a pyfar.Signal object.")
+
+        # zero pad, if target function is given and has different n_samples
```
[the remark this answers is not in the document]
hoyer-a: It's to make sure that target function and signal to be inverted are broadcastbale, which are only used in `invert()`

pyfar/dsp/regularization.py
hoyer-a: Indent was wrong, but according to the python docs the RuntimeError("...") is right.

pyfar/dsp/regularization.py
hoyer-a: Do we want to treat 'beta' as an equivalent for the normalization factor. Technically in this implementation we can still adjust beta after applying normalization.

pyfar/dsp/regularization.py
hoyer-a: Good question, i removed it.

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,362 @@
+"""
+Documentation for regularization class.
+"""
+
+import pyfar as pf
+from pyfar.dsp.dsp import _cross_fade
+import numpy as np
+
+class RegularizedSpectrumInversion():
```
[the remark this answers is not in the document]
f-brinkmann: We decided to leave the name as is, because it can also be used for general inversion as well and might not be found if the name is too specific.

sikersten (commented):
Thanks for this great implementation! I have some questions on the documentation.

- I see that the regu_type parameter catches if an object is instanced without classmethod. Is there a different solution where it does not appear as parameter in the docs? E.g., only letting it exist, if the object is instanced correctly?
- I'm confused by the relation of the following quantities: beta, epsilson, `normalization_factor`, `magnitude_spectrum(regularization)`. What's the idea behind normalizing epsilon first and then multiply it with beta again?
- I guess it could help to provide meaningful ranges for these parameters. Maybe even that these ranges as contraints?

f-brinkmann:
> What's the idea behind normalizing epsilon first and then multiply it with beta again?

@sikersten:  the first normalization is convenient in many cases, because it forces epsilon to be roughly in the same magnitude range as the function to be inverted. Using this the effect of beta will be independent from the overall level of the signal to be inverted. We could make that more clear in the docs

> I guess it could help to provide meaningful ranges for these parameters. Maybe even set these ranges as contraints?

We could write something like: If applying normalization (to the regularization) beta is typically between 0 and 1. However, I would not restrict it to this range and leave this up the the user.

@hoyer-a if it is fine to you, I could attemp to update the docs accordingly...

f-brinkmann:
We decided that

- we want to have an empty init. Otherwise irrelevant information is shown in the docs
- `regu_type` should be renamed to `regularization_type`

sikersten:
What's the reason beta is not a property of the class, but a parameter for the invert function?

f-brinkmann:
> What's the reason beta is not a property of the class, but a parameter for the invert function?

That paremeter is often iteratively adjusted. It would thus be a little tedious to have it as a property because each iteration would then require to set beta first, and to invert and insepct the result in a second step.

mberz:
After some additional brainstorming we had the following ideas:
- `beta` and `normalize_regularization` are somewhat redundant and should be merged into `beta` which can be numeric or a string, i.e. `'max'`, `'mean'`.
- changing the function variables of invert to class properties with respective setter and getter functions. Those are
  - `signal`
  - `target`
  - `beta`
  - `regularization` (getter only)
- setter need to implement validity checks depending on the used `from_(...)` classmethod. 
- `from_(...)` class methods should then also handle setting the new class properties

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,386 @@
...
+        signal = self.signal
+
+        # zero pad, if target function is given and has different n_samples
+        if target is not None and \
+            target.n_samples != signal.n_samples:
+            n_samples = max(signal.n_samples, target.n_samples)
+            signal = pf.dsp.pad_zeros(signal,
+                                      (n_samples-signal.n_samples))
+            target = pf.dsp.pad_zeros(target,
+                                      (n_samples-target.n_samples))
```
f-brinkmann, line 336: This might have been my idea - but I think we should undo this. If the target or signal is zero phase, padding zeros at the end would be a bad idea. I think it would be more save to force the user to take care of passing everything in the correct length.

f-brinkmann: In this case using `target = self.target` and `signal = self.signal` might not be needed anymore.

f-brinkmann (changes requested):
Thanks, this looks really good now. I had a few minor and a few easy to change structural comments, that we could discuss in the weekly before you implement them.

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,386 @@
...
+            >>>
+            >>> # Regulated inversion
+            >>> Inversion = pf.dsp.RegularizedSpectrumInversion.from_magnitude_spectrum(
+            ...     hptf, regularization, beta=.1, target=target)
+            >>> inverted = Inversion.invert()
+            >>>
+            >>> # Plot the result
+            >>> ax = pf.plot.time_freq(hptf, label='HpTF')
+            >>> pf.plot.time_freq(inverted, label='HpTF inverted')
+            >>> pf.plot.time_freq(hptf * inverted, label='Equalized HpTF')
```
ahms5, line 319:
```suggestion
            >>>
            >>> # Plot the result
            >>> ax = pf.plot.time_freq(hptf, label='HpTF')
            >>> pf.plot.time_freq(inverted, label='HpTF inverted')
            >>> pf.plot.time_freq(hptf * Inversion, label='Equalized HpTF')
```

so that the inversion class returns the invert for multiplications
would it be cool to do sth like this?

ahms5 (commented):
I have a nice to have idea, what do you think?

pyfar/dsp/regularization.py
f-brinkmann: It would be more convenient but less implicit. I would prever the current way. However, `invert` is not a function anymore, but a property. So the call simplifies to `Inversion.invert`.

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,386 @@
...
+                           "using a classmethod.")
+
+    def __repr__(self):
+        return f"Regularization object with regularization "\
+            f"of type '{self._regularization_type}'."
+
+    @classmethod
+    def from_frequency_range(cls, signal, frequency_range, beta=1,
+                             target=None):
+        r"""
```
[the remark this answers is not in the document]
hoyer-a: We propose an extra parameter to set the regularization inside the frequency range, which is now always `0`.

pyfar/dsp/regularization.py
```diff
@@ -5,333 +5,373 @@
...
-        >>> pf.plot.time_freq(hptf * inverted, label='Equalized HpTF')
-        >>> pf.plot.time_freq(target, linestyle='--', label='Target')
-        >>> pf.plot.freq(regularization, ax=ax[1], label='Regularization')
-        >>> ax[0].set_xlim(0 , .005)
-        >>> ax[1].set_ylim(-40 , 15)
-        >>> ax[1].legend(loc='lower center', ncols=3)
+    The inversion is done in two steps:
+
+    1. Define :math:`S(f)`, :math:`\epsilon(f)`, :math:`\beta`, and
+       :math:`D(f)` using one of the ``from_()`` methods listed below.
```
sikersten, line 39:
```suggestion
       :math:`D(f)` using one of the ``from_...()`` methods listed below.
```

pyfar/dsp/regularization.py
```diff
@@ -5,333 +5,373 @@
...
+    2. Compute the inverse :math:`S(f)^{-1}` using :py:func:`~invert`
+
+    The parameters that are defined in the first step are often iteratively
+    adjusted. Examples are given in the documentation of the `from_()`
```
sikersten, line 43:
```suggestion
    adjusted. Examples are given in the documentation of the `from_...()`
```

pyfar/dsp/regularization.py
```diff
@@ -5,333 +5,373 @@
...
 
-        Returns
-        -------
-        inverse : Signal
-            The inverted input signal.
-        """ # noqa: E501
+    @property
+    def beta(self):
+        r"""
+        Get or set the :math:`\beta` to control the amount off regularization.
```
sikersten, line 273:
```suggestion
        Get or set the :math:`\beta` to control the amount of regularization.
```

pyfar/dsp/regularization.py
```diff
@@ -5,333 +5,373 @@
...
+            ``'max'``
+                Normalize the regularization :math:`\epsilon(f)` to the maximum
+                magnitude of a given signal :math:`S(f)`.
+
+                :math:`Normalization Factor = \frac{\max(|S(f)|)}{\max(|\epsilon(f)|)}`
+            ``'mean'``
+                Normalize the regularization :math:`\epsilon(f)` to the mean
+                magnitude of a given signal :math:`S(f).`
+
+                :math:`Normalization Factor = \frac{\text{mean}(|S(f)|)}{\text{mean}(|\epsilon(f)|)}`
```
sikersten, line 105: This is beta, isn't it?
```suggestion
            ``numerical value``
                Usually between ``0`` and ``1``, with ``0`` being no
                regularization applied.
            ``'energy'``
                Normalize the regularization to match the signal's energy.

                :math:`\beta= \frac{\frac{1}{N}\sum_{k=0}^{N-1}|S[k]|^2}{\frac{1}{N}\sum_{k=0}^{N-1}|\epsilon[k]|^2}`
            ``'max'``
                Normalize the regularization :math:`\epsilon(f)` to the maximum
                magnitude of a given signal :math:`S(f)`.

                :math:`\beta = \frac{\max(|S(f)|)}{\max(|\epsilon(f)|)}`
            ``'mean'``
                Normalize the regularization :math:`\epsilon(f)` to the mean
                magnitude of a given signal :math:`S(f).`

                :math:`\beta = \frac{\text{mean}(|S(f)|)}{\text{mean}(|\epsilon(f)|)}`
```

pyfar/dsp/regularization.py
```diff
@@ -5,333 +5,373 @@
...
 
-        if len(frequency_range) != 2:
+            The default is ``1``.
```
sikersten, line 107: Just a note: In the current implementation at https://pyfar.readthedocs.io/en/stable/modules/pyfar.dsp.html#pyfar.dsp.regularized_spectrum_inversion, I think beta='max'. We should be aware of changing this behavior here

pyfar/dsp/regularization.py
```diff
@@ -5,333 +5,373 @@
...
-        instance = cls(regu_type = "magnitude spectrum")
-        instance._regu = regularization
-
-        return instance
-
-    def invert(self, signal, beta=1, target=None,
-               normalize_regularization=None):
-        r"""
-        Invert the input using frequency-dependent regularized inversion.
+        `regularization` must match the length of the signal to be inverted.
```
sikersten, line 178: Is the length the only property that needs to match? Probably also sampling rate?

pyfar/dsp/regularization.py
```diff
@@ -5,333 +5,373 @@
...
+        high frequencies, and use a band-pass as target function. Note that the
+        equalized HpTF, which is obtained from a convolution of the HpTF with its
+        inverse filter, approaches the target function in both, time and frequency.
+
+        .. plot::
+
+            >>> import pyfar as pf
+            >>> import numpy as np
+            >>>
+            >>> # Hedphone transfer function to be inverted
```
sikersten, line 228:
```suggestion
            >>> # Headphone transfer function to be inverted
```

pyfar/dsp/regularization.py
```diff
@@ -5,333 +5,373 @@
...
+            ...
+            >>> sweep = pf.signals.exponential_sweep_freq(
+            ...     2**16, [50, 16e3], 1000, 10e3)
+            ...
+            >>> # inversion
+            >>> Inversion = pf.dsp.RegularizedSpectrumInversion.from_frequency_range(
+            >>>     sweep, [50, 16e3], beta='max')
+            >>> inverted = Inversion.invert
+            ...
+            >>> # obtain the scaled regularization function
```
sikersten, line 133:
```suggestion
            >>> # Obtain the scaled regularization function
```

sikersten (commented):
Just went through the documentation, very nice! Only found some typos

pyfar/dsp/regularization.py
hoyer-a: Ah yes, good point.

pyfar/dsp/regularization.py
hoyer-a: Maybe we can discuss that briefly tomorrow.

pyfar/dsp/regularization.py
```diff
@@ -36,11 +36,11 @@ class RegularizedSpectrumInversion():
     The inversion is done in two steps:
 
     1. Define :math:`S(f)`, :math:`\epsilon(f)`, :math:`\beta`, and
-       :math:`D(f)` using one of the ``from_()`` methods listed below.
+       :math:`D(f)` using one of the ``from_...()`` methods listed below.
     2. Compute the inverse :math:`S(f)^{-1}` using :py:func:`~invert`
 
     The parameters that are defined in the first step are often iteratively
-    adjusted. Examples are given in the documentation of the `from_()`
+    adjusted. Examples are given in the documentation of the `from_()...`
```
sikersten, line 43:
```suggestion
    adjusted. Examples are given in the documentation of the `from_...()`
```

pyfar/dsp/regularization.py
```diff
@@ -270,7 +275,7 @@ def from_magnitude_spectrum(cls, signal, regularization, beta=1,
     @property
     def beta(self):
         r"""
-        Get or set the :math:`\beta` to control the amount off regularization.
+        Get or set the :math:`\beta` to control the amount of regularization.
```
sikersten, line 278:
```suggestion
        Amount of regularization :math:`\beta`.
```

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,409 @@
...
+        elif beta == 'energy':
+            self._beta = \
+                pf.dsp.energy(self.signal) / pf.dsp.energy(self.regularization)
+        else:
+            self._beta = beta
+
+    @property
+    def regularization(self):
+        r"""
+        Get the regularization :math:`\epsilon(f)` without scaling by
```
sikersten, line 306:
```suggestion
        Regularization :math:`\epsilon(f)` without scaling by
```

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,409 @@
...
+                self._get_regularization_from_frequency_range(self.signal)
+        elif self._regularization_type == "magnitude spectrum":
+            self._regularization =  \
+                self._get_regularization_from_regularization_signal(
+                    self._regularization_signal)
+        return self._regularization
+
+    @property
+    def signal(self):
+        r"""Get or set the signal :math:`S(f)` to be inverted."""
```
sikersten, line 321:
```suggestion
        r"""Signal :math:`S(f)` to be inverted."""
```

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,409 @@
...
+        """Set signal to be inverted."""
+        if not isinstance(signal, pf.Signal):
+            raise TypeError("Regularization of type "
+                            f"'{self._regularization_type}' requires an input"
+                            " signal of type pyfar.Signal.")
+        self._signal = signal
+
+    @property
+    def target(self):
+        """Get or set the target function :math:`D(f)`."""
```
sikersten, line 335:
```suggestion
        """Target function :math:`D(f)`."""
```

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,409 @@
...
+        frequency range.
+
+        Parameters
+        ----------
+        signal : Signal
+            Signal to be inverted.
+        frequency_range : array like
+            Array like containing the lower and upper frequency limit in Hz.
+        regularization_within: float, optional
+            Set regularization inside frequency range. The default is ``0``.
```
sikersten, line 86:
```suggestion
            Regularization inside frequency range. The default is ``0``.
```

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,409 @@
...
+                Normalize the regularization :math:`\epsilon(f)` to the mean
+                magnitude of a given signal :math:`S(f).`
+
+                :math:`\beta = \frac{\text{mean}(|S(f)|)}{\text{mean}(|\epsilon(f)|)}`
+
+            The default is ``1``.
+        target : Signal, optional
+            Target function for the regularization. The default ``None`` uses a
+            zero-phase spectrum with an amplitude of 1 as target equalling an
+            impulse in the time domain.
```
sikersten, line 113:
```suggestion
            zero-phase spectrum with an amplitude of 1 as target, equal to an
            impulse in the time domain.
```

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,409 @@
...
+            f"of type '{self._regularization_type}'."
+
+    @classmethod
+    def from_frequency_range(cls, signal, frequency_range,
+                             regularization_within=0, beta=1, target=None):
+        r"""
+        Regularization from a given frequency range.
+
+        Defines a frequency range within which the regularization
+        :math:`\epsilon(f)` is set to ``regularization_within``.
```
sikersten, line 72:
```suggestion
        :math:`\epsilon(f)` is set to `regularization_within`.
```

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,409 @@
...
+            Signal to be inverted.
+        frequency_range : array like
+            Array like containing the lower and upper frequency limit in Hz.
+        regularization_within: float, optional
+            Set regularization inside frequency range. The default is ``0``.
+        beta : float, string, optional
+            Beta parameter to control the scaling of the regularization as in
+            :eq:`regularized_inversion`. Can be a
+
+            ``numerical value``
```
sikersten, line 91:
```suggestion
            numerical value
```

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,409 @@
...
+        signal : Signal
+            Signal to be inverted.
+        regularization : Signal
+            Signal defining the regularization :math:`\epsilon(f)` in
+            ``regularization.freq``.
+        beta : float, string, optional
+            Beta parameter to control the scaling of the regularization as in
+            :eq:`regularized_inversion`. Can be a
+
+            ``numerical value``
```
sikersten, line 193:
```suggestion
            numerical value
```

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,409 @@
...
+            ``'mean'``
+                Normalize the regularization :math:`\epsilon(f)` to the mean
+                magnitude of a given signal :math:`S(f).`
+
+                :math:`\beta = \frac{\text{mean}(|S(f)|)}{\text{mean}(|\epsilon(f)|)}`
+
+            The default is ``1``.
+        target : Signal, optional
+            Target function for the regularization. The default ``None`` uses a
+            zero-phase spectrum with an amplitude of 1 as target equalling an
```
sikersten, line 214:
```suggestion
            zero-phase spectrum with an amplitude of 1 as target, equal to an
```

sikersten (commented):
Some minor comments on the docstrings. I leave it up to you if you want to follow these suggestions :)

pyfar/dsp/regularization.py
```diff
@@ -40,7 +40,7 @@ class RegularizedSpectrumInversion():
     2. Compute the inverse :math:`S(f)^{-1}` using :py:func:`~invert`
 
     The parameters that are defined in the first step are often iteratively
-    adjusted. Examples are given in the documentation of the `from_()...`
+    adjusted. Examples are given in the documentation of the `from_...()`
```
sikersten, line 43:
```suggestion
    adjusted. Examples are given in the documentation of the ``from_...()``
```

sikersten (dismissed):

sikersten (approved):

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,408 @@
...
+
+    The inverse is computed as [#]_:
+
+    .. math::
+        :label: regularized_inversion
+
+        S^{-1}(f) = \frac{S^*(f)}{S^*(f)S(f) + \beta |\epsilon(f)|^2} D(f)
+
+    with :math:`f` being the frequency, :math:`S(f)` the spectrum of the signal
+    to be inverted, :math:`(\cdot)^*` the complex conjugate,
```
ahms5, line 25:
```suggestion
    to be inverted, :math:`*` the complex conjugate,
```
looks somehow odd in my opinion

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,408 @@
...
+    The inversion is done in two steps:
+
+    1. Define :math:`S(f)`, :math:`\epsilon(f)`, :math:`\beta`, and
+       :math:`D(f)` using one of the ``from_...()`` methods listed below.
+    2. Compute the inverse :math:`S(f)^{-1}` using :py:func:`~invert`
+
+    The parameters that are defined in the first step are often iteratively
+    adjusted. Examples are given in the documentation of the ``from_...()``
+    methods.
+
```
ahms5, line 45:
```suggestion
```
isnt it nessesary to mention? It is already clear from above and we dont need to teaser the examples here.

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,408 @@
...
+            f"of type '{self._regularization_type}'."
+
+    @classmethod
+    def from_frequency_range(cls, signal, frequency_range,
+                             regularization_within=0, beta=1, target=None):
+        r"""
+        Regularization from a given frequency range.
+
+        Defines a frequency range within which the regularization
+        :math:`\epsilon(f)` is set to `regularization_within`.
```
ahms5, line 72:
```suggestion
        :math:`\epsilon(f)` is set to ``regularization_within``.
```

our guidelines are suggesting double ``

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,408 @@
...
+        Outside the frequency range the regularization is
+        :math:`\epsilon(f)=1` and can be controlled using the `beta` parameter.
```
ahms5, line 74:
```suggestion
        :math:`\epsilon(f)=1` and can be controlled using the ``beta`` parameter.
```

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,408 @@
...
+        ----------
+        signal : Signal
+            Signal to be inverted.
+        frequency_range : array like
+            Array like containing the lower and upper frequency limit in Hz.
+        regularization_within: float, optional
+            Set regularization inside frequency range. The default is `0`.
+        beta : float, string, optional
+            Beta parameter to control the scaling of the regularization as in
+            :eq:`regularized_inversion`. Can be a
```
ahms5, line 89:
```suggestion
            Equation :eq:`regularized_inversion`. Can be a
```

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,408 @@
...
+        return instance
+
+    @classmethod
+    def from_magnitude_spectrum(cls, signal, regularization, beta=1,
+                                target=None):
+        r"""
+        Regularization from a given magnitude spectrum.
+
+        Regularization passed as :py:class:`~pyfar.Signal`. The length and
+        sampling rate of `regularization` must match the length and sampling
```
ahms5, line 179:
```suggestion
        sampling rate of ``regularization`` must match the length and sampling
```

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,408 @@
...
+    @property
+    def beta(self):
+        r"""
+        Amount of regularization :math:`\beta`.
+        """
+        return self._beta
+
+    @beta.setter
+    def beta(self, beta):
+        """Set beta parameter for sacling of regularization function."""
```
ahms5, line 284:
```suggestion
        """Set beta parameter for scaling of regularization function."""
```

pyfar/dsp/regularization.py
```diff
@@ -0,0 +1,408 @@
...
+
+    @property
+    def beta(self):
+        r"""
+        Amount of regularization :math:`\beta`.
+        """
+        return self._beta
+
+    @beta.setter
+    def beta(self, beta):
```
ahms5, line 283: here it smells like a similar behviour as the Filter.state. Do we really want to have a setter were we allow numbers as well as strings defining a certain beahviour? maybe it would be more elegant to allow a number to set and to define an other method to set beta for energy', 'mean', 'max'.

ahms5 (changes requested):
thanks for taking care! Im wondering about the beta parameter (also in the from_* construcors) do we really want to allow numbers and strings? It smells like Filter.state, which might become confusing. Maybe we can add a method which calcualtes the beta based on presettings.

f-brinkmann:
@ahms5 (@hoyer-a) I think the question regarding possible values (numeric & string) for `beta` is a little different than the issue with the filter state. Beta is used consistently throughout the class and the parameter is always called `beta`. In the filter class there were different parameters/methods for the state.

We tried to already think about future use cases (more `from_`methods). For them, beta will either be the same, or they don't have it because they are parametrized in a different way.

Having a separate method for `beta` might produce cleaner code, but I'm not sure if that would be a cleaner concept. This is what I could think of:

``` python
signal = some_signal
# this sets beta to the default value of 1
inverter = pf.dsp.RegularizedSpectrumInversion.from_frequency_range(signal, [50, 20e3])
# this sets beta to the desired value. The step above is required, we need the signal
# and the regularization function first. We would have multiple methods for setting beta.
inverter.set_beta('max')
inverted = inverter.invert
```

This is how it currently looks

``` python
signal = some_signal
inverter = pf.dsp.RegularizedSpectrumInversion.from_frequency_range(signal, [50, 20e3], beta='max')
inverted = inverter.invert
```

Should we discuss this in the weekly?

hoyer-a:
Since beta is a parameter that needs to be adjusted when using the class, i think it would be better to just have the setter and no extra function. In this case, allowing to quickly switch between different scalings or values seems to be more important than clean separation. We allow mixed input like this in a few other cases, but i'm fine with both

hoyer-a:
Since beta is a parameter that needs to be adjusted when using the class, i think it would be better to just have the setter and no extra function. In this case, allowing to quickly switch between different scalings or values seems to be more important than clean separation. We allow mixed input like this in a few other cases, but i'm fine with both

ahms5:
the only thing I'm worried about is the dual usecase in the parameter ``beta``, IDEs hate this trick, maybe we could do sth like this:

```
signal = some_signal
# here 'None', 'max', 'energy', 'mean' are allowed ('None' would be 1), 
# here beta can also have strings depending on the from_ method, if required
inverter = pf.dsp.RegularizedSpectrumInversion.from_frequency_range(signal, [50, 20e3], beta='max') 
# if the users want to set a custom numeric value, they can use the property, which just allows numeric values between 0 and 1. 
inverter.beta = 0.5
inverted = inverter.invert
```
we could also think about changing the name beta in the from methods to sth like ``beta_normalization`` or ``beta_method``, ...

i think most of the time the user would use 1 or one of the string settings.
in this case all parameters would just be used by one type, either string or numbers, and everthing is well separated.

f-brinkmann:
We decided the following
- `beta` always return what it gets, i.e., a number or a string
- a new property `beta_value` will always return the number

pyfar/dsp/regularization.py
```diff
@@ -69,9 +69,10 @@ def from_frequency_range(cls, signal, frequency_range,
         Regularization from a given frequency range.
 
         Defines a frequency range within which the regularization
-        :math:`\epsilon(f)` is set to `regularization_within`.
+        :math:`\epsilon(f)` is set to ``regularization_within``.
         Outside the frequency range the regularization is
-        :math:`\epsilon(f)=1` and can be controlled using the `beta` parameter.
+        :math:`\epsilon(f)=1` and can be controlled using the ``beta``
```
f-brinkmann, line 74: see comment above

pyfar/dsp/regularization.py
```diff
@@ -176,7 +177,7 @@ def from_magnitude_spectrum(cls, signal, regularization, beta=1,
         Regularization from a given magnitude spectrum.
 
         Regularization passed as :py:class:`~pyfar.Signal`. The length and
-        sampling rate of `regularization` must match the length and sampling
+        sampling rate of ``regularization`` must match the length and sampling
```
f-brinkmann, line 180: see comment above

pyfar/dsp/regularization.py
```diff
@@ -69,9 +69,10 @@ def from_frequency_range(cls, signal, frequency_range,
         Regularization from a given frequency range.
 
         Defines a frequency range within which the regularization
-        :math:`\epsilon(f)` is set to `regularization_within`.
+        :math:`\epsilon(f)` is set to ``regularization_within``.
```
f-brinkmann, line 72: According to our guidelines it should indeed be in single quotes. We use italic font to reference parameters and monospaced font (double quotes) to indicate actual code. Can you change this here and below?

f-brinkmann (dismissed):
Only had comments on the docstring left and saw that the test raise some warnings. Can you try to get rid of these? Otherwise approved

ahms5 (dismissed):

sikersten (dismissed):

ahms5 (approved):

pyfar/dsp/regularization.py
```diff
@@ -22,7 +23,7 @@ class RegularizedSpectrumInversion():
         S^{-1}(f) = \frac{S^*(f)}{S^*(f)S(f) + \beta |\epsilon(f)|^2} D(f)
 
     with :math:`f` being the frequency, :math:`S(f)` the spectrum of the signal
-    to be inverted, :math:`(\cdot)^*` the complex conjugate,
+    to be inverted, :math:`*` the complex conjugate,
```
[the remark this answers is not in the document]
hoyer-a: Jep

ahms5 (approved):

f-brinkmann (approved):

PR #722, opened 2024-11-15
Normaliztation: Force energetic properties to match the target after normalization

### LINKED ISSUE #721: dsp.normalize's reference_methods "power" and "energy" behave unexpectedly
## General

* pyfar version: 0.6.8 (issue persists in current code though)
* Python version: 3.12.4
* Operating System: windows

## Description

I tried to normalize several signals of different length to the same total energy using pyfar. I expected this to be possible using 
```
pyfar.dsp.normalize(signal, reference_method="energy", target=1)
```
such that each signal would end up with a total energy of 1. However, since the input is a pressure signal and the reference value is energetic, the output signals do not have energies of 1. In fact, they do not even have the same energies:

For instance, with an input signal with an energy of 20, normalize will output a signal with an energy of 0.05 (1/20), whereas an input energy of 5 will output an energy of 0.2 (1/5).
I talked about this with @f-brinkmann and we both think the function call above should return a signal with an energy of 1.
So, when the `reference_method` is set to "power" or "energy", we believe the signal should be divided by the root of the calculated reference value, not the value itself.

## Code Example
```py
shorter = pf.signals.noise(2000, rms=0.1)
longer = pf.signals.noise(4000, rms=0.1)

for input_signal, name in [(shorter, "shorter"), (longer, "longer")]:
    input_energy = pf.dsp.energy(input_signal)
    output_signal = pf.dsp.normalize(input_signal, reference_method="energy")
    output_energy = pf.dsp.energy(output_signal)
    
    print(f"{name}:")
    print(f"  in:  {input_energy}")
    print(f"  out: {output_energy}")
```

Outputs:
```
shorter:
  in:  [20.]
  out: [0.05]
longer:
  in:  [40.]
  out: [0.025]
```

f-brinkmann:

Closes #721 

- use square root of a signals energy and power for normalization
- use square root of target as well.
- update docs and tests

### DIFF
--- a/pyfar/dsp/dsp.py
+++ b/pyfar/dsp/dsp.py
@@ -2442,8 +2442,14 @@ def normalize(signal, reference_method='max', domain='auto',
             Compute the mean absolute values per channel.
         ``'energy'``
             Compute the energy per channel using :py:func:`~pyfar.dsp.energy`.
+            Note that the square root of the energy is used as `reference`,
+            which is required for the energy of the normalized signal to match
+            the `target`.
         ``'power'``
             Compute the power per channel using :py:func:`~pyfar.dsp.power`.
+            Note that the square root of the power is used as `reference`,
+            which is required for the power of the normalized signal to match
+            the `target`.
         ``'rms'``
             Compute the RMS per channel using :py:func:`~pyfar.dsp.rms`.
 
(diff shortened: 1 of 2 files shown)


hoyer-a (approved):

sikersten (commented):
Thanks for taking care! I'm wondering if it's clear that for `channel_handling='mean'` the mean energy is calculated before taking the square root, and not the other way around. What to you think?

f-brinkmann:
> I'm wondering if it's clear that for channel_handling='mean' the mean energy is calculated before taking the square root, and not the other way around. What to you think?

Thanks for catching. It was correct, but probably not clear. In addition, I detected thet the normalization for `'energy'` and `'power'` was only correct for `target=1`. I fixed that as well and extended the tests.

PR #728, opened 2024-11-29
Enhance head-related impulse responses

f-brinkmann:

Technically not a bug, because it was mentioned in the docs - but unessesarily complicated.

- Make `pyfar.signals.files.head_related_impulse_responses` work with array likes and pyfar Coordinate objects

### DIFF
--- a/pyfar/signals/files.py
+++ b/pyfar/signals/files.py
@@ -354,6 +354,8 @@ def head_related_impulse_responses(
             (0 degree elevation) for azimuth angles of 30 and 330 degrees.
         :py:class:`~pyfar.Coordinates`
             Return HRIRs at positions defined by a pyfar Coordinates object.
+            Note that the HRIRs were measured at a fixed radius of 1.7 m.
+            Hence, the radius if `position` will be ignored.
 
         The default is ``[[0, 0]]``, which returns the HRIR for frontal sound
         incidence. A ValueError is raised if the requested position is not
--- a/pyfar/signals/files.py
+++ b/pyfar/signals/files.py
@@ -400,27 +400,32 @@ def head_related_impulse_responses(
         idx = (idx[np.argsort(polar)], )
     else:
 
-        # get coordinates - required for more verbose error handling below
         if isinstance(position, pf.Coordinates):
-            position = pf.rad2deg(position.spherical_elevation)[..., :2]
+            # force radius to be equal to those of HRIRs
+            position.radius = 1.7
+        else:
+            # convert to pyfar coordinates for convenience
+            position = np.atleast_2d(position) / 180 * np.pi
+            position = pf.Coordinates.from_spherical_elevation(
+                position[:, 0], position[:, 1], 1.7)
 
-        idx = []
-        for pos in position:
-            find = pf.Coordinates.from_spherical_elevation(
-                pos[0] / 180 * np.pi, pos[1] / 180 * np.pi, 1.7)
-            idx_current, distance = sources.find_nearest(
-                find, distance_measure='spherical_radians')
-            if distance < tolerance_rad:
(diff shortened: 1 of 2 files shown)


pyfar/signals/files.py
```diff
@@ -389,15 +390,20 @@ def head_related_impulse_responses(
...
+    elif isinstance(position, str) and position == "median":
         idx = sources.lateral == 0
         idx = np.where(idx)[0]
         # sort positions according to polar angle
         polar = sources.polar[idx].flatten()
         idx = (idx[np.argsort(polar)], )
     else:
+
+        # get coordinates - required for more verbose error handling below
+        if isinstance(position, pf.Coordinates):
```
ahms5, line 404: maybe convert non coords into pyfar coordintes, we are using them later anyway. In this was its very inefficient

ahms5: and the part below would be more clean

f-brinkmann: Done as suggested

pyfar/signals/files.py
```diff
@@ -348,11 +348,12 @@ def head_related_impulse_responses(
...
             Return median plane HRIRs with an angular resolution of 2 degrees.
-        List of coordinates
-            Return HRIRs at specific positions defined by a list of azimuth
-            and elevation values in degrees. For example
+        Array like
+            Return HRIRs at azimuth and elevation in degrees. For example
             ``[[30, 0], [330, 0]]`` returns HRIRs on the horizontal plane
             (0 degree elevation) for azimuth angles of 30 and 330 degrees.
+        :py:class:`~pyfar.Coordinates`
+            Return HRIRs at positions defined by a pyfar Coordinates object.
```
sikersten, line 356: I'd suggest adding the info, that all radii are set to 1.7 m here, or that the radii are ignored.

sikersten (approved):
Thanks! Only a minor comments, otherwise approved.

pyfar/signals/files.py
f-brinkmann: good point - I've added a note on that.

mberz (approved):

pyfar/signals/files.py
```diff
@@ -348,11 +348,14 @@ def head_related_impulse_responses(
...
+            Note that the HRIRs were measured at a fixed radius of 1.7 m.
+            Hence, the radius if `position` will be ignored.
```
ahms5, line 358:
```suggestion
            Hence, the radius in `position` will be ignored.
```

pyfar/signals/files.py
```diff
@@ -389,33 +392,43 @@ def head_related_impulse_responses(
...
-                    "information."))
+
+        if isinstance(position, pf.Coordinates):
+            # force radius to be equal to those of HRIRs
+            position.radius = 1.7
+        else:
+            # convert to pyfar coordinates for convenience
+            position = np.atleast_2d(position) / 180 * np.pi
+            position = pf.Coordinates.from_spherical_elevation(
+                position[:, 0], position[:, 1], 1.7)
```
ahms5, line 412:
```suggestion
                position[..., 0], position[..., 1], 1.7)
```
would be more save in case of multidimensional arrays

pyfar/signals/files.py
ahms5: Is it clear how the coorinates are defined? or do we need to metion sth like the view is at (1, 0, 0) and up at (0,0,1) or that we suggest to define them in spherical_elevation coordinates?

ahms5 (commented):
just very small comment or thoughts. otherwise approved

mberz (approved):

ahms5 (approved):

PR #736, opened 2024-12-18
return exact results for octave and third octave for limits out of range

ahms5:
return exact results for octave and third octave for limits out of range.

bug:
```
pf.dsp.filter.fractional_octave_frequencies(
    num_fractions=3, frequency_range=[20000, 96000])
```
returns
```
(array([20000.]), array([19952.62314969]))
```
is now fixed. if frequency_range is not in the standard range, just the exact solution is returned.

### DIFF
--- a/pyfar/dsp/filter/fractional_octaves.py
+++ b/pyfar/dsp/filter/fractional_octaves.py
@@ -30,9 +30,9 @@ def fractional_octave_frequencies(
     Returns
     -------
     nominal : array, float
-        The nominal center frequencies in Hz specified in the standard.
-        Nominal frequencies are only returned for octave bands and third octave
-        bands.
+        The nominal center frequencies are only defined for octave bands and
+        third octave bands in the range between 25 Hz and 20 kHz.
+        For all other cases, ``None``is returned.
     exact : array, float
         The exact center frequencies in Hz, resulting in a uniform distribution
         of frequency bands over the frequency range.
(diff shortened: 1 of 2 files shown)


pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -50,7 +50,7 @@ def fractional_octave_frequencies(
         raise ValueError(
             "The second frequency needs to be higher than the first.")
 
-    if num_fractions in [1, 3]:
+    if num_fractions in [1, 3] and (f_lims[0] > 15) and (f_lims[1] < 22e3):
```
f-brinkmann, line 53: Why not
```suggestion
    if num_fractions in [1, 3] and (f_lims[0] >= 25) and (f_lims[1] <= 20e3):
```
as in the Standard?

f-brinkmann (changes requested):
Thanks for catching! I would suggest two update the docstring as well

```
nominal : array, float, None
        The nominal center frequencies in Hz specified in the standard.
        The Nominal frequencies are only defined for octave bands and third octave
        bands in the range between 25 Hz and 20 kHz. For all other cases, ``None``is returned.
```

pyfar/dsp/filter/fractional_octaves.py
ahms5: changed it to exact limits, could not use ``25*2**(-1/6)`` because its larger than 20 and 20 is the standart input of frequency_range

f-brinkmann (approved):

PR #742, opened 2024-12-27
hot: fix radius check in find_within

ahms5:


tests/test_coordinates_find.py
```diff
@@ -207,6 +207,12 @@ def test_find_within_error():
     with pytest.raises(ValueError, match=match):
         coords.find_within(find, 1, return_sorted=-1)
 
+    find.radius = .1
+    match = 'find_within only works if all points have the same'
+    with pytest.raises(ValueError, match=match):
+        coords.find_within(
+            find, distance_measure='spherical_radians', radius_tol=0.1)
+
```
hoyer-a, line 216:
```suggestion
    find.radius = .1
    radius_tol = 0.1
    match = 'find_within only works if all points have the same radius. '
            f'Differences are larger than {radius_tol}'
    with pytest.raises(ValueError, match=match):
        coords.find_within(
            find, distance_measure='spherical_radians', radius_tol=radius_tol)


```

hoyer-a (commented):
Thanks! Just one thing.

hoyer-a (approved):

pyfar/classes/coordinates.py
```diff
@@ -1853,11 +1853,11 @@ def find_within(
                 points, distance + atol, return_sorted=return_sorted)
         if distance_measure in ['spherical_radians', 'spherical_meter']:
             # determine validate radius
-            radius = self.radius
+            radius = np.concatenate((self.radius, find.radius))
```
f-brinkmann, line 1856: There could be cases where this fails. Using this should be safe

```suggestion
            radius = np.concatenate((self.radius.flatten(), find.radius.flatten()))
```

f-brinkmann (changes requested):
Thanks, I suggest a change for improving the robustness. Otherwise approved.

pyfar/classes/coordinates.py
ahms5: Ive also changed it in find_nearest

f-brinkmann (approved):

PR #743, opened 2024-12-27
add coverage test parallel to pytest

ahms5:


.circleci/config.yml
```diff
@@ -155,6 +155,33 @@ jobs:
...
 
+  coverage:
+    parameters:
+      version:
+        description: "version tag"
+        default: "latest"
+        type: string
+    executor:
+      name: python-docker
+      version: <<parameters.version>>
```
ahms5, line 166:
```suggestion
    docker:
    - image: cimg/python:3.10-browsers
```

f-brinkmann (dismissed):
Nice idea to have this also on circle :)

mberz:
I like the idea. However, with a runtime of over 7 min it's almost twice as time costly as running the test. Do you know why it's taking this long?
To me it seems a resource intensive for something that is not very crucial.

mberz:
Creating the coverage report runs all test anyways, right? Can't this then be done when running the tests the first time or is it an truly separate operation?

PR #752, opened 2025-01-03
improve error message for read_sofa

ahms5:
previous error message was:
``Position:Type {pos_type} is not supported.`` without naming the type


f-brinkmann (approved):
Good point

hoyer-a (approved):

PR #755, opened 2025-01-17
Filter impulse response

### LINKED ISSUE #749: Feature request: easy Filter to Signal conversion like filter.to_ir()
As a regular pyfar user, I often find the filter classes to be inconvenient. The main problems I have with filter objects are:
- They cannot be directly plotted using pf.plot functions
- so far, FilterFIR always uses direct convolution, which is almost always painfully slow compared to FFT/overlap-add convolutions
- They don't work with utilities like dsp.linear_phase()

As a consequence, I find myself writing code like the following in pretty much every script I write with pyfar:
```py
dirac = pf.signals.impulse(ir_length, filter.sampling_rate)
filter_response = filter.process(dirac)
# the actual plotting or processing I want to do with the filter
```

or for FilterFIR:
```py
filter_response = pf.Signal(filter.coefficients, filter.sampling_rate)
```

I discussed this with @f-brinkmann and we concluded that instead of changing all the plotting and utility functions to support filter objects, a convenience function for filters that converts them into a Signal object containing their impulse response would help will all of these concerns.
The method could look like this:
`filter.to_ir(target_length)`, where filter is the Filter object to convert and target_length the number of samples that the impulse response should have. 
If we want more explicit naming, it could also be called `filter.to_impulse_response(len)` or similar, but this would make it less convenient to use in one-liners for plotting etc.

f-brinkmann:

Closes #749

- add `impulse_response` method to `pyfar.Filter` class
- add testing thereof

### DIFF
--- a/pyfar/classes/filter.py
+++ b/pyfar/classes/filter.py
@@ -424,7 +424,7 @@ class FilterFIR(Filter):
 
         Parameters
         ----------
-        n_samples : int, str, optional
+        n_samples : int, optional
             Length in samples for which the impulse response is computed. The
             default is ``None`` in which case the length of the impulse
             response is determined from the filter :py:func:`~FilterFIR.order`.
--- a/pyfar/classes/filter.py
+++ b/pyfar/classes/filter.py
@@ -264,7 +264,9 @@ class Filter(object):
 
     def impulse_response(self, n_samples):
         """
-        Compute the finite impulse response of the filter.
+        Compute or approximate the impulse response of the filter.
+
+        See `impulse_response` methods in derived classes for more details.
 
         Parameters
         ----------
@@ -426,6 +428,8 @@ class FilterFIR(Filter):
             Length in samples for which the impulse response is computed. The
             default is ``None`` in which case the length of the impulse
             response is determined from the filter :py:func:`~FilterFIR.order`.
+            A warning is returned if ``n_samples`` is to short to compute the
+            entire impulse response.
 
         Returns
         -------
--- a/pyfar/classes/filter.py
+++ b/pyfar/classes/filter.py
@@ -264,15 +264,12 @@ class Filter(object):
 
     def impulse_response(self, n_samples):
         """
-        Compute the impulse response of the filter.
-
-        The state of the filter is reset before computing the impulse response
-        and restored afterwards.
+        Compute the finite impulse response of the filter.
 
         Parameters
         ----------
         n_samples : int
-            Length in samples up to which the impulse response is computed.
+            Length in samples for which the impulse response is computed.
 
         Returns
         -------
@@ -421,15 +418,12 @@ class FilterFIR(Filter):
 
     def impulse_response(self, n_samples=None):
         """
-        Compute the impulse response of the filter.
-
-        The state of the filter is reset before computing the impulse response
-        and restored afterwards.
+        Compute the finite impulse response of the filter.
 
         Parameters
         ----------
         n_samples : int, str, optional
-            Length in samples up to which the impulse response is computed. The
+            Length in samples for which the impulse response is computed. The
             default is ``None`` in which case the length of the impulse
             response is determined from the filter :py:func:`~FilterFIR.order`.
 
@@ -520,6 +514,26 @@ class FilterIIR(Filter):
             for idx, coeff in enumerate(self._coefficients):
                 new_state[idx, ...] = spsignal.lfilter_zi(coeff[0], coeff[1])
         return super().init_state(state=new_state)
+
+    def impulse_response(self, n_samples):
+        """
+        Compute the infinite impulse response of the filter for a length of
+        `n_samples`.
+
+        Parameters
+        ----------
+        n_samples : int
+            Length in samples for which the impulse response is computed. Note
+            that this must be sufficiently large for `impulse_response` to be
+            a good estimate of the theoretically infinitely long impulse
+            response of the filter.
+
+        Returns
+        -------
+        impulse_response : Signal
+            The impulse response of the filter.
+        """
+        return super().impulse_response(n_samples)
 
     @staticmethod
     def _process(coefficients, data, zi=None):
@@ -614,6 +628,26 @@ class FilterSOS(Filter):
             for idx, coeff in enumerate(self._coefficients):
                 new_state[idx, ...] = spsignal.sosfilt_zi(coeff)
         return super().init_state(state=new_state)
+
+    def impulse_response(self, n_samples):
+        """
+        Approximate the infinite impulse response of the filter by a finite
+        impulse response.
+
+        Parameters
+        ----------
+        n_samples : int
+            Length in samples for which the impulse response is computed. Note
+            that this must be sufficiently large for `impulse_response` to be
+            a good estimate of the theoretically infinitely long impulse
+            response of the filter.
+
+        Returns
+        -------
+        impulse_response : Signal
+            The impulse response of the filter.
+        """
+        return super().impulse_response(n_samples)
 
     @staticmethod
     def _process(sos, data, zi=None):
--- a/pyfar/classes/filter.py
+++ b/pyfar/classes/filter.py
@@ -277,7 +277,7 @@ class Filter(object):
             The impulse response of the filter.
         """
 
-        # track the sate (better than copying the entire filter)
+        # track the state (better than copying the entire filter)
         if self.state is not None:
             state = self.state.copy()
             self._state = None
--- a/pyfar/classes/filter.py
+++ b/pyfar/classes/filter.py
@@ -262,6 +262,38 @@ class Filter(object):
 
         return filtered_signal
 
+    def impulse_response(self, n_samples):
+        """
+        Compute the impulse response of the filter.
+
+        Parameters
+        ----------
+        n_samples : int
+            Length in samples up to which the impulse response is computed.
+
+        Returns
+        -------
+        impulse_response : Signal
+            The impulse response of the filter.
+        """
+
+        # track the sate (better than copying the entire filter)
+        if self.state is not None:
+            state = self.state.copy()
+            self._state = None
+        else:
+            state = None
+
+        # get impulse response
+        impulse_response = self.process(pf.signals.impulse(
+            n_samples, sampling_rate=self.sampling_rate), reset=True)
+
+        # reset the state if required
+        if state is not None:
+            self._state = state
+
+        return impulse_response
+
     def reset(self):
         """Reset the filter state by filling it with zeros."""
         if self._state is not None:
--- a/pyfar/classes/filter.py
+++ b/pyfar/classes/filter.py
@@ -277,6 +277,14 @@ class Filter(object):
             The impulse response of the filter.
         """
 
+        # check the impulse response length for FIR filters
+        if type(self) == FilterFIR:
+            if self.coefficients.shape[-1] > n_samples:
+                warnings.warn(
+                    ('n_samples should be at least as long as the filter, '
+                     f'which is 4 {self.coefficients.shape[-1]}'),
+                    stacklevel=1)
+
         # track the state (better than copying the entire filter)
         if self.state is not None:
             state = self.state.copy()
--- a/pyfar/classes/filter.py
+++ b/pyfar/classes/filter.py
@@ -517,16 +517,17 @@ class FilterIIR(Filter):
 
     def impulse_response(self, n_samples):
         """
-        Compute the infinite impulse response of the filter for a length of
-        `n_samples`.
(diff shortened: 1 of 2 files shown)


tluebeck (dismissed):
Looks good to me. I just wonder if it wouldn't make sense to at least throw a warning if n_samples is unreasonable. In the case of FIR, if n_samples is shorter than the filter coefficients for example.

jf-som:
Not sure if this is the right place for me to add thoughts, so let me know if not.
Regardless, here are a few things I noticed with the suggested implementation:

- Typo at line 280 ("track the sate" rather than "track the state")
- I suggest that the docstring comment should inform how the filter state is handled (it looks like it uses the current state, but does not modify it, which I don't think is obvious to the user)
- adding to what @tluebeck suggests, what do you think about allowing `None` as `n_samples` argument in case of FIR filters, in which case the function uses the filter's length as default?

f-brinkmann:
> I suggest that the docstring comment should inform how the filter state is handled (it looks like it uses the current state, but does not modify it, which I don't think is obvious to the user)

@jf-som, the state is set to `None` before computing the impulse response and restored afterwards. This will always return the acrtual IR regardless of the state that the filter object has when being passed to `impulse_response`. In addition, calling `impulse_response` does not change the state of the filter object. I think this both expected behavior and I would not mention it in the docs. Doing so might cause additional confusion.

pyfar/classes/filter.py
```diff
@@ -262,6 +262,46 @@ def process(self, signal, reset=False):
...
+            Length in samples up to which the impulse response is computed.
+
+        Returns
+        -------
+        impulse_response : Signal
+            The impulse response of the filter.
+        """
+
+        # check the impulse response length for FIR filters
+        if type(self) == FilterFIR:
```
ahms5, line 281: not sure if I like the idea to have a special behaviour for FIR filters (even if it make scence), an other option is to add a note in the domumentation, that the user need to take care of it.

ahms5 (changes requested):
thank you for taking care!

pyfar/classes/filter.py
jf-som: I agree if statements with type checks are unclean.
A clean option to achieve the same effect is to use the object oriented approach (polymorphism), meaning Filter implements the impulse_response() function without the check, and then FilterFIR overrides it impulse_response() by performing the check, then calling Filter's function if the check is passed.

(Here we could also handle the None value if passed to FilterFIR.impulse_response(), in case my suggestion is still on the table)

pyfar/classes/filter.py
```diff
@@ -262,6 +262,46 @@ def process(self, signal, reset=False):
 
         return filtered_signal
 
+    def impulse_response(self, n_samples):
+        """
+        Compute the impulse response of the filter.
+
+        Parameters
```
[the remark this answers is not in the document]
f-brinkmann: I chose a slightly different formulation :)

pyfar/classes/filter.py
```diff
@@ -262,6 +262,46 @@ def process(self, signal, reset=False):
...
+        impulse_response : Signal
+            The impulse response of the filter.
+        """
+
+        # check the impulse response length for FIR filters
+        if type(self) == FilterFIR:
+            if self.coefficients.shape[-1] > n_samples:
+                warnings.warn(
+                    ('n_samples should be at least as long as the filter, '
+                     f'which is 4 {self.coefficients.shape[-1]}'),
```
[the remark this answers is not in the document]
f-brinkmann: ups

pyfar/classes/filter.py
```diff
@@ -262,6 +262,38 @@ def process(self, signal, reset=False):
...
+        Returns
+        -------
+        impulse_response : Signal
+            The impulse response of the filter.
+        """
+
+        # track the sate (better than copying the entire filter)
+        if self.state is not None:
+            state = self.state.copy()
+            self._state = None
```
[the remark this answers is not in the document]
f-brinkmann: If reset is true, the state is filled with zeros before processing. We thus need this, but maybe its more clear this way

pyfar/classes/filter.py
f-brinkmann: Implemented a method in FilterFIR as suggested.

pyfar/classes/filter.py
```diff
@@ -383,6 +419,36 @@ def init_state(self, cshape, state='zeros'):
...
+            response is determined from the filter :py:func:`~FilterFIR.order`.
+
+        Returns
+        -------
+        impulse_response : Signal
+            The impulse response of the filter.
+        """
+
+        # set or check the impulse response length
+        if n_samples == 'auto':
```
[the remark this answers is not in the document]
f-brinkmann: done

ahms5 (dismissed):

jf-som (approved):
Looks good!

One optional thing we could do would be to use the @override decorator (as described in this SO answer: https://stackoverflow.com/a/75802188 ) to make it more obvious to maintainers that the FilterFIR implementation overrides the implementation in Filter. This is technically not mandatory in python, but nice to have in my opinion.

mberz (commented):
Commented without reviewing above 🤦

f-brinkmann:
> Commented without reviewing above 🤦

I was looking for something like https://nl.mathworks.com/help/signal/ref/impzlength.html#d126e117775 in scipy but did not find anything. Moreover I did not want to implement a warning for IIR filters by only checking the order, so I decided to do nothing about it.

I would suggest to leave things as they are for now but open an isse to port https://nl.mathworks.com/help/signal/ref/impzlength.html#d126e117775. Does not look to hard from a quick check, but maybe out of the scope here.

f-brinkmann:
@mberz do you agree with the above or should this go into this Pull?

pyfar/classes/filter.py
```diff
@@ -262,6 +262,42 @@ def process(self, signal, reset=False):
 
         return filtered_signal
 
+    def impulse_response(self, n_samples):
+        """
+        Compute the impulse response of the filter.
```
mberz, line 267:
```suggestion
        Compute the finite impulse response of the filter.
```

pyfar/classes/filter.py
```diff
@@ -262,6 +262,42 @@ def process(self, signal, reset=False):
...
+
+        The state of the filter is reset before computing the impulse response
+        and restored afterwards.
+
+        Parameters
+        ----------
+        n_samples : int
+            Length in samples up to which the impulse response is computed.
```
mberz, line 275:
```suggestion
            Length in samples for which the impulse response is computed.
```

pyfar/classes/filter.py
```diff
@@ -383,6 +419,36 @@ def init_state(self, cshape, state='zeros'):
...
+        """
+        Compute the impulse response of the filter.
+
+        The state of the filter is reset before computing the impulse response
+        and restored afterwards.
+
+        Parameters
+        ----------
+        n_samples : int, str, optional
+            Length in samples up to which the impulse response is computed. The
```
mberz, line 432:
```suggestion
            Length in samples for which the impulse response is computed. The
```

pyfar/classes/filter.py
```diff
@@ -383,6 +419,36 @@ def init_state(self, cshape, state='zeros'):
                 new_state[idx, ...] = spsignal.lfilter_zi(coeff[0], coeff[1])
         super().init_state(state=new_state)
 
+    def impulse_response(self, n_samples=None):
+        """
+        Compute the impulse response of the filter.
+
+        The state of the filter is reset before computing the impulse response
+        and restored afterwards.
```
mberz, line 427:
```suggestion
```
See comment above.

pyfar/classes/filter.py
```diff
@@ -383,6 +419,36 @@ def init_state(self, cshape, state='zeros'):
                 new_state[idx, ...] = spsignal.lfilter_zi(coeff[0], coeff[1])
         super().init_state(state=new_state)
 
+    def impulse_response(self, n_samples=None):
+        """
+        Compute the impulse response of the filter.
```
mberz, line 424:
```suggestion
        Compute the finite impulse response of the filter.
```

mberz (changes requested):
I don't mind having the length check in a separate PR. However, I think it would be beneficial to add a sentence on it in the respective methods for FilterIIR and FIlterSOS, which would need to be generated for this.

I've alread created an issue for the missing length check #766.

pyfar/classes/filter.py
```diff
@@ -262,6 +262,42 @@ def process(self, signal, reset=False):
 
         return filtered_signal
 
+    def impulse_response(self, n_samples):
+        """
+        Compute the impulse response of the filter.
+
+        The state of the filter is reset before computing the impulse response
+        and restored afterwards.
```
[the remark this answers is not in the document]
jf-som: I think it's better to keep this info. It says 2 things:
1. The state of the filter is unaffected by this method.
2. The returned impulse response in unaffected by the filter's current state

(1) is rather obvious, as you say, but I think it's good to clarify (2), since the response of a filter is different when it's in its zero state compared to when it just processed some larger samples, changing the beginning of the IR.

Perhaps we could rephrase it to make it simpler though?

pyfar/classes/filter.py
f-brinkmann: changed as suggested

pyfar/classes/filter.py
f-brinkmann: changed as suggested.

pyfar/classes/filter.py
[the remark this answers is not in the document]
f-brinkmann: I tend to agree with mberz and removed the phrase. I would rather think of this as expected behavior and expect a note only if the user would have to take care of this.

pyfar/classes/filter.py
f-brinkmann: removed as well

pyfar/classes/filter.py
f-brinkmann: done as suggested.

pyfar/classes/filter.py
```diff
@@ -521,6 +515,26 @@ def init_state(self, cshape, state):
                 new_state[idx, ...] = spsignal.lfilter_zi(coeff[0], coeff[1])
         return super().init_state(state=new_state)
 
+    def impulse_response(self, n_samples):
+        """
+        Compute the infinite impulse response of the filter for a length of
+        `n_samples`.
```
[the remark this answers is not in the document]
jf-som: Shouldn't it be "approximate the finite impulse response" or 
"approximate the infinite impulse response as a finite impulse response"? 
Infinite would always imply recursive definitions, but we return a signal object. Also it would conflict the superclass comment if it says finite but the subclass here says infinite.

pyfar/classes/filter.py
```diff
@@ -383,6 +416,33 @@ def init_state(self, cshape, state='zeros'):
                 new_state[idx, ...] = spsignal.lfilter_zi(coeff[0], coeff[1])
         super().init_state(state=new_state)
 
+    def impulse_response(self, n_samples=None):
+        """
+        Compute the finite impulse response of the filter.
+
+        Parameters
```
ahms5, line 423:
```suggestion
        This method will return a warning if the ``n_samples`` is smaller than the number of coefficients.
        
        Parameters
```
I would like to add the behaviour to the docstring as well.

pyfar/classes/filter.py
```diff
@@ -455,6 +515,27 @@ def init_state(self, cshape, state):
...
         return super().init_state(state=new_state)
 
+    def impulse_response(self, n_samples):
+        """
+        Approximate the infinite impulse response of the filter by a finite
+        impulse response.
+
+        Note that the number of samples must be sufficiently long for
+        `impulse_response` to be a good approximation of the theoretically
+        infinitely long impulse response of the filter.
```
ahms5, line 525: i like

ahms5 (changes requested):
just small doc stuff

mberz (dismissed):

pyfar/classes/filter.py
```diff
@@ -383,6 +418,35 @@ def init_state(self, cshape, state='zeros'):
                 new_state[idx, ...] = spsignal.lfilter_zi(coeff[0], coeff[1])
         super().init_state(state=new_state)
 
+    def impulse_response(self, n_samples=None):
+        """
+        Compute the finite impulse response of the filter.
+
+        Parameters
+        ----------
+        n_samples : int, str, optional
```
jf-som, line 427:
```suggestion
        n_samples : int, optional
```
I don't think we can pass strings here, can we? If we do, mention which strings are possible in the explanation.

jf-som (changes requested):
One wrong comment, otherwise all good.

PR #756, opened 2025-01-17
Add missing default in `pyfar.FilterIIR.init_state`

f-brinkmann:

Technically a bug, but I assume this function is rarely used. Make it a bug or release with the next features?

- Add missing default in `pyfar.FilterIIR.init_state`
- Test default value
- Implement and test raising an error for invalud values

### DIFF
--- a/pyfar/classes/filter.py
+++ b/pyfar/classes/filter.py
@@ -164,6 +164,15 @@ class Filter(object):
         self._state = state
         self._initialized = True
 
+    @staticmethod
+    def _check_state_keyword(state):
+        """
+        Check the value of the state keyword for 'ini_state' class methods.
+        """
+        if state not in ['zeros', 'step']:
+            raise ValueError(
+                f"state is '{state}' but must be 'zeros' or 'step'")
+
     @property
     def coefficients(self):
         """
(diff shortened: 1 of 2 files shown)


hoyer-a:
Should we add a check `if state in ['zeros', 'step']`?
Otherwise one could put in any string and still get zeros.

ahms5 (changes requested):
agreed with anton, and add tests for defaults, and I think It should go to main.

f-brinkmann:
Changed/added everything as suggested

pyfar/classes/filter.py
```diff
@@ -164,6 +164,15 @@ def init_state(self, state='zeros'):
         self._state = state
```
ahms5, line 164: should we remove the state='zero' here, because in this method we set the state as the array and not as a config string.
this funtion init_state is more a set_state anyway, even its not checking the inputs, (which should be done acutally). maybe shift the check of state from the init into this one.

overall its very confusing, that the state in the init is an array and in init_state (except for Filter) is a string. Maybe we should consider another name, but thats an other story.

ahms5 (commented):
maybe not directly to this PR but related.

pyfar/classes/filter.py
ahms5: moved into #757 
but maybe we should remove the wron default here state='zero'

f-brinkmann: Yes, this is wrong. Fortunately it is called from the child classes in any case

ahms5 (approved):

hoyer-a (approved):

PR #759, opened 2025-01-22
Feature: frequency weighting filter

### LINKED ISSUE #758: feature: frequency weighting filters
Add a feature to provide A and C weighting curves as described in IEC/DIN-EN 61672-1 as IIR (SOS) filters

jf-som:

Closes #758

-Adds a function to design efficient A/C-weighting filters via least squares approximation (since there is no analytical design method for digital filters of this kind)

### Notes
Since I am unfamiliar with the test module structure etc, this is currently all in one file to be cleaned up and integrated into the module structure later.
For now, this is a draft, based on which we can discuss the details of the implementation.

This is how the function works:
- Calculate the ideal A/C-weighting curves from the formulae given in the norm
- Transform the ideal analog weighting filter to a warped digital filter via bilinear transform as a starting point
- run a least squares optimizer that moves the zeros and poles of the SOS filter on the z-plane until the resulting filter response does not get closer to the target curve anymore. The zeros and poles are restricted to the real axis for simplicity and because it works well
- Check whether the curve is within the error margins provided by the norm before returning it. Currently only prints a warning if the class 1 requirements are not met.

I have set the function up in a way such that the default parameters yield class 1 compliant filters for all common sample rates (48k, 44.1k, 16k as well as their 2x, 4x, 8, 0.5x, 0.25x, 0.125x multiples), while also providing optional parameters to give users more control in case they want/need to manually find the best approximation possible.

This way, the optimizer "just works" when used with only the requires arguments, but may find "only" a good local optimum rather than the best, global optimum.

I have added a simple test for the common sample rates to demonstrate the default parameters work quite well, as well as a manual testing segment with plots for more testing. 
The attached images show how a typical resulting filter's response looks like when optimized with default parameters (Example has fs=48k and A weighting).

grafik
grafik

### DIFF
--- a/pyfar/dsp/filter/frequency_weighting.py
+++ b/pyfar/dsp/filter/frequency_weighting.py
@@ -1,0 +1,355 @@
+"""
+Frequency weighting (A, C) according to IEC/DIN-EN 61672-1.
+"""
+
+# %%
+from collections.abc import Callable
+
+import pyfar as pf
+import scipy.signal as sps
+import scipy.optimize as spo
+import numpy as np
+import matplotlib.pyplot as plt
+
+
+# constants for the formulas
+F_1 = 20.598997057618316
+F_2 = 107.65264864304629
+F_3 = 737.8622307362901
+F_4 = 12194.217147998012
+A_1000 = -2
+C_1000 = -0.062
+A_WEIGHTING_ZEROS = [0, 0, 0, 0]
+A_WEIGHTING_POLES = [F_1, F_1, F_4, F_4, F_2, F_3]
+C_WEIGHTING_ZEROS = [0, 0]
+C_WEIGHTING_POLES = [F_1, F_1, F_4, F_4]
+
+
+def calculate_A_weighted_level(f):
+    """
+    Calculates the level correction in dB of a frequency component
+    when using the A weighting.
+    """
+    bracket_term = (F_4**2 * f**4) / ((f**2 + F_1**2) * (f**2 + F_2**2)
+                                      ** 0.5 * (f**2 + F_3**2)**0.5
+                                      * (f**2 + F_4**2))
+    return 10 * np.log10(bracket_term**2) - A_1000
+
+
+def calculate_C_weighted_level(f):
+    """
+    Calculates the level correction in dB of a frequency component
+    when using the C weighting.
+    """
+    bracket_term = (F_4**2 * f**2) / ((f**2 + F_1**2) * (f**2 + F_4**2))
+    return 10 * np.log10(bracket_term**2) - C_1000
+
+
+def calculate_weighted_level(weighting: str,
+                             frequency: float | np.ndarray,
+                             ) -> float | np.ndarray:
+    """
+    Calculates the level correction in dB of a frequency component
+    when using the A or C weighting.
+    """
+    if weighting == "A":
+        return calculate_A_weighted_level(frequency)
+    elif weighting == "C":
+        return calculate_C_weighted_level(frequency)
+    else:
+        raise ValueError("Unrecognized weighting:", weighting)
+
+
+def _zpk_from_analog(weighting: str, fs: float,
+                     ) -> tuple[np.ndarray, np.ndarray, float]:
+    """
+    Transforms the analog (s-plane) A or C weighting filter as specified in
+    IEC/DIN-EN 61672-1 to a digital filter (z-plane) via bilinear transform
+    and returns it as zeros poles and gain.
+    The resulting filter is heavily warped near the Nyquist frequency.
+    """
+    if weighting == "A":
+        zeros, poles = (A_WEIGHTING_ZEROS, A_WEIGHTING_POLES)
+    elif weighting == "C":
+        zeros, poles = (C_WEIGHTING_ZEROS, C_WEIGHTING_POLES)
+    else:
+        raise ValueError("Unrecognized weighting:", weighting)
+
+    poles_angular = np.array(poles) * -2 * np.pi
+    return sps.bilinear_zpk(zeros, poles_angular, 1, fs)
+
+
+def design_frequency_weighting_filter(sampling_rate: float,
+                                      target_weighting="A",
+                                      n_frequencies=100,
+                                      error_weighting: Callable[[
+                                          np.ndarray], np.ndarray] = None,
+                                      **kwargs,
+                                      ) -> pf.FilterSOS:
+    """
+    Creates an efficient SOS filter approximating the A or C weighting defined
+    in IEC/DIN-EN 61672-1.
+    When using default parameters, the returned filter is compliant with a
+    class 1 sound level meter as described in the norm for all tested
+    sampling rates (see Notes).
+    This function will run a least-squares algorithm to iteratively approach
+    the target weighting curve.
+    Therefore, it may be much faster to create the filter once and reuse that
+    filter when repeatedly weighting audio data of identical sampling rates.
+
+    Parameters
+    ----------
+    sampling_rate: float
+        The sampling rate of the returned filter.
+
+    target_weighting: str, optional
+        Specifies which frequency weighting curve to approximate.
+        Must be either "A" or "C". The default is "A".
+
+    n_frequencies: int, optional
+        At how many frequencies to evaluate the filter coefficients during
+        optimization. Less frequencies means faster iterations, but
+        potentially worse results. The evaluation frequencies are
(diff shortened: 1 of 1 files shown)


pyfar/io/io.py
f-brinkmann, line 1: Is this coming from a different pull request?

tests/test_io.py
f-brinkmann, line 1: Is this coming from a different pull-request

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,355 @@
...
+F_4 = 12194.217147998012
+A_1000 = -2
+C_1000 = -0.062
+A_WEIGHTING_ZEROS = [0, 0, 0, 0]
+A_WEIGHTING_POLES = [F_1, F_1, F_4, F_4, F_2, F_3]
+C_WEIGHTING_ZEROS = [0, 0]
+C_WEIGHTING_POLES = [F_1, F_1, F_4, F_4]
+
+
+def calculate_A_weighted_level(f):
```
f-brinkmann, line 28: We could discuss which functions should be public. All public functions should have complete docstrings.

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,355 @@
...
+
+    error_weighting: callable
+        A function that can be used to emphasize the approximation errors in
+        specific frequency ranges. The function should take a float array as
+        argument, specifying the normalized frequencies between 0 and 1
+        (where 1 is the Nyquist frequency) to weight, and returns a float
+        array as output containing the weights.
+        The default value is None, in which case the errors of all
+        (logarithmically spaced) frequencies are equally weighted. This
+        usually leds to larger errors in the higher frequencies. By passing
```
f-brinkmann, line 123:
```suggestion
        usually leds to larger errors for higher frequencies. By passing
```

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,355 @@
...
+        a function that emphasizes high frequencies, it is possible to reduce
+        this effect and get a filter potentially closer to the target curve.
+        Example: `error_weighting=lambda nf: 100**nf`. This example often
+        leads to better results for typical sampling rates, but much worse
+        for very high rates.
+
+    **kwargs: dict
+        Keyword args that are passed to the
+        scipy.optimize.least_squares() call.
```
f-brinkmann, line 132: We could link to the documentation using :py:func:`scipy.optimize.least_squares`

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,355 @@
...
+    Transforms the analog (s-plane) A or C weighting filter as specified in
+    IEC/DIN-EN 61672-1 to a digital filter (z-plane) via bilinear transform
+    and returns it as zeros poles and gain.
+    The resulting filter is heavily warped near the Nyquist frequency.
+    """
+    if weighting == "A":
+        zeros, poles = (A_WEIGHTING_ZEROS, A_WEIGHTING_POLES)
+    elif weighting == "C":
+        zeros, poles = (C_WEIGHTING_ZEROS, C_WEIGHTING_POLES)
+    else:
```
f-brinkmann, line 75: Since its a private function, we maybe don't need the additional check

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,355 @@
...
+        """
+        fixed_zeros = [1, 1, 1, 1] if target_weighting == "A" else [1, 1]
+        zeros = np.concat([fixed_zeros, x[0:2]])
+        return (zeros, x[2:-1], x[-1])
+
+    # build the cost function
+    frequencies = np.logspace(1, np.log10(sampling_rate/2), n_frequencies)
+    target_levels = calculate_weighted_level(target_weighting, frequencies)
+
+    def compute_residuals(x: np.ndarray) -> np.ndarray:
```
f-brinkmann, line 180: maybe more cleaner to make this a private function of the module?

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,355 @@
...
+    # zeros and poles along the real axis on the z-plane until
+    # a local optimum is reached.
+    optimization_result = spo.least_squares(compute_residuals, x0, **kwargs)
+
+    # turn the result into a filter object
+    z, p, k = x2zpk(optimization_result.x)
+    sos = sps.zpk2sos(z, p, k)
+    weighting_filter = pf.FilterSOS([sos], sampling_rate)
+
+    # ensure the filter is correct
```
f-brinkmann, line 210:
```suggestion
    # check if the filter is class 1 compliant
```

f-brinkmann (commented):
Thanks for the effort. It generally looks good to me. I think it would be good to discuss what functions should be public and where they should go before doing continuing with the implementation

pyfar/dsp/filter/frequency_weighting.py
ahms5: [here](https://pyfar-gallery.readthedocs.io/en/latest/contribute/contribution_guidelines.html#writing-the-documentation) you can find more information about our docsting policy.

pyfar/io/io.py
jf-som: This is the last commit from the main branch. It happened because I based my branch on main, but then selected to merge it into develop instead. Since this is an already merged feature, it should make no difference.

pyfar/dsp/filter/frequency_weighting.py
jf-som: Agree, though reading it again I also realize it should be "leads" and not "leds"... oops

pyfar/dsp/filter/frequency_weighting.py
jf-som: True, though I like to be safe against accidents during future changes.

pyfar/dsp/filter/frequency_weighting.py
jf-som: You mean to move `compute_residuals()` outside of the parent function?
I thought about this, but I decided to leave it as a nested closure, because:
- it has no use outside of this function
- we would need to pass a lot more arguments into it (frequencies, sampling_rate, target_levels, error_weighting)
- we would have to define a local function anyway when passing `compute_residuals` as an argument into `least_squares`, unless we want to write is as a lambda that is several lines long.

So I think making it a closure (local function that captures surrounding variables) was the easier and cleaner move.
The same reasoning applies to x2zpk, except for my third bullet point.

pyfar/dsp/filter/frequency_weighting.py
jf-som: Yes, I was treating these formula-like functions as private at first, but then thought they might be useful for the user as well and wanted to hear your opinions first before making super detailed documentation.
I personally thing having either the two direct formulas )`calculate_A_weighted_level` and `calculate_C_weighted_level`) or the combined function `calculate_weighted_level` (probably better renamed) could be nice if the user just wants to know the A weighting for a specific frequency.

- Do you agree this would be good to have public?
- Which version (two functions or one function with "A" or "C" parameter) do you prefer? (or both?)
- Does this belong into the filter module or should it go with the other planned functionality, like the third-band correction constants for A/C weighting that are still on the ToDo?

pyfar/dsp/filter/frequency_weighting.py
jf-som: done

pyfar/dsp/filter/frequency_weighting.py
jf-som: fixed

jf-som:
These are the remaining points of discussion I see for this feature currently:

- [ ]  which frequency weighting formula functions should be public and where? (see this conversation thread: https://github.com/pyfar/pyfar/pull/759#discussion_r1926885163)
- [ ] should this function accept Signal objects the same way all other filter functions do? This would make the module more uniform, although one could argue that this function is special since it has no perfect solution and takes longer to compute, making code potentially very slow when this function is used in a loop.
- [ ] Related to the point above, should we apply caching (like so: https://docs.python.org/dev/library/functools.html#functools.lru_cache) to the zpk-calculation? This would make the function very fast on repeated calls with the same parameters, for instance when calling the frequency weighting function to process many signals which all have the same sampling rates.
- [ ] The final name for this function. If we decide it should accept Signal objects, perhaps just `frequency_weighting`, but when we decide it's too special I suggest including the word "design" or similar.

Issue #762, opened 2025-01-24
feature: combined sound pressure level convenience function


jf-som:
The concept here was to have a cure-all level metering function capable of providing a variety of different SPL metering tools in all their combinations:
- Frequency weighting: A/C/Z
- Time weighting: fast, slow or L_eq
- frequency band resolution: full-band, octave-band or third-octave-band

This requires other sub-issues of #675 to be finished first.

Open questions and things to consider:
- depending on the time weighting and band resolution, the output could be everything from a single scalar (full-spectrum L_eq) to a 2D numpy array (fast/slow weighting + octave/3rd-octave bands). Do we accept different return types depending on input parameters?
- Are we fine with running the A/C weighting filter approximation algorithm from #759 every time this function is called or do we allow passing it as parameter, caching it, etc... ?

Issue #675, opened 2024-08-26
It would be nice to have Bewertungskurven and Pegelmessung


f-brinkmann:

tjueterb:
A- B- and C-weighting could be adapted from this:

https://github.com/endolith/waveform-analysis/blob/master/waveform_analysis/weighting_filters/ABC_weighting.py

f-brinkmann:
> A- B- and C-weighting could be adapted from this:
> https://github.com/endolith/waveform-analysis/blob/master/waveform_analysis/weighting_filters/ABC_weighting.py

The implementation here is not in agreement with the ISO norm. Differences between the magnitude response of the IIR filters and the norm are too large. Johannes found a better way of realizing the IIR filters and tries to get it working

f-brinkmann:
We want this and this is the first working proposal for the structure

- IIR filter ABCD -> pyfar.dsp.filter
- Octave/third-octave correction values for ABCD -> pf.dsp.filter
- slow/fast time weighting -> pyfar.dsp.filter
- compute level from Signal -> pf.dsp.level

Do we want something from: https://mosqito.readthedocs.io?

jf-som:
Alright, I get get to implementing this soon. Some initial questions:

filters and correction values:
- are B and D curves necessary? They aren't even defined in the IEC/DIN EN 6172-1, where A, C, fast and slow are defined

compute level from Signal:
- do you mean it should be a function in pf.dsp called sound_pressure_level() or a group of functions in a new folder called sound_pressure_level?
- which level? Leq, short-term Leq, weighted...?

slow/fast time weighting:
- imo this is not intuitive in to go in dsp.filter, because time weighting always outputs power/level over time. The first step in the algorithm is squaring the input signal. However, for dsp.filter I'd expect the output to be a pressure signal like the input.
- This leads to the question of the output type. A Signal would be an awkward choice, since the output is a level/energy over time function, so the a frequency domain transform of it makes little sense. Perhaps a TimeData or just a numpy array of the same shape as the input_signal.time are more suitable?

mosquito:
I feel like that's a topic for another issue/thread, though I do think loudness etc would be similarly useful as weighted levels

mberz:
As far as I see, parts of this are already integrated in the next release (v0.8.0).
Please update if the third sub-issue is still intended to go into the next release or if we should schedule that for a future (undefined) release.

jf-som:
This doesn't have to be in v0.8, no.
[...]

PR #763, opened 2025-01-27
Feature/time weighted levels

### LINKED ISSUE #760: feature: time weighted sound pressure level
Create a function that performs fast and slow time weighting according to IEC/DIN EN 61672

jf-som:

Closes #760 

- adding a new function to calculate time weighted sound pressure levels in accordance with IEC / DIN EN 61672-1

Here is how plots look for the pf.signals.files.guitar example: 
image

### ToDo and to discuss:

- [ ] So far the function handles each channel separately. Should this function have options to merge channels?
- [ ] What to do about the sections with 0 energy in the numpy.log10 step? Right now it simply suppresses numpy warnings and returns these values as negative infinity, but we could add a parameter so the user can provide a default value.
- [ ] How to handle reference values: Right now it defaults to use 1 as reference, meaning the result is in dBFS. There are also reasons to use 2e-5 as default to get dB SPL, but then we still to know the sensitivity unless we just expect the input to be in Pascals, which i find unlikely in most use cases.

### DIFF
--- a/pyfar/dsp/__init__.py
+++ b/pyfar/dsp/__init__.py
@@ -34,6 +34,10 @@ from .interpolation import (
     InterpolateSpectrum,
 )
 
+from .levels import (
+    time_weighted_levels,
+)
+
 from . import filter
 from . import fft
 
@@ -69,4 +73,5 @@ __all__ = [
     'normalize',
     'fractional_time_shift',
     'correlate',
+    'time_weighted_levels',
 ]
--- a/pyfar/dsp/levels.py
+++ b/pyfar/dsp/levels.py
@@ -1,0 +1,63 @@
+"""
+Functions used to calculate the levels of signals.
+"""
+
+import numpy as np
+import scipy.signal as sps
+
+
+def time_weighted_levels(signal, weighting: str,
+                         reference=1) -> np.ndarray:
+    """
+    Calculates sound pressure levels with exponential time weighting
+    according to IEC / DIN EN 61672-1.
+
+    Paramters:
+    ---------
+
+    signal: Signal
+        The signal object to calculate the levels for
+
+    type: str
+        The time weighting type. Options are "F" (="fast") and "S" (="slow"),
+        which correspond to level decays of -34.7 dB and -4.3 dB,
+        respectively.
+
+    reference: float
+        The reference sound pressure in Pascal to calculate the level
+        for. Defaults to 1, in which case the returned level is in dBFS
+        (decibels to full scale). If the input signal is in Pascal, use
+        `2e-5` to obtain a sound pressure level (SPL).
+
+    Returns
+    -------
+    A numpy array of the same shape as `signal.time` with the current time
+    weighted level for each point in time and channel.
+    """
+    weighting = weighting.upper()
+    if weighting in ["F", "FAST"]:
+        time_constant = 0.125
+    elif weighting in ["S", "SLOW"]:
+        time_constant = 1
+    else:
+        raise ValueError(f"Unknown 'type' parameter: {weighting}")
+
+    energies = signal.time**2
+
+    # the e-function defined in the norm, but while the norm defines it
+    # as an integral, we can implement it recursively as a scaling factor
+    # for performance.
+    sample_duration = 1 / signal.sampling_rate
+    exp_func_decay = np.exp(-sample_duration / time_constant)
+
+    # this is effectively the same as looping over all samples and doing
+    # > weighted[i] = exp_func_decay * weighted[i-1] + energies[i]
+    # but it's faster to use an lfilter call to make use of scipy's speed
+    weighted = sps.lfilter([1, 0], [1, -exp_func_decay], energies)
+
+    # normalize the integral
+    normalized = weighted / time_constant / signal.sampling_rate
+
+    with np.errstate(divide="ignore"):
+        levels = 10 * np.log10(normalized / reference**2)
+    return levels
--- a/tests/test_time_weighted_levels.py
+++ b/tests/test_time_weighted_levels.py
@@ -1,18 +1,17 @@
-import pytest
 import pyfar as pf
 import numpy as np
 
 def test_time_weighted_levels_decay():
     fs = 48000
     x = pf.signals.impulse(2 * fs, sampling_rate=fs)
-    
+
     levels_F = pf.dsp.time_weighted_levels(x, "F")
     diff1_F =  levels_F[0][fs] - levels_F[0][0]
     diff2_F = levels_F[0][fs + 100] - levels_F[0][100]
     F_DECAY = -34.7 # DIN EN 61672-1 §5.8.2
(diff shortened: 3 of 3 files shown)


pyfar/dsp/levels.py
```diff
@@ -0,0 +1,63 @@
...
+
+import numpy as np
+import scipy.signal as sps
+
+
+def time_weighted_levels(signal, weighting: str,
+                         reference=1) -> np.ndarray:
+    """
+    Calculates sound pressure levels with exponential time weighting
+    according to IEC / DIN EN 61672-1.
```
f-brinkmann, line 13: - Would be nice to add the reference below.
- For convenience, we could also give the formula for the level and the time weighting for people without access to the standard.
- I would suggest to wait with this until the frequency weightings are merged and add an option for that here as well. In this case the input parameter `weighting` should probably become `time_weighting` or `temporal_weigting` (the term used in the standard)

pyfar/dsp/__init__.py
```diff
@@ -34,6 +34,10 @@
     InterpolateSpectrum,
 )
 
+from .levels import (
```
f-brinkmann, line 37: Lets discuss in the weekly where this and related pulls should be imported into pyfar. There is more stuff such as the L_eq and and L_E to which we might want to implement in the future.

pyfar/dsp/levels.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Paramters:
+    ---------
+
+    signal: Signal
+        The signal object to calculate the levels for
+
+    type: str
+        The time weighting type. Options are "F" (="fast") and "S" (="slow"),
```
f-brinkmann, line 22: We allways render actual inputs as code and for being verbose I would suggest to expect full names
```suggestion
        The time weighting type. Options are ``'fast``' and ``'slow``',
```

pyfar/dsp/levels.py
```diff
@@ -0,0 +1,63 @@
+"""
+Functions used to calculate the levels of signals.
+"""
+
+import numpy as np
+import scipy.signal as sps
+
+
+def time_weighted_levels(signal, weighting: str,
```
f-brinkmann, line 9: I dont have access to the english version. Is this the official name for that level computation? We should use the exact term from the standard in my opinion.

pyfar/dsp/levels.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Paramters:
+    ---------
+
+    signal: Signal
+        The signal object to calculate the levels for
+
+    type: str
+        The time weighting type. Options are "F" (="fast") and "S" (="slow"),
+        which correspond to level decays of -34.7 dB and -4.3 dB,
```
f-brinkmann, line 23:
```suggestion
        which correspond to level decays of -34.7 dB/s and -4.3 dB/s,
```

pyfar/dsp/levels.py
```diff
@@ -0,0 +1,63 @@
...
+
+    reference: float
+        The reference sound pressure in Pascal to calculate the level
+        for. Defaults to 1, in which case the returned level is in dBFS
+        (decibels to full scale). If the input signal is in Pascal, use
+        `2e-5` to obtain a sound pressure level (SPL).
+
+    Returns
+    -------
+    A numpy array of the same shape as `signal.time` with the current time
```
f-brinkmann, line 34: I would suggest to return a TimeData object. This way the level can easily be plotted using `pf.plot.time(level, dB=False)`

pyfar/dsp/levels.py
```diff
@@ -0,0 +1,63 @@
...
+    weighted level for each point in time and channel.
+    """
```
f-brinkmann, line 36: It would be nice to add an example, e.g., the one from the pull request.

pyfar/dsp/levels.py
```diff
@@ -0,0 +1,63 @@
...
+
+    # this is effectively the same as looping over all samples and doing
+    # > weighted[i] = exp_func_decay * weighted[i-1] + energies[i]
+    # but it's faster to use an lfilter call to make use of scipy's speed
+    weighted = sps.lfilter([1, 0], [1, -exp_func_decay], energies)
+
+    # normalize the integral
+    normalized = weighted / time_constant / signal.sampling_rate
+
+    with np.errstate(divide="ignore"):
```
f-brinkmann, line 61: `-inf` are a little annoying sometimes. I would suggest one of the following
- use `normalized[normalized == 0] = finfo(normalized).eps` instead, or
- add a parameter to let the user decide which approach should be taken

I tend to prefer the first solution and mention it in the docstring.

tests/test_time_weighted_levels.py
f-brinkmann, line 1: The name of the file should be test_module_function.py - but for this, we first have to decide where this should go...

tests/test_time_weighted_levels.py
```diff
@@ -0,0 +1,38 @@
+import pyfar as pf
+import numpy as np
+
+def test_time_weighted_levels_decay():
```
f-brinkmann, line 4: Could be parametrized to reduce code redundancy
```suggestion
@pytest.mark.parametrize(['time_weighting', 'decay'], [('fast', -34.7), ('slow', -4.3)])
def test_time_weighted_levels_decay(time_weighting, decay):
```

f-brinkmann (changes requested):
Thanks for the effort. I mainly had structural comments and comments on the docstring. Also commented on the open points in the pull request description.

pyfar/dsp/levels.py
jf-som: - will add the reference
- I could add the formula, though since it's long and very much not self-explanatory, I don't think it's helpful to anyone who doesn't know the norm anyway.
- The idea was to have a separate convenience function combining this function with the frequency weighting and other options, as described here #762 . There we would also add the L_eq option and perhaps L_E, which are not part of the time weighting from the norm per se. So I would suggest we keep this function here to just this section of the norm and then do all the combinations with other utility elsewhere.

pyfar/dsp/levels.py
jf-som: The norm defines the two types as "F" and "S" and adds a side note that they stand for fast and slow. So the official names are "F" and "S", though it was my idea to allow the full names too for ease of use.

mberz:
Hey @jf-som, I'm currently going through old Issues and stale PRs.
Is this still active, or has the feature been implemented as part of a different PR?

jf-som:
Hi @mberz . This wasn't merged anywhere yet, all we have so far is the work in progress branch `feature/time_weighting`, I just didn't find the time since then to work on it.
I still plan to implement the remainder of the levels module ( #675 ), including the time weighting here, at some point though.

mberz:
Thanks for the update. I've removed the milestone on this until you continue working on the PR.

jf-som:
This is stale and since then I made a new branch `feature/time_weighting` that puts improved functions into the level module instead, as decided in #675

PR #765, opened 2025-02-04
Add filter state setter

### LINKED ISSUE #757: Filterclasses confusing state parameter
```
import pyfar as pf
filter_fir = pf.FilterFIR(coefficients=1, sampling_rate=100, state=0)
filter_fir.init_state(cshape=(1,), state='zeros')
```

In this example, the word "state" has two different meanings. First, it refers to the current state of the filter (a numpy array), and second, it refers to a configuration. This can be a bit confusing. Usually, filter objects are created using the `pyfar.dsp.filter` module, so users rarely interact with them directly.

this came up in #756

hoyer-a:
Add setter for filter state property

Closes #757

### DIFF
--- a/pyfar/classes/filter.py
+++ b/pyfar/classes/filter.py
@@ -200,16 +200,7 @@ class Filter(object):
 
     @property
     def state(self):
-        """
-        Get or set the filter state.
-
-        The filter state sets the initial condition of the filter and can for
-        example be used for block-wise filtering. The required shape is
-        described in the Class docstring.
-
-        Note that the state can also be initialized with the
-        ``init_state`` method.
-        """
+        """Get or set the filter state."""
         return self._state
 
     @state.setter
@@ -396,7 +387,24 @@ class FilterFIR(Filter):
         coeff = np.stack((b, a), axis=-2)
         self._coefficients = _atleast_3d_first_dim(coeff)
 
-    @Filter.state.setter
+    @property
+    def state(self):
+        """
+        Get or set the filter state.
+
+        The filter state sets the initial condition of the filter and can for
+        example be used for block-wise filtering.
+
+        Note that the state can also be initialized with the
+        ``init_state`` method.
+
+        Shape of the state for ``FilterFIR``, with ``cshape``
+        being the channel shape of the :py:class:`~pyfar.Signal` to be
+        filtered: ``(n_filter_chan, *cshape, order)``
+        """
+        return self._state
+
+    @state.setter
     def state(self, state):
         """
         Set initial state of FIR filter.
@@ -418,6 +426,7 @@ class FilterFIR(Filter):
                     "The state does not match the filter structure. Required "
                     "shape for FilterFIR is (n_channels, *cshape, order).")
 
+        # sets filter state in parent class
         Filter.state.fset(self, state)
 
     def init_state(self, cshape, state='zeros'):
@@ -506,7 +515,24 @@ class FilterIIR(Filter):
         """The order of the filter."""
         return np.max(self._coefficients.shape[-2:]) - 1
 
-    @Filter.state.setter
+    @property
+    def state(self):
+        """
+        Get or set the filter state.
+
+        The filter state sets the initial condition of the filter and can for
+        example be used for block-wise filtering.
+
+        Note that the state can also be initialized with the
+        ``init_state`` method.
+
+        Shape of the state for ``FilterIIR``, with ``cshape``
+        being the channel shape of the :py:class:`~pyfar.Signal` to be
+        filtered: ``(n_filter_chan, *cshape, order)``
+        """
+        return self._state
+
+    @state.setter
     def state(self, state):
         """
         Set initial state of IIR filter.
@@ -528,6 +554,7 @@ class FilterIIR(Filter):
                     "The state does not match the filter structure. Required "
                     "shape for FilterIIR is (n_channels, *cshape, order).")
 
+        # sets filter state in parent class
         Filter.state.fset(self, state)
 
     def init_state(self, cshape, state='zeros'):
@@ -634,7 +661,24 @@ class FilterSOS(Filter):
         """The number of sections."""
         return self._coefficients.shape[-2]
 
-    @Filter.state.setter
+    @property
+    def state(self):
+        """
+        Get or set the filter state.
+
+        The filter state sets the initial condition of the filter and can for
+        example be used for block-wise filtering.
+
+        Note that the state can also be initialized with the
+        ``init_state`` method.
+
+        Shape of the state for ``FilterSOS``, with ``cshape``
+        being the channel shape of the :py:class:`~pyfar.Signal` to be
+        filtered: ``n_filter_chan, *cshape, n_sections, 2)``
+        """
+        return self._state
+
+    @state.setter
     def state(self, state):
         """
         Set initial state of SOS filter.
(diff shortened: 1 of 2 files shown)


f-brinkmann (changes requested):
Thanks for starting this. If this is a public method, I think we should check `state.shape` which depends on `coefficients.shape`. One way to do that would be to have a setter in each Filter class that checks the shape and then calls the setter of the parent class. We should also document what shape is required and how the state is defined. You can see the required shapes from the below and I assume the definition of the state is given in scipy

```python
import pyfar as pf
import numpy as np

fs = 44100
fir = pf.FilterFIR([1, .5, .25, .125], fs)
iir = pf.FilterIIR([[1, .5, .25, .125], [1, .5, .25, .125]], fs)
sos = pf.FilterSOS([[[1, .5, .25, 1, .5, .25], [1, .5, .25, 1, .5, .25]]], fs)

for fo in [fir, iir, sos]:
    fo.init_state((1, ), 'zeros')
    fo.process(pf.signals.noise(1000))
    print(fo.state.shape)
```

mberz (changes requested):
Indeed, the setter requires checks matching the state and the coefficient shapes.
Potentially, functionality can be shared with the process method, which requires shape matching with dimensions of the input signal as well.

The shape of the state agrees with the scipy docs, see also for example the filters init_state method.

hoyer-a:
Would be better to have filter-specific documentation for `state`, but that requires a `state` property in the child classes just for the docs. Information about the correct shape is in the parameter docs in `__init__`, maybe that's enough.

And testing the shape unfortunately happens in 2 steps right now:
1. checking for correct shape in regard to the coefficients' shape in `state.setter`
2.  checking for correct cshape in `process`

pyfar/classes/filter.py
```diff
@@ -209,11 +201,30 @@ def n_channels(self):
...
-        The current state of the filter as an array with dimensions
-        corresponding to the order of the filter and number of filter channels.
+        The state of the filter as an array with dimensions
+        corresponding to filter structure and the cshape of the
+        signal to be filtered.
+
+        The state sets the initial conditions for the filter delays and can
+        be used to change the filter's transient.
+
+        Details about the required shape are contained in the class docstring.
```
f-brinkmann, line 211:
```suggestion
        Get or set the filter state.
        
        The filter state sets the initial condition of the filter and can for example be used for block-wise filtering. The required shape is described in the Class docstring. Note that the state can also be initialized with the ``init_state`` method.
```

f-brinkmann (changes requested):
I'm fine with the minimal documentation but would suggest a different docstring (maybe wait for comments from others before chaning this...)

hoyer-a:
At the moment we allow setting the state via init with `ndims < 3` (4 dims in SOS), which then gets expanded using `_atleast_3d_first_dim`.
For now, a warning will be thrown if the state has incorrect shape. In future versions, the state should be required to have the correct shape, and incorrect shapes will no longer be allowed.

example:
```
coeff = np.array([1, 1/2, 0, 0, 0, 0])
filt = pf.FilterFIR(coeff, 44.1e3, state=[0, 0, 0, 0, 0])

filt.state.shape
(1, 1, 5)
```

f-brinkmann (approved):

pyfar/classes/filter.py
```diff
@@ -378,6 +396,30 @@ def coefficients(self, value):
...
+        if state is not None:
+            state = np.asarray(state)
+            if (state.shape[-1] != self.order
+                or state.ndim < 3
+                or state.shape[0] != self.n_channels):
+                raise ValueError(
+                    "The state does not match the filter structure. Required "
+                    "shape for FilterFIR is (n_channels, *cshape, order).")
+
+        Filter.state.fset(self, state)
```
ahms5, line 421: we could add a comment what this fset is, since its not very commonly used in pyfar. In other words, i have never seen this, and I also not sure what is does.

ahms5 (changes requested):
looks fine, jsut small doc stuff.

pyfar/classes/filter.py
hoyer-a: `fset` sets the value of a property, this is called internally everytime a property value is set.
Here we do it to extend the the parent classes setter functionality.

pyfar/classes/filter.py
```diff
@@ -396,7 +387,24 @@ def coefficients(self, value):
...
+    @property
+    def state(self):
+        """
+        Get or set the filter state.
+
+        The filter state sets the initial condition of the filter and can for
+        example be used for block-wise filtering.
+
+        Note that the state can also be initialized with the
+        ``init_state`` method.
```
ahms5, line 399:
```suggestion
        Note that the state can also be initialized with the
        :py:func:`~pyfar.FilterFIR.init_state` method.
```

pyfar/classes/filter.py
```diff
@@ -396,7 +387,24 @@ def coefficients(self, value):
...
+
+        Shape of the state for ``FilterFIR``, with ``cshape``
+        being the channel shape of the :py:class:`~pyfar.Signal` to be
+        filtered: ``(n_filter_chan, *cshape, order)``
```
ahms5, line 403:
```suggestion
        Shape of the state must be  ``(n_filter_channels, *cshape, order)``,
        with ``cshape`` being the channel shape of the :py:class:`~pyfar.Signal` to be filtered.
```

pyfar/classes/filter.py
```diff
@@ -506,7 +515,24 @@ def order(self):
...
+    @property
+    def state(self):
+        """
+        Get or set the filter state.
+
+        The filter state sets the initial condition of the filter and can for
+        example be used for block-wise filtering.
+
+        Note that the state can also be initialized with the
+        ``init_state`` method.
```
ahms5, line 527:
```suggestion
        Note that the state can also be initialized with the
        :py:func:`~pyfar.FilterIIR.init_state` method.
```

pyfar/classes/filter.py
```diff
@@ -634,7 +661,24 @@ def n_sections(self):
...
```
ahms5, line 673:
```suggestion
        Note that the state can also be initialized with the
        :py:func:`~pyfar.FilterSOS.init_state` method.
```

pyfar/classes/filter.py
```diff
@@ -506,7 +515,24 @@ def order(self):
...
+
+        Shape of the state for ``FilterIIR``, with ``cshape``
+        being the channel shape of the :py:class:`~pyfar.Signal` to be
+        filtered: ``(n_filter_chan, *cshape, order)``
```
ahms5, line 531:
```suggestion
        Shape of the state must be  ``(n_filter_channels, *cshape, order)``,
        with ``cshape`` being the channel shape of the :py:class:`~pyfar.Signal` to be filtered.
```

pyfar/classes/filter.py
```diff
@@ -634,7 +661,24 @@ def n_sections(self):
...
+
+        The filter state sets the initial condition of the filter and can for
+        example be used for block-wise filtering.
+
+        Note that the state can also be initialized with the
+        ``init_state`` method.
+
+        Shape of the state for ``FilterSOS``, with ``cshape``
+        being the channel shape of the :py:class:`~pyfar.Signal` to be
+        filtered: ``n_filter_chan, *cshape, n_sections, 2)``
```
ahms5, line 677:
```suggestion
        Shape of the state must be  ``(n_filter_channels, *cshape, n_sections, 2))``,
        with ``cshape`` being the channel shape of the :py:class:`~pyfar.Signal` to be filtered.
```

ahms5 (changes requested):
almost done, thank you for taking care.

ahms5 (approved):

PR #767, opened 2025-02-07
HOT: replace version label check by check for any label

ahms5:
if we rmove the version labels all tests are failing, so my idea was to replace it by a check for any label (eg related by topic or type).

an alternative apporach would be to just remove the test

### DIFF
--- a/.github/workflows/has_label.yml
+++ b/.github/workflows/has_label.yml
@@ -1,4 +1,4 @@
-name: PR Check for any Label
+name: pull_request: label
 
 on:
   pull_request:
--- a/.github/workflows/has_label.yml
+++ b/.github/workflows/has_label.yml
@@ -19,10 +19,8 @@ jobs:
           LABEL_NAMES=$(echo "$PR_DATA" | jq -r '.labels[].name')
           echo "Labels: $LABEL_NAMES"
 
-          REGEX="^v[0-9]+\.[0-9]+\.[0-9]+$"
-          MATCHES=$(echo "$LABEL_NAMES" | grep -E "$REGEX")
-          if [ -z "$MATCHES" ]; then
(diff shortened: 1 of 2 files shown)


mberz:
Sorry about that. 💩
I wanted to check if the label can simply be replaced by a milestone.

ahms5:
I would also recomment to replace the labels by milestones

sikersten:
Would it be possible to replace it with a check for the milestone?

ahms5:
> Would it be possible to replace the test with a test for the milestone?

done

mberz (approved):

.github/workflows/has_label.yml
```diff
@@ -1,4 +1,4 @@
-name: PR Check Label for version Label
+name: PR Check for any Label
```
sikersten, line 1:
```suggestion
name: PR: label
```

.github/workflows/has_version_milestone.yml
```diff
@@ -0,0 +1,29 @@
...
+      - opened
+      - labeled
+      - unlabeled
+      - synchronize
+
+jobs:
+  check-labels:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check for version milestone
```
sikersten, line 15:
```suggestion
      - name: PR: version milestone
```

sikersten (approved):
Thanks for taking care! Suggested only some cosmetics

f-brinkmann (approved):

.github/workflows/has_label.yml
```diff
@@ -12,17 +12,15 @@ jobs:
   check-labels:
     runs-on: ubuntu-latest
     steps:
-      - name: Check for version label
+      - name: Check for label
```
ahms5, line 15:
```suggestion
      - name: pull_request: label
```

PR #768, opened 2025-02-09
Impulse response length estimation for Filter objects

### LINKED ISSUE #766: Missing warning for too short IIR filter impulse responses
The length of the impulse response of IIR filters is not checked, so no warning is returned if the returned IR is too short. 
This feature is already implemented for FIR filters.

We should implement this for IIR filters as well to harmonize the behavior accross filter classes.
An example implemented in MATLAB can be found here https://nl.mathworks.com/help/signal/ref/impzlength.html#d126e117775 

See also #755

f-brinkmann:

Closes #766 

- Add `impulse_response_length` method to all class functions. This is a port of [impzlength](https://de.mathworks.com/help/signal/ref/impzlength.html). These methods are tested against values computed with `impzlength`.
- Use new method to set default impulse response length for all filter classes
- Raise a warning if user values for impulse response length are smaller than the estimated length

### DIFF
--- a/pyfar/classes/filter.py
+++ b/pyfar/classes/filter.py
@@ -603,6 +603,8 @@ class FilterIIR(Filter):
             An integer array of shape (C, ) containing the length in samples
             where `C` denotes the number of channels of the filter.
         """
+        # This is a direct python port of the parts of Matlab's impzlength that
+        # refer to IIR filters
 
         channels = self.coefficients.shape[0]
         estimated_length = np.zeros(channels)
@@ -821,6 +823,8 @@ class FilterSOS(Filter):
             An integer array of shape (C, ) containing the length in samples
             where `C` denotes the number of channels of the filter.
         """
+        # This is a direct python port of the parts of Matlab's impzlength that
+        # refer to SOS filters
 
         channels = self.coefficients.shape[0]
         sections = self.coefficients.shape[1]
--- a/pyfar/classes/filter.py
+++ b/pyfar/classes/filter.py
@@ -292,8 +292,10 @@ class Filter(object):
         Returns
         -------
         impulse_response : Signal
-            The impulse response of the filter with a ``cshape = (C, )`` where
-            `C` denotes the number of channels of the filter.
+            The impulse response of the filter of with a
+            ``cshape = (n_channels, )``, with the channel shape
+            :py:func:`~pyfar.Signal.cshape` and the number of filter channels
+            :py:func:`~n_channels`.
         """
         # set or check the impulse response length
         minimum_impulse_response_length = int(np.max(
@@ -461,8 +463,8 @@ class FilterFIR(Filter):
         Returns
         -------
         minimum_impulse_response_length : array
-            An integer array of shape (C, ) containing the length in samples
-            where `C` denotes the number of channels of the filter.
+            An integer array of shape(n_channels, ) containing the length in
+            samples with the number of filter channels :py:func:`~n_channels`.
         """
 
         # get filter coefficients
@@ -501,8 +503,10 @@ class FilterFIR(Filter):
         Returns
         -------
         impulse_response : Signal
-            The impulse response of the filter of with a ``cshape = (C, )``
-            where `C` denotes the number of channels of the filter.
+            The impulse response of the filter of with a
+            ``cshape = (n_channels, )``, with the channel shape
+            :py:func:`~pyfar.Signal.cshape` and the number of filter channels
+            :py:func:`~n_channels`.
         """
         return super().impulse_response(n_samples)
 
@@ -597,8 +601,10 @@ class FilterIIR(Filter):
         Returns
         -------
         impulse_response : Signal
-            The impulse response of the filter of with a ``cshape = (C, )``
-            where `C` denotes the number of channels of the filter.
+            The impulse response of the filter of with a
+            ``cshape = (n_channels, )``, with the channel shape
+            :py:func:`~pyfar.Signal.cshape` and the number of filter channels
+            :py:func:`~n_channels`.
         """
         return super().impulse_response(n_samples)
 
@@ -622,8 +628,8 @@ class FilterIIR(Filter):
         Returns
         -------
         minimum_impulse_response_length : array
-            An integer array of shape (C, ) containing the length in samples
-            where `C` denotes the number of channels of the filter.
+            An integer array of shape(n_channels, ) containing the length in
+            samples with the number of filter channels :py:func:`~n_channels`.
         """
         # This is a direct python port of the parts of Matlab's impzlength that
         # refer to IIR filters
@@ -835,8 +841,10 @@ class FilterSOS(Filter):
         Returns
         -------
         impulse_response : Signal
-            The impulse response of the filter with a ``cshape = (C, )`` where
-            `C` denotes the number of channels of the filter.
+            The impulse response of the filter of with a
+            ``cshape = (n_channels, )``, with the channel shape
+            :py:func:`~pyfar.Signal.cshape` and the number of filter channels
+            :py:func:`~n_channels`.
         """
         return super().impulse_response(n_samples)
 
@@ -860,8 +868,8 @@ class FilterSOS(Filter):
         Returns
         -------
         minimum_impulse_response_length : array
-            An integer array of shape (C, ) containing the length in samples
-            where `C` denotes the number of channels of the filter.
+            An integer array of shape(n_channels, ) containing the length in
+            samples with the number of filter channels :py:func:`~n_channels`.
         """
         # This is a direct python port of the parts of Matlab's impzlength that
         # refer to SOS filters
--- a/pyfar/classes/filter.py
+++ b/pyfar/classes/filter.py
@@ -457,21 +457,14 @@ class FilterFIR(Filter):
         Parameters
         ----------
         tolerance : float, optional
-            A tolerance to estimate the length of noise FIR filters.
-
-            - ``None`` returns the full length by finding the last non-zero
-               coefficient per channel.
-            - ``float`` estimates the length by finding the last coefficient
-              with an absolute value greater or equal to the absolute maximum
-              of the coefficients per channel multiplied by `tolerance`. For
-              example if ``tolerance = 0.001`` trailing values below
-              :math:`20 \log_{10}(0.001)=-60` dB would be ignored in the length
-              estimation.
-
-            The returned lengths using ``tolerance=None`` will always be larger
-            or equal to the returned lengths if `tolerance` is a float value.
-            The default is ``None``.
-
+            Tolerance for estimating the minimum length of noisy FIR filters.
+            The length is estimated by finding the last coefficient with an
+            absolute value greater or equal to the absolute maximum of the
+            filter coefficients per channel multiplied by `tolerance`. For
+            example if ``tolerance = 0.001`` trailing values below
+            :math:`20 \log_{10}(0.001)=-60` dB would be ignored in the length
+            estimation. The default ``None`` uses the numerical precision
+            ``2 * numpy.finfo(float).resolution`` as a strict threshold.
         unit : string, optional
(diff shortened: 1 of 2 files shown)


pyfar/classes/filter.py
```diff
@@ -537,27 +562,124 @@ def init_state(self, cshape, state='zeros'):
...
+
+            if np.any(np.abs(poles) > 1.0001):
+                # This is an unstable filter. The closer the outmost pole is to
+                # the unit circle, the longer the impulse response.
+                estimated_length[channel] += \
+                    6 / np.log10(np.max(np.abs(poles[np.abs(poles) > 1])))
+            else:
+                # This is a stable filter. Minimum height is 1e-5 of original
+                # amplitude
+                idx = np.abs(poles - 1) < 1e-5
```
narahahn, line 634: I think `np.abs(poles-1)` gives the distance between the poles and (1, 0) on the real axis.
I guess it is meant to be `np.abs(np.abs(poles)-1)` similar to lines 638 and 640 below?

pyfar/classes/filter.py
```diff
@@ -537,27 +562,124 @@ def init_state(self, cshape, state='zeros'):
...
+                poles[idx] = -poles[idx]
```
narahahn, line 635: Why are those poles reflected wrt to the origin?
It this supposed to bring the unstable poles into the unit circle? If yes, it can be done like this:
`poles[idx] = 1/np.abs(poles[idx]) * np.exp(1j*np.angle(poles[idx]))`

pyfar/classes/filter.py
```diff
@@ -292,6 +292,14 @@ def impulse_response(self, n_samples):
         impulse_response : Signal
             The impulse response of the filter.
         """
+        # set or check the impulse response length
+        impulse_response_length = int(np.max(self.impulse_response_length()))
+        if n_samples is None:
+            n_samples = impulse_response_length
```
ahms5, line 298: seems like n_samples has a defualt, but its not mentioned in the docs

pyfar/classes/filter.py
```diff
@@ -434,33 +442,50 @@ def init_state(self, cshape, state='zeros'):
                 new_state[idx, ...] = spsignal.lfilter_zi(coeff[0], coeff[1])
         super().init_state(state=new_state)
 
+    def impulse_response_length(self):
```
ahms5, line 445: ir length sounds like seconds, but not like samples, maybe we can rename to `impulse_response_samples` or similar

pyfar/classes/filter.py
```diff
@@ -434,33 +442,50 @@ def init_state(self, cshape, state='zeros'):
...
+        """
+        Get the length of filter impulse response in samples.
+
+        The length is computed from the last non-zero coefficient per channel.
+
+        Returns
+        -------
+        impulse_response_length : array
+            An integer array of shape (C, ) containing the length in samples
+            where `C` denotes the number of channels of the filter.
```
ahms5, line 455: is it intended, that we dont use the ``cshape``concept here?

pyfar/classes/filter.py
```diff
@@ -434,33 +442,50 @@ def init_state(self, cshape, state='zeros'):
...
+        # find last non-zero entry per channel
+        estimated_length = np.apply_along_axis(
+            lambda b: np.max(np.nonzero(b)), axis=-1, arr=b) + 1
+
+        # restore input data shape
+        estimated_length = np.reshape(estimated_length, b.shape[:-1])
+
+        return estimated_length.astype(int)
+
     def impulse_response(self, n_samples=None):
```
ahms5, line 470: do we need to overwirte it? its the same as Filter.impulse_response() and impulse_response_length is implemented for each class.

ahms5 (changes requested):
thank you! I was wondering if we can just enhired the `impulse_response` method from Filter, because we shifted the individual behavior to `impulse_resposne_length`

pyfar/classes/filter.py
f-brinkmann: Yes we need to and decided to do so in the weekly. The reason is that the calculation of the impulse response length is different for each filter, hence we need individual docs for each filter class.

pyfar/classes/filter.py
f-brinkmann: True, I added it to the docs. Note that this method is overwritten in each derived class and that this docstring is not shown in the docs for this reason. But still better to have a clean internal documentation :)

pyfar/classes/filter.py
f-brinkmann: I would be fine with changing the name. Lets maybe find the final name in the weekly. Otherwise it might be a ping pong between different names...

pyfar/classes/filter.py
f-brinkmann: According to the docstrings, filters can not have multidimensional channel shapes. Hence, I did not use cshape here.

f-brinkmann:
> thank you! I was wondering if we can just enhired the `impulse_response` method from Filter, because we shifted the individual behavior to `impulse_resposne_length`

We wanted individual docs for each method, because the estimation of the impulse response length differs between filter types.

ahms5:
Open discussion:
- [ ] better name for ``impulse_response_length``
- [ ] does ``impulse_response`` need individual docstring, or is it ok to just refer to ``impulse_response_length``

pyfar/classes/filter.py
mberz: `impulse_response_samples` to me sounds like the value of the samples, not the length of the IR in samples.
samples is only the unit used to quantify the length. So length would be fitting better in my opinion.

f-brinkmann: We decided to call this `minimum_impulse_response_length`

pyfar/classes/filter.py
```diff
@@ -284,14 +284,25 @@ def impulse_response(self, n_samples):
 
         Parameters
         ----------
-        n_samples : int
-            Length in samples for which the impulse response is computed.
+        n_samples : int, None
+            Length in samples for which the impulse response is computed. If
+            this is ``None``the length is estimated using
+            `minimum_impulse_response_length((`.
```
ahms5, line 290:
```suggestion
            this is ``None`` the length is estimated using
            `minimum_impulse_response_length`.
```

pyfar/classes/filter.py
```diff
@@ -284,14 +284,25 @@ def impulse_response(self, n_samples):
...
 
         Returns
         -------
         impulse_response : Signal
             The impulse response of the filter.
```
ahms5, line 295:
```suggestion
            The impulse response of the filter of with a ``cshape = (C, )``
            where `C` denotes the number of channels of the filter.
```

ahms5 (changes requested):
Thanks, jsut small comments

sikersten (commented):
Thanks for the effort! Only minor comments

pyfar/classes/filter.py
f-brinkmann: I went for Annes suggestion. n_channels would be more verbose but would still require an explicit definition and I though '``cshape = (n_channeles, )`` where `n_channels` denotes the number of channels of the filter.' would be a little lengthy.

f-brinkmann: I changed this for all `impulse_response` functions.

pyfar/classes/filter.py
```diff
@@ -537,27 +579,143 @@ def init_state(self, cshape, state='zeros'):
...
+                # the unit circle, the longer the impulse response.
+                estimated_length[channel] += \
+                    6 / np.log10(np.max(np.abs(poles[np.abs(poles) > 1])))
+            else:
+                # This is a stable filter. Minimum height is 1e-5 of original
+                # amplitude
+                idx = np.abs(poles - 1) < 1e-5
+                poles[idx] = -poles[idx]
+
+                # poles on and close to unit circls
```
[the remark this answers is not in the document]
f-brinkmann: Thanks for catching

pyfar/classes/filter.py
```diff
@@ -434,33 +445,64 @@ def init_state(self, cshape, state='zeros'):
                 new_state[idx, ...] = spsignal.lfilter_zi(coeff[0], coeff[1])
         super().init_state(state=new_state)
 
+    def minimum_impulse_response_length(self, unit='samples'):
+        """
+        Get the minimum length of filter impulse response.
```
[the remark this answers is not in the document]
f-brinkmann: Yes, thanks. I changed this for all minimum_impulse_response_length functions.

pyfar/classes/filter.py
```diff
@@ -434,33 +445,64 @@ def init_state(self, cshape, state='zeros'):
...
+        ----------
+        unit : string, optional
+            The unit in which the length is returned. Can be ``'samples'`` or
+            ``'s'`` (seconds). The default is ``'samples'``.
+
+        Returns
+        -------
+        minimum_impulse_response_length : array
+            An integer array of shape (C, ) containing the length in samples
+            where `C` denotes the number of channels of the filter.
```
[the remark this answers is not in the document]
f-brinkmann: See comment above.

pyfar/classes/filter.py
```diff
@@ -434,33 +445,64 @@ def init_state(self, cshape, state='zeros'):
...
+            :py:func:`~FilterFIR.minimum_impulse_response_length`.
+            A warning is returned if ``n_samples`` is shorter than the value
+            returned by :py:func:`~FilterFIR.minimum_impulse_response_length`.
 
         Returns
         -------
         impulse_response : Signal
-            The impulse response of the filter.
+            The impulse response of the filter of with a ``cshape = (C, )``
+            where `C` denotes the number of channels of the filter.
```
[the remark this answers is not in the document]
f-brinkmann: See comment above.

pyfar/classes/filter.py
```diff
@@ -537,27 +579,143 @@ def init_state(self, cshape, state='zeros'):
...
+            :py:func:`~FilterIIR.minimum_impulse_response_length`.
+            A warning is returned if ``n_samples`` is shorter than the value
+            returned by :py:func:`~FilterIIR.minimum_impulse_response_length`.
 
         Returns
         -------
         impulse_response : Signal
-            The impulse response of the filter.
+            The impulse response of the filter of with a ``cshape = (C, )``
+            where `C` denotes the number of channels of the filter.
```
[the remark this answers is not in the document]
f-brinkmann: See comment above.

pyfar/classes/filter.py
```diff
@@ -675,6 +838,76 @@ def impulse_response(self, n_samples):
...
+            ``'s'`` (seconds). The default is ``'samples'``.
+        tolerance : float, optional
+            Tolerance for the accuracy. Smaller tolerances will result in
+            larger impulse response lengths. The default is ``5e-5``.
+
+        Returns
+        -------
+        minimum_impulse_response_length : array
+            An integer array of shape (C, ) containing the length in samples
+            where `C` denotes the number of channels of the filter.
```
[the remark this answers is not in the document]
f-brinkmann: See comment above.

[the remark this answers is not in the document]
mberz: Simon is right here: Filters do have the property n_channels

mberz (changes requested):

pyfar/classes/filter.py
```diff
@@ -446,35 +448,55 @@ def init_state(self, cshape, state='zeros'):
                 new_state[idx, ...] = spsignal.lfilter_zi(coeff[0], coeff[1])
         super().init_state(state=new_state)
 
-    def minimum_impulse_response_length(self, unit='samples'):
-        """
+    def minimum_impulse_response_length(self, unit='samples', tolerance=None):
```
mberz, line 451: we could also think about setting this to something like 2*eps, so numerical errors are always accounted for.

f-brinkmann: even better!

PR #774, opened 2025-02-24
Feature: add density of air

### LINKED ISSUE #428: We need to implement calculations of commonly used constants
We need an implementation to calculate and store typically used constants, such as air density, speed of sound, etc.
Ideally, this could be implemented as a Dataclass or similar.

ahms5:

- adds density of air calculation based on the ideal gas method 
-
-
closes #428 since its the last step


f-brinkmann (changes requested):
Only small comments. I could not check the formulas, because I dont have the book. Everything else looks fine.

pyfar/constants/constants.py
```diff
@@ -42,3 +43,86 @@ def saturation_vapor_pressure_magnus(temperature):
...
+    R_a = R / mu_a
+
+    # next to Equation 6.69
+    alpha = mu_a / mu_w
+
+    # partial pressure of water vapor in Pa
+    if saturation_vapor_pressure is None:
+        p = saturation_vapor_pressure_magnus(temperature)  # Pa
+    else:
+        p = np.array(saturation_vapor_pressure, dtype=float)  # Pa
```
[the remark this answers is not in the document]
ahms5: since all input parameters are converted into arrays, the broadcast should happen during the calculation automatically. We could add a chek if all are broadcastable, to catch an possible error here, and make it easier for the user.
```
try:
    _ = np.broadcast_shapes(
        temperature, relative_humidity, atmospheric_pressure, 
        saturation_vapor_pressure)
catch:
    raise ValueError('not broadcastable bla')
```
I think there is no need to broadcast them before, at least in this case. #787 might be useful for other cases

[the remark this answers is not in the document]
f-brinkmann: Yes agree, the idea was to have a more verbose error and #787 can do with or witout broadcasting.

ahms5:
Open discussion:
see last comment from @f-brinkmann, this might also be a general discussion and not just specific for this pr.

f-brinkmann (approved):

pyfar/constants/constants.py
```diff
@@ -353,3 +353,84 @@ def saturation_vapor_pressure_magnus(temperature):
...
+
+    The density of air is calculated based on chapter 6.3 in [#]_.
+    All input parameters must be broadcastable to the same shape.
+
+    Parameters
+    ----------
+    temperature : float, array_like
+        Temperature in degrees Celsius (°C).
+    relative_humidity : float, array_like
+        Relative humidity in the range of 0 to 1.
```
sikersten, line 373:
```suggestion
        Relative humidity in the range from 0 to 1.
```

pyfar/constants/constants.py
```diff
@@ -353,3 +353,84 @@ def saturation_vapor_pressure_magnus(temperature):
...
+    atmospheric_pressure : float, array_like, optional
+        Atmospheric pressure in pascal (Pa), by default
```
sikersten, line 375:
```suggestion
        Atmospheric pressure in Pascal (Pa), by default
```

pyfar/constants/constants.py
```diff
@@ -353,3 +353,84 @@ def saturation_vapor_pressure_magnus(temperature):
...
+        :py:attr:`reference_atmospheric_pressure`.
+    saturation_vapor_pressure : float, array_like, optional
+        Saturation vapor pressure in Pa.
```
sikersten, line 378:
```suggestion
        Saturation vapor pressure in Pascal (Pa).
```

pyfar/constants/constants.py
```diff
@@ -353,3 +353,84 @@ def saturation_vapor_pressure_magnus(temperature):
...
+    relative_humidity = np.array(relative_humidity, dtype=float)
+
+    # Constants according to 6.3.2
+    R = 8.314  # J/(K mol)
+    mu_a = 28.97*1e-3  # kg/mol
+    mu_w = 18.02*1e-3  # kg/mol
+    R_a = R / mu_a
+
+    # next to Equation 6.69
+    alpha = mu_a / mu_w
```
sikersten, line 423:
```suggestion

    # next to Equation 6.69
    R_a = R / mu_a
    alpha = mu_a / mu_w
```

sikersten (changes requested):
Thanks for implementing! Only minor things, otherwise approved.

sikersten (approved):

PR #778, opened 2025-03-02
Disable slicing for x, y and z

### LINKED ISSUE #550: array slicing bug? for non cartesian properties.
## Description
It seems to have sth to do with the conversion to cartesian coordinates.
For some reason we don't jump into the setter method.
Not sure if we can fix it easily.

## What I Did

```
import pyfar as pf
import numpy as np
c = pf.Coordinates.from_spherical_elevation([np.pi, np.pi/2], 0, 1)
c.azimuth[0] = 0
c.azimuth
```
array([3.14159265, 1.57079633])

```
import pyfar as pf
import numpy as np
c = pf.Coordinates([2, 3], 1, 1)
c.cartesian[0, 0] = 0
c.cartesian
```
array([[2., 1., 1.],
          [3., 1., 1.]])

for x,y,z it works
```
import pyfar as pf
import numpy as np
c = pf.Coordinates([1, 2, 3], 0, 0)
c.x[0] = 0
c.x
```
array([0., 2., 3.])
This just works because we return the direct reference to _x here, which also might be very dangous.

hoyer-a:
Disable slicing for all coordinates by returning copies of the internal attributes.

Closes #550 

Not sure if the note in the docs is clear and understandable, feel free to make other suggestions.


pyfar/classes/coordinates.py
```diff
@@ -41,6 +41,10 @@
 systems (`spherical_colatitude` and `spherical_elevation`). The table below
 lists all coordinates.
 
+.. note::
+    All coordinates are returned as copies of the internal data. This means
+    that changing values within slices (e.g. ``coordinates.x[0] = 0``)
+    has no effect on the :py:class:`~pyfar.Coordinates` object's data.
```
f-brinkmann, line 47: Maybe it is clearer if we give an example of how to do this:
```suggestion
    that for example ``coordinates.x[0] = 0`` does not change
    ``coordinates.x``. This can be done using
    
    .. code-block:: python

       new_x = coordinates.x
       new_x[0] = 0
       coordinates.x = new_x
```

f-brinkmann (changes requested):
Thanks for the fix. I suggested a more verbose note that also gives an example of how to change coordinates to make it more clear.

f-brinkmann (dismissed):

f-brinkmann:
If we get a second review, by Friday, we could cherrypick and merge into main for v0.7.2

pyfar/classes/coordinates.py
```diff
@@ -41,6 +41,16 @@
...
+.. note::
+    All coordinates are returned as copies of the internal data. This means
+    that for example ``coordinates.x[0] = 0`` does not change
+    ``coordinates.x``. This can be done using
+
+    .. code-block:: python
+
+       new_x = coordinates.x
+       new_x[0] = 0
+       coordinates.x = new_x
```
ahms5, line 53:
```suggestion
       coordinates.x = new_x

```

ahms5 (changes requested):
Thank you for handling this! I believe this is a bug fix, so it should definitely go to the main branch. To make it more complete, we could add a test to verify that the coordinates remain unchanged when, for example, the ``x`` parameter is modified.

ahms5 (approved):

ahms5:
> If we get a second review, by Friday, we could cherrypick and merge into main for v0.7.2

If we merge it over the weekend, especially during Karneval, it feels like when the Bundestag passes unpopular laws during a soccer championship. 😂

f-brinkmann (approved):

PR #782, opened 2025-03-31
Feature: dirac sequence generator for room acoustics

### LINKED ISSUE #781: dirac sequence generator for room acoustics
add a dirac sequence generator based on Schröder's apporach for a RIR syntesis.

ahms5:

Closes #781

-
-
-


mberz (changes requested):
Disclaimer: I did not review the code. 
I think this should not be part of pyfar, but pyrato instead. It clearly states roomacoustics in multiple places in the code and docstrings. 
If we start mixing content across multiple packages it becomes unreasonably hard to keep track of what is where.

mberz:
I'm honestly a bit confused what the purpose of this function should be if it's a random Poisson distributed sequence, so it's a kind of noise process with random temporal sampling.

I would be on board with having a time domain dirac comb I would agree that it's suitable here, see https://en.wikipedia.org/wiki/Dirac_comb
I think I may not be getting the point here or the name of the function is not well chosen.

ahms5:
lets discuss on Friday :)

PR #783, opened 2025-04-08
Feature/transmission line

sikersten:
Simple transmission line method for [TransmissionMatrix](https://pyfar.readthedocs.io/en/develop/classes/pyfar.transmission_matrix.html) class.

- added create_transmission_line method to TransmissionMatrix class
- added tests

### DIFF
--- a/pyfar/classes/transmission_matrix.py
+++ b/pyfar/classes/transmission_matrix.py
@@ -786,7 +786,6 @@ class TransmissionMatrix(FrequencyData):
 
             >>> import pyfar as pf
             >>> import numpy as np
-            >>> import matplotlib.pyplot as plt
             >>> # Transmission line parameters
             >>> frequencies = np.linspace(20, 20e3, 1000)
             >>> omega = 2*np.pi*frequencies
--- a/pyfar/classes/transmission_matrix.py
+++ b/pyfar/classes/transmission_matrix.py
@@ -32,6 +32,7 @@ from __future__ import annotations # required for Python <= 3.9
 import numpy as np
 import numpy.testing as npt
 from pyfar.classes.audio import FrequencyData
+import numbers
 
 
 class TransmissionMatrix(FrequencyData):
@@ -745,7 +746,7 @@ class TransmissionMatrix(FrequencyData):
             length: float | FrequencyData,
             characteristic_impedance: complex | FrequencyData,
             ) -> TransmissionMatrix:
-        r"""Creates a transmission matrix representing a transmission line.
+        r"""Create a transmission matrix representing a transmission line.
 
         The transmission matrix is calculated from the wavenumber :math:`k`,
         length :math:`l`, and characteristic impedance :math:`Z_0` following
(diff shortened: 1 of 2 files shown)


pyfar/classes/transmission_matrix.py
```diff
@@ -739,6 +740,108 @@ def create_gyrator(transducer_constant :
         frequencies = transducer_constant.frequencies
         return TransmissionMatrix.from_abcd(0, B, C, 0, frequencies)
 
+    @staticmethod
+    def create_transmission_line(
+            wavenumber: complex | FrequencyData,
+            length: float | FrequencyData,
+            characteristic_impedance: complex | FrequencyData,
```
mberz, line 747: To me the input parametrizations seem a bit convoluted.
Having a FrequencyData object to specify the length does sound like a very niche case to me. Do we really want to support all of these?
If only k and Z are FrequencyData objects the input checking and usablility would be much simpler I guess.

mberz: With your last changes it's already making a lot more sense.

pyfar/classes/transmission_matrix.py
```diff
@@ -739,6 +740,108 @@ def create_gyrator(transducer_constant :
         frequencies = transducer_constant.frequencies
         return TransmissionMatrix.from_abcd(0, B, C, 0, frequencies)
 
+    @staticmethod
+    def create_transmission_line(
+            wavenumber: FrequencyData,
+            length: float | FrequencyData,
+            characteristic_impedance: complex | FrequencyData,
```
mberz, line 747: I'm assuming this can also be a float?
Type hint and docstring are not matching for length and impedance.

pyfar/classes/transmission_matrix.py
mberz: I'm wondering if it wouldn't be much simpler if instead of separately passing k and l, they could be combined into a single dimensionless variable kl

mberz: Mh, I do not have a better name in mind. 
For me it would be okay here, as the equation is given in the docstring, but let's think about better options.

f-brinkmann:
We decided to combine k (wave number) and l (length) into a single parameter kl (wave number * length)

pyfar/classes/transmission_matrix.py
sikersten: I introduced `Number` for the characteristic impedance both in type hint and docstring

pyfar/classes/transmission_matrix.py
```diff
@@ -739,6 +740,93 @@ def create_gyrator(transducer_constant :
...
+        Example
+        -------
+        Lossless acoustic duct with approximate ear canal dimensions.
+        Plot its input impedance with rigid termination.
+
+        .. plot::
+
+            >>> import pyfar as pf
+            >>> import numpy as np
+            >>> import matplotlib.pyplot as plt
```
ahms5, line 789:
```suggestion
```
do we need it?

ahms5 (changes requested):
thanks, nice example, just small changes,

mberz (approved):

ahms5 (approved):

PR #787, opened 2025-04-13
pyfar: _utils broadcasting function

f-brinkmann:

In some cases, we have multipe input arguments that must be converted to numpy arrays and must be broadcastable to the same shape. This is a helper function to take care of this and raise a verbose error, of broadcasting does not work.

### DIFF
--- a/pyfar/_utils.py
+++ b/pyfar/_utils.py
@@ -1,4 +1,5 @@
 from pyfar.classes.warnings import PyfarDeprecationWarning
+import numpy as np
 import warnings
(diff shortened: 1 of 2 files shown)


mberz:
@ahms5 What's the reason this was moved to "On Hold"?

ahms5:
This PR was created for this discussion here https://github.com/pyfar/pyfar/pull/774#discussion_r2041071946. maybe we should discuss how we check for function inputs and if we should provide pyfar helping functions for this.

PR #788, opened 2025-04-16
Fix secondary sidebar display of class attributes

### LINKED ISSUE #785: Readthedocs: long attributes aren't fully displayed in secondary sidebar
Longer attributes are not displayed fully in secondary sidebar.
Example: 

Image

<br>
We could display just the attribute names per class e.g. `T`, `cdim` etc.

Example: 
Image

hoyer-a:
Closes #785 

Adjust display of secondary sidebar using custom js.
Show only class `attribute` in sidebar instead of `class.attribute`.

### DIFF
--- a/docs/_static/custom.js
+++ b/docs/_static/custom.js
@@ -1,5 +1,17 @@
-document.addEventListener("DOMContentLoaded", () => {
-    // Only target spans inside the page ToC
+/**
+ * Cleans up the third-level TOC entries in the secondary sidebar.
+ *
+ * This function is used to simplify the attribute names in secondary sidebar
+ * entries by removing the class name prefix.
+ *
+ * Example:
+ *   "Class.attribute" → "attribute"
+ *
+ * @function cleanTocAttributeEntries
+ * @returns {void}
+ */
+function cleanTocAttributeEntries() {
+    // Only target spans inside the secondary sidebar
     const sidebarToc = document.querySelector(".bd-sidebar-secondary");
     if (!sidebarToc) return;
 
@@ -11,4 +23,9 @@ document.addEventListener("DOMContentLoaded", () => {
         const firstDotIndex = original.indexOf(".");
         span.textContent = original.slice(firstDotIndex + 1); // Keep everything after the first dot
     });
+};
+
+// Run after html document has been parsed
+document.addEventListener("DOMContentLoaded", () => {
+    cleanTocAttributeEntries();
 });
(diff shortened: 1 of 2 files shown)


mberz:
I realized that the width of the right hand sidebar is still quite narrow. Maybe we can also increase the width of that a bit more or is this scaled automatically?

mberz:
Ah, that's too bad. I was hoping it would make the page a bit wider.

docs/_static/custom.js
```diff
@@ -0,0 +1,15 @@
...
+    const sidebarToc = document.querySelector(".bd-sidebar-secondary");
+    if (!sidebarToc) return;
+
+    // Select spans only inside <li class="toc-h3 nav-item toc-entry">
+    const tocSpans = sidebarToc.querySelectorAll("li.toc-h3.nav-item.toc-entry .pre");
+
+    tocSpans.forEach(span => {
+        const original = span.textContent;
+        if (original.includes(".")) {
+            span.textContent = original.split(".").pop(); // Keep only what comes after the last dot
```
f-brinkmann, line 12: Wouldn't it be safer to keep everything after the first dot? Otherwise information for nested classes would not be displayed correctly. E.g., if we had something like 'Signal.Coordinates.elevation' it would only display 'elevation'.

f-brinkmann (changes requested):
Had only a small question. Docs look way better and cleaner than before :)

docs/_static/custom.js
hoyer-a: Yes, definitely! Did not think of that.

hoyer-a:
Not sure why the test failed, had no problems locally. Can someone maybe restart it on circleci?

f-brinkmann:
> Not sure why the test failed, had no problems locally. Can someone maybe restart it on circleci?

I think only @mberz has the power to do that 🙈

mberz:
> > Not sure why the test failed, had no problems locally. Can someone maybe restart it on circleci?
> 
> I think only @mberz has the power to do that 🙈

Theoretically everyone who is part of the pyfar organization should be able to rerun tests. You need to be logged in though.

docs/_static/custom.js
```diff
@@ -0,0 +1,14 @@
...
+
+    // Select spans only inside <li class="toc-h3 nav-item toc-entry">
+    const tocSpans = sidebarToc.querySelectorAll("li.toc-h3.nav-item.toc-entry .pre");
+
+    tocSpans.forEach(span => {
+        const original = span.textContent;
+        const firstDotIndex = original.indexOf(".");
+        span.textContent = original.slice(firstDotIndex + 1); // Keep everything after the first dot
+    });
+});
```
mberz, line 14: Could you please a bit of documentation what this function does. 
Otherwise this piece of information may be forgotten in the future ;)

mberz (commented):
Please add some documentation, otherwise all good.

ahms5 (approved):

PR #795, opened 2025-04-25
Increase TOC depth for left sidebar to 2

mberz:
With the current level it's required to always go back to the main API docs page to change to a different sub-module or sub-page.
These can now be navigated using the left hand side-bar.


hoyer-a (approved):
nice!

Issue #780, opened 2025-03-28
Feature request: State-space Systems


artpelling:
Hi all, 

as discussed, I would like to add support for state-space systems to `pyfar`. I've briefly played around with some code, and it should be pretty straightforward to add. However, I noticed that the current code structure is a bit too tailored to filters. There would be two possible ways to deal with it. Either write a whole new module/class for state-space systems or (the one I would prefer) rearrange the class hierarchy a little. 

I've copied and pasted a quick mock-up below to get an idea of what this could look like. The main problem I face is that I cannot directly inherit from `Filter` because the entire `coefficients` mechanic does not work for state-space systems. I propose an abstract base class `LTISystem` (or any other similar name) that implements basic common functionalities between both approaches and provides a common interface. Especially stuff like `sampling_rate`, `process`, `impulse_response` and the whole `state` initialization
mechanic are present in all approaches.

``` python
#!/usr/bin/env python3

from abc import abstractmethod, ABC
import numpy as np
import pyfar as pf
from copy import deepcopy


class LTISystem(ABC):
    """Abstract class that implements a common interface for LTISystems."""
    def __init__(self, sampling_rate, comment=""):
        assert sampling_rate > 0
        self._sampling_rate = sampling_rate
        self.comment = comment

    @property
    def sampling_rate(self):
        """Sampling rate of the filter in Hz. The sampling rate is set upon
        initialization and cannot be changed after the object has been created.
        """
        return self._sampling_rate

    @abstractmethod
    def _process(self):
        pass

    def process(self, signal, reset=False):
        if not isinstance(signal, pf.Signal):
            raise ValueError("The input needs to be a Signal object.")

        if self.sampling_rate != signal.sampling_rate:
            raise ValueError(
                "The sampling rates of filter and signal do not match")

        if reset is True:
            self.reset()

        filtered_signal = deepcopy(signal)
        filtered_signal.time = self._process(signal)

        return filtered_signal

    def impulse_response(self, n_samples):
        """
        Compute or approximate the impulse response of the filter.

        See `impulse_response` methods in derived classes for more details.

        Parameters
        ----------
        n_samples : int
            Length in samples for which the impulse response is computed.

        Returns
        -------
        impulse_response : Signal
            The impulse response of the filter.
        """

        # track the state (better than copying the entire filter)
        if self.state is not None:
            state = self.state.copy()
            self._state = None
            reset = True
        else:
            reset = False

        # get impulse response
[...]

PR #798, opened 2025-04-27
add abstract LTI system class

artpelling:

- removes private `_inititalized` attribute of `Filter`
- adds new base class for all LTI systems

### DIFF
--- a/pyfar/classes/filter.py
+++ b/pyfar/classes/filter.py
@@ -106,7 +106,77 @@ def _repr_string(filter_type, order, n_channels, sampling_rate):
     return representation
 
 
-class Filter(object):
+class LTISystem(object):
+    """Base class for LTI systems."""
+
+    def __init__(self, sampling_rate=None, state=None, comment=""):
+        self._state = state
+        self._sampling_rate = sampling_rate
+        self.comment = comment
+
+    @property
+    def sampling_rate(self):
+        """Sampling rate of the filter in Hz. The sampling rate is set upon
+        initialization and cannot be changed after the object has been created.
+        """
+        return self._sampling_rate
+
+    @property
+    def state(self):
+        """
+        The current state of the filter as an array with dimensions
+        corresponding to the order of the filter and number of filter channels.
+        """
+        return self._state
+
+    def reset(self):
+        """Reset the filter state by filling it with zeros."""
+        if self._state is not None:
+            self._state = np.zeros_like(self._state)
+        else:
+            self._state = None
+
+    @property
+    def comment(self):
+        """Get comment."""
+        return self._comment
+
+    @comment.setter
+    def comment(self, value):
+        """Set comment."""
+        if not isinstance(value, str):
+            raise TypeError("comment has to be of type string.")
+        else:
+            self._comment = value
+
+    def copy(self):
+        """Return a copy of the Filter object."""
+        return deepcopy(self)
+
+    def _encode(self):
+        """Return dictionary for the encoding."""
+        return self.copy().__dict__
+
+    @classmethod
+    def _decode(cls, obj_dict):
+        """Decode object based on its respective object dictionary."""
+        # initializing this way satisfies FIR, IIR and SOS initialization
+        obj = cls(np.zeros((1, 6)), None)
+        obj.__dict__.update(obj_dict)
+        return obj
+
+    def __eq__(self, other):
+        """Check for equality of two objects."""
+        return not deepdiff.DeepDiff(self, other)
+
+    def process(self, signal, reset=False):
+        raise NotImplementedError("Abstract class method.")
+
+    def impulse_response(self, n_samples):
+        raise NotImplementedError("Abstract class method.")
+
+
+class Filter(LTISystem):
     """
     Container class for digital filters.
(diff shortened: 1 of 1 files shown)


artpelling:
following the suggestion in #780, I've added a base class for LTI systems. Does `pyfar` use abstract base classes? So far I didn't encounter them, so I stuck with the style in `classes/filter.py`...

ahms5 (commented):
thank you, just small comments to improve docstrings

artpelling:
I made `Filter` an `ABC` but now the tests are failing because `Filter` can no longer be instantiated. I cannot think of a use case of instantiating a `Filter` without a `_process` method. Is such a use intended? If not, we should remove `Filter` from the tests.

f-brinkmann:
> I made `Filter` an `ABC` but now the tests are failing because `Filter` can no longer be instantiated. I cannot think of a use case of instantiating a `Filter` without a `_process` method. Is such a use intended? If not, we should remove `Filter` from the tests.

Yes I think thise tests can be removed. Filter was always intended as a base class used only for deriving other classes.

ahms5 (approved):

mberz:
> I made `Filter` an `ABC` but now the tests are failing because `Filter` can no longer be instantiated. I cannot think of a use case of instantiating a `Filter` without a `_process` method. Is such a use intended? If not, we should remove `Filter` from the tests.

You can use something like this to set all abstracmethods, which allows testing of implemented functionality within abstract classes, that is as long as you don't call the abstract method itself of course.


```python

@patch.multiple(Filter, __abstractmethods__=set())
def test_abstract_class
``

**Comment by artpelling** [comment]
> I'm marking this as "implementation in progress" (given the last commit message contains WIP)

yes please! I need to give the tests some more thought. Not entirely happy atm

File: pyfar/classes/filter.py [hunk 1]
```diff
@@ -194,29 +261,15 @@ def coefficients(self, value):
...
-    @property
-    def state(self):
-        """
-        The current state of the filter as an array with dimensions
-        corresponding to the order of the filter and number of filter channels.
-        """
-        return self._state
-
     @staticmethod
+    @abstractmethod
```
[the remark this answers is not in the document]
artpelling: No

pyfar/classes/filter.py
```diff
@@ -106,12 +109,74 @@ def _repr_string(filter_type, order, n_channels, sampling_rate):
...
+        The sampling rate is set upon initialization and cannot be changed
+        after the object has been created.
+        """
+        return self._sampling_rate
+
+    @property
+    def state(self):
+        """
+        The current state of the system as an array with dimensions
+        corresponding to the order of the filter and number of filter channels.
```
sikersten: For this abstract class, neither an "order" or a number of channels is defined. Is there a way to define the state dimensions more generally?

pyfar/classes/filter.py
```diff
@@ -106,12 +109,74 @@ def _repr_string(filter_type, order, n_channels, sampling_rate):
...
+        """Check for equality of two objects."""
+        return not deepdiff.DeepDiff(self, other)
+
+    @abstractmethod
+    def process(self, signal, reset=False):
+        """Abstract method that computes the system's forced response."""
+
+    @abstractmethod
+    def impulse_response(self, n_samples):
+        """Abstract method that computes the system impulse response."""
```
sikersten:
```suggestion
        """Abstract method that computes the system's impulse response."""
```

pyfar/classes/filter.py
```diff
@@ -106,12 +109,74 @@ def _repr_string(filter_type, order, n_channels, sampling_rate):
     return representation
 
 
-class Filter(object):
-    """
-    Container class for digital filters.
-    This is an abstract class method, only used for the shared processing
-    method used for the application of a filter on a signal.
-    """
+class LTISystem(ABC):
```
sikersten: Since this class is not inteded for direct use, I would suggest making it private:

```suggestion
class _LTISystem(ABC):
```

sikersten: I think we did not really had this rule when the `Filter` class was implemented.

pyfar/classes/filter.py
artpelling: I don't think so. One can argue that all practically relevant systems are finite dimensional and therefore have an order. I think that is too technical here, I've just made the docstring more generic. Especially the part about filter channels does not apply here and needs to go.

pyfar/classes/filter.py
```diff
@@ -106,12 +109,73 @@ def _repr_string(filter_type, order, n_channels, sampling_rate):
...
 
-class Filter(object):
-    """
-    Container class for digital filters.
-    This is an abstract class method, only used for the shared processing
-    method used for the application of a filter on a signal.
-    """
+class _LTISystem(ABC):
+    """Abstract base class for LTI systems."""
+
```
mberz, line 124: Could you please add a docstring here similar to FIlterFIR etc.?

artpelling: Sure, I've also changed the docstrings of the other classes such that they have the same structure. I took the `Signal` class as an example. Hope that was good.

PR #803, opened 2025-05-20
Coordinates._set_points: separate internal checks from setting the variables

ahms5:
Which issue(s) does this pull request close?

This might be related to https://github.com/pyfar/spharpy/pull/121/files.
The problem is that if additional checks are required by subclasses (e.g. SamplingSphere), the _set_points need to be reimplemented, which causes duplicate code. This can be overcome by separating the input checks and setting the internal variables into two methods.

- Rename _set_points to _check_and_set_points to make the purpose clearer.
- introduce '_set_points', which simply sets x, y and z to self.x, self.y and self.z.
- introduce '_check_points', which performs the input checks and broadcasting to the arrays.

### DIFF
--- a/pyfar/classes/coordinates.py
+++ b/pyfar/classes/coordinates.py
@@ -168,7 +168,7 @@ class Coordinates():
         super(Coordinates, self).__init__()
 
         # init cartesian coordinates
-        self._set_points(x, y, z)
+        self._check_and_set_points(x, y, z)
 
         # save meta data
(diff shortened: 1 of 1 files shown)


f-brinkmann (approved):
I like this but we would have to release a patch soon to be usefull for spharpy

ahms5:
we could do the same for the _set_weights, good point to do it in a patch release, I'll create a PR for that merging to main

PR #805, opened 2025-05-20
 Patch: Coordinates._set_weights: separate internal checks from setting the variables

ahms5:

Very similar to  #804, but for the Coordinates._set_weights() method

- move weights checks from ``_set_weights`` into a dedicated method which just do the input checks called ``_check_weights``
- does not change the behaviour of Coordinates

### DIFF
--- a/pyfar/classes/coordinates.py
+++ b/pyfar/classes/coordinates.py
@@ -2733,16 +2733,41 @@ class Coordinates():
 
     def _set_weights(self, weights):
         """
-        Check and set sampling weights.
+        If not None convert to float64 numpy array and check for size.
+
+        Parameters
+        ----------
+        weights : array like, number
+            the weights for each point.
 
         Set self._weights, which is an atleast_1d numpy array of shape
         [L,M,...,N].
         """
+        # check weights and convert to numpy array if not None
+        weights = self._check_weights(weights)
 
-        # check input
+        # set class variable
+        self._weights = weights
+
+    def _check_weights(self, weights):
+        """
+        Convert weights into float64 numpy array and check versus the csize.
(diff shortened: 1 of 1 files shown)


f-brinkmann (approved):

hoyer-a (approved):

pyfar/classes/coordinates.py
```diff
@@ -2733,16 +2733,41 @@ def _set_points(self, x, y, z):
...
+    def _check_weights(self, weights):
+        """
+        Convert weights into float64 numpy array and check versus the csize.
+        It will be reshaped to the cshape if the csize matches.
 
-        # check input
+        Parameters
+        ----------
+        weights : array like, number
+            the weights for each point, should be of size of self.csize.
```
mberz, line 2760:
```suggestion
            The weights for each point, should be of size of self.csize.
```

pyfar/classes/coordinates.py
```diff
@@ -2733,16 +2733,41 @@ def _set_points(self, x, y, z):
...
+        # check weights and convert to numpy array if not None
+        weights = self._check_weights(weights)
+
+        # set class variable
+        self._weights = weights
+
+    def _check_weights(self, weights):
+        """
+        Convert weights into float64 numpy array and check versus the csize.
+        It will be reshaped to the cshape if the csize matches.
```
mberz, line 2755:
```suggestion
        Check if the data is broadcastable to match the cshape. 
```

mberz: Maybe something like this?

mberz (commented):
Maybe improve the docstrings a bit. Otherwise good.

PR #807, opened 2025-05-23
Download signals.files from new location

f-brinkmann:

Files were previously stored on pyfar.org, which was inconvienent to access and maintain for most pyfar members. They are now stored on github: https://github.com/pyfar/files


mberz (approved):

ahms5 (commented):
how about using [pooch.retrieve](https://www.fatiando.org/pooch/latest/retrieve.html) here?
pros:
- we dont need to think how to chach it
- we do md5 hash checks, to make sure everthing went right.
- seems like the cache would be the same for each installation of pyfar, so we just need to download it once. And not everytime again, if reinstall it.
cons:
- another dependency

ahms5 (approved):

PR #811, opened 2025-06-24
Feature: frequency weighting constants and filter

### LINKED ISSUE #761: sub-feature: (third-)octave band frequency weighting constants
Provide the A and C weighting level corrections from IEC/DIN EN 61672 available in pyfar, to use them in combination with the fractional_octave filters

jf-som:

Closes #761 and #758

Implementing octave/third-band correction values, A/C-weighting curves and a frequency weighting filter based on these curves, as described in #675.
This merge request is essentially a revamped version of #759. I decided to do this in a new branch as the previous is somewhat old by now and there were a lot of changes requested since then.

In summary, the frequency weighting filter works by running a least-squares optimization on the zeros and poles on the z-plane, which are then turned into an SOS filter. 

Points of interest for discussion:

**Redundant third-band definition:** 
The definition and selection of nominal frequency bands in a range in `constants.frequency_weighting_band_corrections` is largely redundant with `dsp.filter.fractional_octave` functions. However, since the two functions are based on different standards (61260 vs 61672), there are some differences: 61260 allows other octave fractions than 1 and 3, whereas 61672 only defines octave and third bands. On the other hand, 61672 defines bands down to 10 Hz instead of 25 Hz. 
Should we merge these helper functions anyway?

**Performance/Caching:**
While `dsp.filter.frequency_weighting_filter` adapts the usual style of other filter functions, it is different in that it runs an approximation internally, thus taking more computation time (a fraction of a second each call). If used in a loop, or by several different level functions later on, this might add up and make scripts slower. 
I thought about caching, but the current implementation is not cachable to my knowledge, since it allows users to specify an error_weighting function and kwargs for the optimizer as arguments, which are not hashable (https://docs.python.org/3/library/functools.html#functools.lru_cache). 
If we remove these options, the design function would still produce good results (class 1 compliant at all usual sampling rates) and become cachable, but we would prevent users from finding more optimal solutions, or ones that work even for unusual/extreme sampling rates.
So we could:
1. Leave it as is and accept that it might be slow for repeated calls (A hint is provided in the docstring comment)
2. Remove the unhashable parameters, limiting what the function can do (while also hiding potentially unwanted complexity)
3. Add a cachable wrapper function around the design function that is used in case only the mandatory parameters are specified. So we get caching in most cases, except when the optional parameters are used. This should be the best of both worlds, at the cost of slightly more complex code.

I prefer option 3 but want to hear your opinions first.

### DIFF
--- a/pyfar/constants/frequency_weighting.py
+++ b/pyfar/constants/frequency_weighting.py
@@ -1,0 +1,137 @@
+"""File containing all frequency weighting functions."""
+from typing import Literal
+import numpy as np
+
+# constants for the weighting curve formulas
+_F_1 = 20.598997057618316
+_F_2 = 107.65264864304629
+_F_3 = 737.8622307362901
+_F_4 = 12194.217147998012
+_A_1000 = -2
+_C_1000 = -0.062
+
+
+def _calculate_A_weighted_level(f: float | np.ndarray) -> float | np.ndarray:
+    """
+    Calculates the level correction in dB of a frequency component
+    when using the A weighting.
+    """
+    bracket_term = (_F_4**2 * f**4) / ((f**2 + _F_1**2) * (f**2 + _F_2**2)
+                                      ** 0.5 * (f**2 + _F_3**2)**0.5
+                                      * (f**2 + _F_4**2))
+    return 10 * np.log10(bracket_term**2) - _A_1000
+
+
+def _calculate_C_weighted_level(f: float | np.ndarray) -> float | np.ndarray:
+    """
+    Calculates the level correction in dB of a frequency component
+    when using the C weighting.
+    """
+    bracket_term = (_F_4**2 * f**2) / ((f**2 + _F_1**2) * (f**2 + _F_4**2))
+    return 10 * np.log10(bracket_term**2) - _C_1000
+
+
+def frequency_weighting_curve(weighting: Literal["A", "C"],
+            frequency: float | np.ndarray) -> float | np.ndarray:
+    """
+    Calculates the level correction in dB of a frequency component when using
+    the A or C weighting defined in IEC 61672-1.
+
+    Parameters
+    ----------
+    weighting: str ('A' or 'C')
+        Which frequency weighting to use.
+
+    frequency: float or array-like
+        The frequency or frequencies at which to evaluate the weighting curve.
+        The return value matches the shape of this argument.
+
+    Returns
+    -------
+    weights: float or array-like
+        The weights in dB in the same shape as the frequency arguments.
+    """
+    if weighting == "A":
+        return _calculate_A_weighted_level(frequency)
+    elif weighting == "C":
+        return _calculate_C_weighted_level(frequency)
+    else:
+        raise ValueError("Allowed literals for weighting are 'A' and 'C'")
+
+
+# constants for nominal band corrections.
+# This is somewhat redundant with dsp.filter.fractional_octaves -> move helper
+# functions to a new place where it can be accessed from everyhwere?
+_NOMINAL_THIRDBAND_FREQUENCIES = np.array([
+    10, 12.5, 16, 20, 25, 31.5, 40, 50, 63, 80, 100, 125, 160, 200, 250, 315,
+    400, 500, 630, 800, 1000, 1250, 1600, 2000, 2500, 3150, 4000, 5000, 6300,
+    8000, 10000, 12500, 16000, 20000,
+])
+_THIRDBAND_WEIGHTINGS_A = np.array([
+    -70.4, -63.4, -56.7, -50.5, -44.7, -39.4, -34.6, -30.2, -26.2, -22.5,
+    -19.1, -16.1, -13.4, -10.9, -8.6, -6.6, -4.8, -3.2, -1.9, -0.8,
+    0.0, 0.6, 1.0, 1.2, 1.3, 1.2, 1.0, 0.5, -0.1, -1.1,
+    -2.5, -4.3, -6.6, -9.3,
+])
+_THIRDBAND_WEIGHTINGS_C = np.array([
+    -14.3, -11.2, -8.5, -6.2, -4.4, -3.0, -2.0, -1.3, -0.8, -0.5,
+    -0.3, -0.2, -0.1, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0,
+    0.0, 0.0, -0.1, -0.2, -0.3, -0.5, -0.8, -1.3, -2.0, -3.0, -4.4,
+    -6.2, -8.5, -11.2,
+])
+
+
+def frequency_weighting_band_corrections(
+    weighting: Literal["A", "C"],
+    bands: Literal["third", "octave"],
+    frequency_range: tuple[float, float],
+) -> tuple[np.ndarray, np.ndarray]:
+    """
+    Returns the A or C frequency weighting band corrections as specified in
+    IEC 62672-1 for the given frequency range. The lowest defined center
+    frequency is 10 Hz and the highest is 20 kHz.
+
+    Parameters
+    ----------
+    weighting: str ('A' or 'C')
+        Which frequency weighting type to use.
+
+    bands: str ('octave' or 'third')
+        Whether to use octave bands or third bands.
+
+    frequency_range: (float, float)
+        A range of what nominal center frequencies to include.
+
+    Returns
+    -------
+    nominal_frequencies: NDArray
+        The nominal center frequencies included in the given range.
+
+    weights: NDArray
+        The correction values in dB for the specific frequency weighting.
+    """
+    if weighting == "A":
+        all_weights = _THIRDBAND_WEIGHTINGS_A
+    elif weighting == "C":
+        all_weights = _THIRDBAND_WEIGHTINGS_C
+    else:
+        raise ValueError("Allowed literals for weighting are 'A' and 'C'")
+
+    if bands == "octave":
+        nominals = _NOMINAL_THIRDBAND_FREQUENCIES[2::3]
+        band_weights = all_weights[2::3]
+    elif bands == "third":
+        nominals =_NOMINAL_THIRDBAND_FREQUENCIES
+        band_weights = all_weights
+    else:
+        raise ValueError("Allowed literals for bands are 'octave' and 'third'")
+
+    mask = (nominals >= frequency_range[0]) & (nominals <= frequency_range[1])
+    nominals_in_range = nominals[mask]
+    weights_in_range = band_weights[mask]
+
+    if len(nominals_in_range) == 0:
+        raise ValueError("Frequency range must include at least one value"
+                         + "between 10 and 20000")
+
+    return (nominals_in_range, weights_in_range)
--- a/pyfar/constants/frequency_weighting.py
+++ b/pyfar/constants/frequency_weighting.py
@@ -1,5 +1,5 @@
 """File containing all frequency weighting functions."""
-from typing import Literal
+from typing import Literal, Union
 import numpy as np
 
 # constants for the weighting curve formulas
@@ -11,7 +11,8 @@ _A_1000 = -2
 _C_1000 = -0.062
 
 
-def _calculate_A_weighted_level(f: float | np.ndarray) -> float | np.ndarray:
+def _calculate_A_weighted_level(f: Union[float, np.ndarray],
+                                ) -> Union[float, np.ndarray]:
     """
     Calculates the level correction in dB of a frequency component
     when using the A weighting.
@@ -22,7 +23,8 @@ def _calculate_A_weighted_level(f: float | np.ndarray) -> float | np.ndarray:
     return 10 * np.log10(bracket_term**2) - _A_1000
 
 
-def _calculate_C_weighted_level(f: float | np.ndarray) -> float | np.ndarray:
+def _calculate_C_weighted_level(f: Union[float, np.ndarray],
+                                ) -> Union[float, np.ndarray]:
     """
     Calculates the level correction in dB of a frequency component
     when using the C weighting.
@@ -32,7 +34,7 @@ def _calculate_C_weighted_level(f: float | np.ndarray) -> float | np.ndarray:
 
 
 def frequency_weighting_curve(weighting: Literal["A", "C"],
-            frequency: float | np.ndarray) -> float | np.ndarray:
+            frequency: Union[float, np.ndarray]) -> Union[float, np.ndarray]:
     """
     Calculates the level correction in dB of a frequency component when using
     the A or C weighting defined in IEC 61672-1.
--- a/pyfar/dsp/filter/frequency_weighting.py
+++ b/pyfar/dsp/filter/frequency_weighting.py
@@ -1,0 +1,336 @@
+"""
+Functions and helpers to create frequency weighting filters
+according to IEC 61672-1.
+"""
+from typing import Callable, Literal
+import numpy as np
+import pyfar as pf
+import pyfar.constants as pfc
+import scipy.signal as sps
+import scipy.optimize as spo
+from pyfar.constants.frequency_weighting import _F_1, _F_2, _F_3, _F_4
+
+
+_A_WEIGHTING_ZEROS = [0, 0, 0, 0]
+_A_WEIGHTING_POLES = [_F_1, _F_1, _F_4, _F_4, _F_2, _F_3]
+_C_WEIGHTING_ZEROS = [0, 0]
+_C_WEIGHTING_POLES = [_F_1, _F_1, _F_4, _F_4]
+
+
+def frequency_weighting_filter(
+        signal,
+        target_weighting: Literal["A", "C"]="A",
+        n_frequencies=100,
+        error_weighting: Callable[[
+            np.ndarray], np.ndarray] = None,
+        sampling_rate: float | None = None,
+        **kwargs,
+        ):
+    """
+    Creates an efficient SOS filter approximating the A or C weighting defined
+    in IEC/DIN-EN 61672-1.
+    When using default parameters, the returned filter is compliant with a
+    class 1 sound level meter as described in the norm for all tested
+    sampling rates (see Notes).
+    This function will run a least-squares algorithm to iteratively approach
+    the target weighting curve.
+    Therefore, it may be much faster to create the filter once and reuse that
+    filter when repeatedly weighting audio data of identical sampling rates.
+
+    Parameters
+    ----------
+    signal : Signal, None
+        The signal to be filtered. Pass ``None`` to create the filter without
+        applying it.
+
+    target_weighting: str, optional
+        Specifies which frequency weighting curve to approximate.
+        Must be either "A" or "C". The default is "A".
+
+    n_frequencies: int, optional
+        At how many frequencies to evaluate the filter coefficients during
+        optimization. Less frequencies means faster iterations, but
+        potentially worse results. The evaluation frequencies are
+        logarithmically spaced between 10 Hz and the Nyquist frequency.
+
+    error_weighting: callable
+        A function that can be used to emphasize the approximation errors in
+        specific frequency ranges. The function should take a float array as
+        argument, specifying the normalized frequencies between 0 and 1
+        (where 1 is the Nyquist frequency) to weight, and returns a float
+        array as output containing the weights.
+        The default value is None, in which case the errors of all
+        (logarithmically spaced) frequencies are equally weighted. This
+        usually leads to larger errors for higher frequencies. By passing
+        a function that emphasizes high frequencies, it is possible to reduce
+        this effect and get a filter potentially closer to the target curve.
+        Example: `error_weighting=lambda nf: 100**nf`. This example often
+        leads to better results for typical sampling rates, but much worse
+        for very high rates.
+
+    sampling_rate: float, conditionally optional
+        The sampling rate of the returned filter.
+
+    **kwargs: dict
+        Keyword args that are passed to the
+        :py:func:`scipy.optimize.least_squares` call.
+
+    Returns
+    -------
+    filter: FilterSOS
+        The frequency weighting filter in second order sections format.
+        If weighting is 'A' the filter order will be 6. 'C' weighting will
+        return a filter of order 4.
+
+    Notes
+    -----
+    Since this function performs an iterative approximation, results may not
+    be perfect depending on the input parameters. With the default parameters,
+    the filter will be at least class 1 compliant for the sampling rates
+    48 kHz, 44.1 kHz, 16 kHz as well as these sampling rates multiplied by
+    2, 4, 8, 0.5, 0.25 and 0.125 each.
+    With non-default parameters and/or other sampling rates, the returned
+    filter may not comply with class 1 requirements,
+    in which case a warning will be printed.
+    """
+    # check input
+    if (signal is None and sampling_rate is None) \
+            or (signal is not None and sampling_rate is not None):
+        raise ValueError('Either signal or sampling_rate must be none.')
+
+    # sampling frequency in Hz
+    fs = signal.sampling_rate if sampling_rate is None else sampling_rate
+
+    sos = _design_frequency_weighting_filter(fs, target_weighting,
+                                            n_frequencies, error_weighting,
+                                            **kwargs)
+    filt = pf.FilterSOS([sos], sampling_rate)
+    filt.comment = (f"Frequency weighting SOS filter of order {filt.order} "
+                    f"to approximate {target_weighting} weighting according "
+                    "to IEC 61672-1")
+
+    # apply or return the filter object
+    if signal is None:
+        return filt
+    else:
+        signal_filt = filt.process(signal)
+        return signal_filt
+
+
+def _design_frequency_weighting_filter(sampling_rate: float,
+                                      target_weighting: Literal["A", "C"]="A",
+                                      n_frequencies=100,
+                                      error_weighting: Callable[[
+                                          np.ndarray], np.ndarray] = None,
+                                      **kwargs,
+                                      ) -> np.ndarray:
+    """
+    Designs SOS filter coefficients approximating the A or C weighting defined
+    in IEC/DIN-EN 61672-1 for an arbitrary sampling rate.
+
+    Parameters
+    ----------
+    sampling_rate: float
+        The sampling rate of the returned filter.
+
+    target_weighting: str, optional
+        Specifies which frequency weighting curve to approximate.
+        Must be either "A" or "C". The default is "A".
+
+    n_frequencies: int, optional
+        At how many frequencies to evaluate the filter coefficients during
+        optimization. Less frequencies means faster iterations, but
+        potentially worse results. The evaluation frequencies are
+        logarithmically spaced between 10 Hz and the Nyquist frequency.
+
+    error_weighting: callable
+        A function that can be used to emphasize the approximation errors in
+        specific frequency ranges. The function should take a float array as
+        argument, specifying the normalized frequencies between 0 and 1
+        (where 1 is the Nyquist frequency) to weight, and returns a float
+        array as output containing the weights.
+        The default value is None, in which case the errors of all
+        (logarithmically spaced) frequencies are equally weighted. This
+        usually leads to larger errors for higher frequencies. By passing
+        a function that emphasizes high frequencies, it is possible to reduce
+        this effect and get a filter potentially closer to the target curve.
+        Example: `error_weighting=lambda nf: 100**nf`. This example often
+        leads to better results for typical sampling rates, but much worse
+        for very high rates.
+
+    **kwargs: dict
+        Keyword args that are passed to the
+        :py:func:`scipy.optimize.least_squares` call.
+
+    Returns
+    -------
+    sos_coefficients: NDarray
+        The coefficients of the designed filter in scipy's sos format.
+
+    Notes
+    -----
+    Since this function performs an iterative approximation, results may not
+    be perfect depending on the input parameters. With the default parameters,
+    the filter will be at least class 1 compliant for the sampling rates
+    48 kHz, 44.1 kHz, 16 kHz as well as these sampling rates multiplied by
+    2, 4, 8, 0.5, 0.25 and 0.125 each.
+    With non-default parameters and/or other sampling rates, the returned
+    filter may not comply with class 1 requirements,
+    in which case a warning will be printed.
+    """
+
+    if target_weighting not in ["A", "C"]:
+        raise ValueError(f"Unrecognized weighting: '{target_weighting}'.",
+                         " Only 'A' and 'C' are implemented")
+
+    # Build the initial guess x0 (the variables to optimize) from
+    # zeros, poles and gain. Use the poles from the bilinear filter,
+    # since they are not too far from the approximation.
+    _z, p0, _k = _zpk_from_analog(target_weighting, sampling_rate)
+    # The zeros at (1, 0) (=0 Hz) are fixed. The remaining two are wrong from
+    # the bilinear transform, so we optimize them starting from 0
+    # (the z-plane origin).
+    z0 = [0, 0]
+    k0 = 1
+    x0 = np.concat([z0, p0, [k0]])
+
+    # build the cost function (as a closure, so that passing it to
+    # least_squares() is easier and cleaner)
+    frequencies = np.logspace(1, np.log10(sampling_rate/2), n_frequencies)
+    target_levels = pfc.frequency_weighting_curve(target_weighting,
+                                                  frequencies)
+
+    def compute_residuals(x: np.ndarray) -> np.ndarray:
+        """
+        Lets least_squares() compute how close the current approximation is to
+        the target. Treats x (z, p and k) as real values, limiting them to
+        the real axis.
+        """
+        z, p, k = _x2zpk(x, target_weighting)
+        _, freq_response = sps.freqz_zpk(
+            z, p, k, frequencies, fs=sampling_rate)
+        approximated_levels = 20 * np.log10(abs(freq_response))
+
+        # use level differences (in dB) as cost
+        level_diffs = target_levels - approximated_levels
+        if error_weighting:
+            level_diffs *= error_weighting(frequencies / (sampling_rate/2))
+        return level_diffs
+
+    # Bounds are necessary to keep poles inside the unit circle,
+    # ensuring the resulting filter is stable.
+    kwargs.setdefault("bounds", (-1, 1))
+    # The actual optimization happens here. This iteratively moves
+    # zeros and poles along the real axis on the z-plane until
+    # a local optimum is reached.
+    optimization_result = spo.least_squares(compute_residuals, x0, **kwargs)
+
+    # turn the result into a filter object
+    z, p, k = _x2zpk(optimization_result.x, target_weighting)
+    sos = sps.zpk2sos(z, p, k)
+
+    # check if the filter is class 1 compliant
+    is_class_1, max_err, mean_err = _check_filter(
+        sampling_rate, sos, target_weighting)
+    if not is_class_1:
+        print(f"Warning: The generated {target_weighting} weighting filter is"
+              + "not class 1 compliant; "
+              + f"Max error: {max_err:.2} dB, mean error: {mean_err:.2} dB.")
+    return sos
+
+
+
+def _zpk_from_analog(weighting: str, fs: float,
+                     ) -> tuple[np.ndarray, np.ndarray, float]:
+    """
+    Transforms the analog (s-plane) A or C weighting filter as specified in
+    IEC/DIN-EN 61672-1 to a digital filter (z-plane) via bilinear transform
+    and returns it as zeros poles and gain.
+    The resulting filter is heavily warped near the Nyquist frequency.
+    """
+    if weighting == "A":
+        zeros, poles = (_A_WEIGHTING_ZEROS, _A_WEIGHTING_POLES)
+    elif weighting == "C":
+        zeros, poles = (_C_WEIGHTING_ZEROS, _C_WEIGHTING_POLES)
+    else:
+        raise ValueError("Unrecognized weighting:", weighting)
+
+    poles_angular = np.array(poles) * -2 * np.pi
+    return sps.bilinear_zpk(zeros, poles_angular, 1, fs)
+
+
+def _x2zpk(x: np.ndarray,
+           target_weighting: Literal["A", "C"],
+           ) -> tuple[np.ndarray, np.ndarray, float]:
+    """
+    Reobtain z, p and k from x (the 1-D array of optimizer inputs).
+    Adds back the fixed zeros at 0 Hz that are excluded from optimization.
+    """
+    fixed_zeros = [1, 1, 1, 1] if target_weighting == "A" else [1, 1]
+    zeros = np.concat([fixed_zeros, x[0:2]])
+    return (zeros, x[2:-1], x[-1])
+
+
+
+def _get_error_margins(f: np.ndarray):
+    """
+    Returns the upper and lower error margin in dB for every frequency in the
+    input array for the A and C weighting according to IEC/DIN-EN 61672-1.
+    """
+    upper = np.zeros_like(f)
+    upper[f <= 10] = 3.0
+    upper[(f > 10) & (f <= 12.5)] = 2.5
+    upper[(f > 12.5) & (f <= 25)] = 2.0
+    upper[(f > 25) & (f <= 31.5)] = 1.5
+    upper[(f > 31.5) & (f <= 800)] = 1.0
+    upper[(f > 800) & (f < 1250)] = 0.7
+    upper[(f >= 1250) & (f < 5000)] = 1.0
+    upper[(f >= 5000) & (f < 10e3)] = 1.5
+    upper[(f >= 10e3) & (f < 16e3)] = 2.0
+    upper[(f >= 16e3) & (f < 20e3)] = 2.5
+    upper[f >= 20e3] = 3.0
+
+    lower = np.zeros_like(f)
+    lower[f <= 12.5] = -np.inf
+    lower[(f > 12.5) & (f <= 16)] = -4.0
+    lower[(f > 16) & (f <= 20)] = -2.0
+    lower[(f > 20) & (f <= 31.5)] = -1.5
+    lower[(f > 31.5) & (f <= 800)] = -1.0
+    lower[(f > 800) & (f < 1250)] = -0.7
+    lower[(f >= 1250) & (f < 5000)] = -1.0
+    lower[(f >= 5000) & (f < 6300)] = -1.5
+    lower[(f >= 6300) & (f < 8000)] = -2.0
+    lower[(f >= 8000) & (f < 10e3)] = -2.5
+    lower[(f >= 10e3) & (f < 12500)] = -3.0
+    lower[(f >= 12500) & (f < 16e3)] = -5.0
+    lower[(f >= 16e3) & (f < 20e3)] = -16.0
+    lower[f >= 20e3] = -np.inf
+
+    return upper, lower
+
+
+def _check_filter(sampling_rate: float,
+                  sos: np.ndarray,
+                  weighting: Literal["A", "C"],
+                  ) -> tuple[bool, float, float]:
+    """
+    Checks whether the filter is class 1 compliant according to the norm and
+    provides simple error statistics.
+    """
+    test_freqs = np.logspace(1, np.log10(sampling_rate / 2), 100_000)
+
+    # calculate the magnitude differences between the filter response
+    # and the target weighting
+    z, p, k = sps.sos2zpk(sos)
+    _, freq_resp = sps.freqz_zpk(z, p, k, test_freqs, fs=sampling_rate)
+    mags = np.abs(freq_resp)
+    mags_dB = 10 * np.log10(mags**2)
+    mag_diffs = mags_dB - pfc.frequency_weighting_curve(weighting, test_freqs)
+
+    upper, lower = _get_error_margins(test_freqs)
+    is_class_1 = np.all(mag_diffs < upper) and np.all(mag_diffs >= lower)
+    max_diff = np.max(np.abs(mag_diffs))
+    mean_diff = np.mean(np.abs(mag_diffs))
+    return is_class_1, max_diff, mean_diff
+
+
--- a/pyfar/dsp/filter/frequency_weighting.py
+++ b/pyfar/dsp/filter/frequency_weighting.py
@@ -2,7 +2,7 @@
 Functions and helpers to create frequency weighting filters
 according to IEC 61672-1.
 """
-from typing import Callable, Literal
+from typing import Callable, Literal, Optional
 import numpy as np
 import pyfar as pf
 import pyfar.constants as pfc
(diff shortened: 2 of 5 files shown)


pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,139 @@
...
+_C_1000 = -0.062
+
+
+def _calculate_A_weighted_level(f: Union[float, np.ndarray],
+                                ) -> Union[float, np.ndarray]:
+    """
+    Calculates the level correction in dB of a frequency component
+    when using the A weighting.
+    """
+    bracket_term = (_F_4**2 * f**4) / ((f**2 + _F_1**2) * (f**2 + _F_2**2)
```
f-brinkmann, line 41: Can you add a comment here and below mentioning the equations from the Standard? This will make things easier to maintain (and review).

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,139 @@
...
+    the A or C weighting defined in IEC 61672-1.
+
+    Parameters
+    ----------
+    weighting: str ('A' or 'C')
+        Which frequency weighting to use.
+
+    frequency: float or array-like
+        The frequency or frequencies at which to evaluate the weighting curve.
+        The return value matches the shape of this argument.
```
f-brinkmann: Should there be a check for supported min and max values?

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,139 @@
...
+        return _calculate_A_weighted_level(frequency)
+    elif weighting == "C":
+        return _calculate_C_weighted_level(frequency)
+    else:
+        raise ValueError("Allowed literals for weighting are 'A' and 'C'")
+
+
+# constants for nominal band corrections.
+# This is somewhat redundant with dsp.filter.fractional_octaves -> move helper
+# functions to a new place where it can be accessed from everyhwere?
```
f-brinkmann: Since the other function can not return the lowest nominal frequencies I would leave this as is, but mention this in the comment.

Please also add a hint to the equation from the standard.

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,139 @@
...
+
+    Returns
+    -------
+    weights: float or array-like
+        The weights in dB in the same shape as the frequency arguments.
+    """
+    if weighting == "A":
+        return _calculate_A_weighted_level(frequency)
+    elif weighting == "C":
+        return _calculate_C_weighted_level(frequency)
```
f-brinkmann: I think it would be more consistent to return FrequencyData objects here and also below. We already do this in `air_attenuation`

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,139 @@
...
+
+    mask = (nominals >= frequency_range[0]) & (nominals <= frequency_range[1])
+    nominals_in_range = nominals[mask]
+    weights_in_range = band_weights[mask]
+
+    if len(nominals_in_range) == 0:
+        raise ValueError("Frequency range must include at least one value"
+                         + "between 10 and 20000")
+
+    return (nominals_in_range, weights_in_range)
```
f-brinkmann, line 193: see above - I would suggest to return a FrequencyData object.

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,139 @@
...
+    -14.3, -11.2, -8.5, -6.2, -4.4, -3.0, -2.0, -1.3, -0.8, -0.5,
+    -0.3, -0.2, -0.1, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0,
+    0.0, 0.0, -0.1, -0.2, -0.3, -0.5, -0.8, -1.3, -2.0, -3.0, -4.4,
+    -6.2, -8.5, -11.2,
+])
+
+
+def frequency_weighting_band_corrections(
+    weighting: Literal["A", "C"],
+    bands: Literal["third", "octave"],
```
f-brinkmann: This is elegant, but for consistency I would suggest to rename to `num_fraction` with `1` denoting octaces and `3` third octaves.

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,139 @@
...
+    frequency_range: tuple[float, float],
+) -> tuple[np.ndarray, np.ndarray]:
+    """
+    Returns the A or C frequency weighting band corrections as specified in
+    IEC 62672-1 for the given frequency range. The lowest defined center
+    frequency is 10 Hz and the highest is 20 kHz.
```
f-brinkmann: I would move the information about the range to the parameter section

tests/test_frequency_weighting.py
```diff
@@ -0,0 +1,54 @@
...
+import pyfar.dsp.filter as pffilt
+
+
+def test_frequency_weighting_constants():
+    weightings = ["A", "C"]
+    band_options = ["octave", "third"]
+    freq_ranges = [
+        (20, 20000),
+        (0, 100),
+        (4000, 100_000),
```
f-brinkmann: This can be done more elegantly using `pytest.mark.parametrize` you can search the code for examples of this. We try to avoid such loops in testing because it is hard to debug. If something goes wrong, you won't know at which iteration of the loop.

tests/test_frequency_weighting.py
```diff
@@ -0,0 +1,54 @@
...
+
+    (nominals, iec_weights) = pyfar.constants. \
+            frequency_weighting_band_corrections(weighting, bands, freq_range)
+    calculated_weights = pyfar.constants. \
+            frequency_weighting_curve(weighting, nominals)
+
+    # maybe there is a better way of testing. tolerance is 0.2, because
+    # we need 0.1 for rounding the weight and then larger error
+    # due to the difference in nominal vs exact frequency being evaluated
+    npt.assert_allclose(calculated_weights, iec_weights, atol=0.3)
```
f-brinkmann: I like the idea of cross-checking the two implementations. After using parametrize (see comment above), its probably obsolete to have this as a separate function.

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,336 @@
...
+    **kwargs: dict
+        Keyword args that are passed to the
+        :py:func:`scipy.optimize.least_squares` call.
+
+    Returns
+    -------
+    filter: FilterSOS
+        The frequency weighting filter in second order sections format.
+        If weighting is 'A' the filter order will be 6. 'C' weighting will
+        return a filter of order 4.
```
f-brinkmann, line 97: It should be documented that a filtered or a filter can be returned. You can check the other filter functions for examples.

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,336 @@
...
+        target_weighting: Literal["A", "C"]="A",
+        n_frequencies=100,
+        error_weighting: Optional[Callable[[
+            np.ndarray], np.ndarray]] = None,
+        sampling_rate: Optional[float] = None,
+        **kwargs,
+        ):
+    """
+    Creates an efficient SOS filter approximating the A or C weighting defined
+    in IEC/DIN-EN 61672-1.
```
f-brinkmann: We try to start with a one liner (see other filter functions for example) I would suggest 'Create and apply A or C weighting filter' and move the remaining information to the body of the docstring.

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,336 @@
...
+    When using default parameters, the returned filter is compliant with a
+    class 1 sound level meter as described in the norm for all tested
+    sampling rates (see Notes).
+    This function will run a least-squares algorithm to iteratively approach
+    the target weighting curve.
+    Therefore, it may be much faster to create the filter once and reuse that
+    filter when repeatedly weighting audio data of identical sampling rates.
```
f-brinkmann: You probably mean default values for the parameters `n_frequencies` and `error_weighting` right? This should be clarified, otherwise people might think the default C-filters are not class 1.

Regarding your question from the pull-request: I could live without caching and find the information provided here sufficient. However, I would suggest to make it more visible using the `.. note::` environment (you can search pyfar for examples) and also have the supported sampling rates directly here. I would also try to shorten the information a little bit.

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,336 @@
...
+
+    target_weighting: str, optional
+        Specifies which frequency weighting curve to approximate.
+        Must be either "A" or "C". The default is "A".
+
+    n_frequencies: int, optional
+        At how many frequencies to evaluate the filter coefficients during
+        optimization. Less frequencies means faster iterations, but
+        potentially worse results. The evaluation frequencies are
+        logarithmically spaced between 10 Hz and the Nyquist frequency.
```
f-brinkmann, line 66: Please mention the default also here.

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,336 @@
...
+        Example: `error_weighting=lambda nf: 100**nf`. This example often
+        leads to better results for typical sampling rates, but much worse
+        for very high rates.
+
+    sampling_rate: float, conditionally optional
+        The sampling rate of the returned filter.
+
+    **kwargs: dict
+        Keyword args that are passed to the
+        :py:func:`scipy.optimize.least_squares` call.
```
f-brinkmann, line 87: This linking won't work but linking to scipy is possible. You can probably search '`~scipy.' within pyfar for examples

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,336 @@
...
+    if (signal is None and sampling_rate is None) \
+            or (signal is not None and sampling_rate is not None):
+        raise ValueError('Either signal or sampling_rate must be none.')
+
+    # sampling frequency in Hz
+    fs = signal.sampling_rate if sampling_rate is None else sampling_rate
+
+    sos = _design_frequency_weighting_filter(fs, target_weighting,
+                                            n_frequencies, error_weighting,
+                                            **kwargs)
```
f-brinkmann: I think the indention should be increased by 1

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,336 @@
...
+    # apply or return the filter object
+    if signal is None:
+        return filt
+    else:
+        signal_filt = filt.process(signal)
+        return signal_filt
+
+
+def _design_frequency_weighting_filter(sampling_rate: float,
+                                      target_weighting: Literal["A", "C"]="A",
```
f-brinkmann: indention should be increased by 1 here and in the lines below

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,336 @@
...
+    fixed_zeros = [1, 1, 1, 1] if target_weighting == "A" else [1, 1]
+    zeros = np.concat([fixed_zeros, x[0:2]])
+    return (zeros, x[2:-1], x[-1])
+
+
+
+def _get_error_margins(f: np.ndarray):
+    """
+    Returns the upper and lower error margin in dB for every frequency in the
+    input array for the A and C weighting according to IEC/DIN-EN 61672-1.
```
f-brinkmann, line 337: Please add a reference to the forumla to make it easier to find.

tests/test_frequency_weighting.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def test_frequency_weighting_filter():
+    base_rates = np.array([48000, 44100, 16000])
+    test_rates = np.array([
+        base_rates,
+        base_rates / 2, base_rates * 2,
+        base_rates / 4, base_rates * 4,
+        base_rates / 8, base_rates * 8,
+    ]).flatten().tolist()
```
f-brinkmann: See above - better use pytest.mark.parametrize

pyfar/dsp/filter/frequency_weighting.py
jf-som: why won't this work? It looks the same as done throughout pyfar. Looks to me like it's the ~ is only used to reference pyfar, whereas other modules don't use ~

pyfar/constants/frequency_weighting.py
jf-som: done

pyfar/constants/frequency_weighting.py
jf-som: added mention and reference

pyfar/constants/frequency_weighting.py
jf-som: done

pyfar/dsp/filter/frequency_weighting.py
jf-som: done

pyfar/dsp/filter/frequency_weighting.py
jf-som: done

pyfar/dsp/filter/frequency_weighting.py
jf-som: Clarified as requested.
I also added 2 notes using your syntax, to separate performance notes from accuracy notes.

pyfar/dsp/filter/frequency_weighting.py
jf-som: added defaults for multiple parameters

pyfar/dsp/filter/frequency_weighting.py
jf-som: done

pyfar/dsp/filter/frequency_weighting.py
jf-som: added reference to the table in the standard. There is no formula afaik.

pyfar/dsp/filter/frequency_weighting.py
jf-som: done

pyfar/constants/frequency_weighting.py
jf-som: It doesn't seem like the standard provides a range limit. The curves simply fall off towards zero towards zero and infinity. We could prohibit negative values, but technically speaking they should be correct and might be useful in the rare case of doing complex-valued FFT calculations with negative frequencies. Since f is always squared in the formulae, the sign of f doesn't matter.

pyfar/constants/frequency_weighting.py
jf-som: The same question will arise later for the `levels` module, were this function will be used. I was thinking the level functions will have some kind of "bands" parameter that will be octave, third or broadband/single/None (to be named), where the user decides between whether to calculate the broadband level or in bands. num_fractions is a bit strange here because there are really only 2 options, 1 or 3. For the levels module, there will be a third option (fullband), that is hard to express as a num_fractions integer. We could make it 0 or None, but both are not really intuitive to me. And it's generally unusual to use an integer for a fixed number of options, for which Enums/Literals are made for.

So perhaps we would use Literal options for "broadband", "octave" and "third" for the levels-functions, and then accordingly only "octave" and "third" for this function.
We should discuss this perhaps.

pyfar/dsp/filter/frequency_weighting.py
f-brinkmann: Ah sorry, my bad.

tests/test_frequency_weighting.py
jf-som: changed it to pytest

tests/test_frequency_weighting.py
jf-som: true. Changed it to parametrize.

pyfar/constants/frequency_weighting.py
jf-som: Would you put the nominal frequencies in the FrequencyData.frequencies then? I think that would be a bit unusual given that FrequencyData is normally used for actual spectra

pyfar/constants/frequency_weighting.py
jf-som: changed it here

tests/test_frequency_weighting.py
jf-som: done

pyfar/dsp/filter/frequency_weighting.py
ahms5: fyi: the ~ makes it short, like [least_squares](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.least_squares.html#scipy.optimize.least_squares) without it it would be [scipy.optimize.least_squares](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.least_squares.html#scipy.optimize.least_squares)

pyfar/constants/frequency_weighting.py
f-brinkmann: > Would you put the nominal frequencies in the FrequencyData.frequencies then?

Yes, thats the what FrequencyData is intended for: Spectra that cannot be converted to the time domain, e.g., because of missing data or non-linearly spaced frequencies.

jf-som: alright then, implemented the change.

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,147 @@
...
+    bands: str ('octave' or 'third')
+        Whether to use octave bands or third bands.
+
+    frequency_range: (float, float)
+        A range of what nominal center frequencies to include. The lowest
+        defined center frequency is 10 Hz and the highest is 20 kHz.
+
+    Returns
+    -------
+    weights_with_nominals: NDArray
```
ahms5: would it make sence to return one object in db and the other not?
simular to air attenuation calculation. same for above

and it it FrequencyData now

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,147 @@
...
+        band_weights = all_weights
+    else:
+        raise ValueError("Allowed literals for bands are 'octave' and 'third'")
+
+    mask = (nominals >= frequency_range[0]) & (nominals <= frequency_range[1])
+    nominals_in_range = nominals[mask]
+    weights_in_range = band_weights[mask]
+
+    if len(nominals_in_range) == 0:
+        raise ValueError("Frequency range must include at least one value " \
```
ahms5: nice, how about an error if the upper limit is e.g. 40k Hz or 1Hz?

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,147 @@
...
+    comment = f"Level corrections for {weighting} weighting according to" \
+               "IEC 61672-1"
+    return pyfar.FrequencyData(weights, frequencies, comment)
+
+
+# Constants for nominal band corrections, taken from table 3 in IEC 61672-1.
+# This is somewhat redundant with dsp.filter.fractional_octaves, but
+# implementing different standards. Here, IEC 61672-1 defines bands as low as
+# 10 Hz, which is not part of fractional_octaves.
+_NOMINAL_THIRDBAND_FREQUENCIES = np.array([
```
ahms5: even if they are not public, maybe we can document them in the same style as the constants in the init file. so each one separated.

and maybe we collect all constants at the top of the file

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,147 @@
+"""File containing all frequency weighting functions."""
+from typing import Literal, Union
+import numpy as np
+import pyfar
+
+# Constants for the weighting curve formulas.
+# See appendix E in IEC 61672-1.
+_F_1 = 20.598997057618316
```
ahms5, line 8: same here:
even if they are not public, maybe we can document them in the same style as the constants in the init file. so each one separated.

ahms5 (changes requested):
I've just checked the constants for now. Thank you for implementing them — they look very professional. I made some suggestions; feel free to argue against them. I just reviewd API and doc stuff.

pyfar/constants/frequency_weighting.py
f-brinkmann: @jf-som we decided to leave this as suggested by you :)

pyfar/constants/frequency_weighting.py
ahms5: please have a look at https://github.com/pyfar/pyfar/issues/815, and feel free to suggest changes.

pyfar/constants/frequency_weighting.py
jf-som: fixed the comment.
Regarding dB vs linear, I don't think it has much value to return both. I expect the user to either need level corrections or linear factors, but not both. The standard also only defines them in dB.
We could add a `db=True` parameter for convenience, which makes the returned FrequencyData contain linear values instead of decibels if `False`. Though in that case, should it convert to pressure or energy units?

Perhaps it would be easier if pyfar had `db_to_pressure()`, `db_to_energy()` and their inverses as convenience functions that accept FrequencyData/TimeData(/Signal) objects, so users could easily do such conversions themselves rather than adding it to each function internally?

pyfar/constants/frequency_weighting.py
jf-som: Giving each constant its own docstring comment would result in multiple very similar comments, so since these constants are private anyway, I thought it was more clear to arrange them as a block with a comment explaining the whole block of constants.

Collecting them at the top vs close to where they are used is both fine for me. 

Perhaps we can get one or two more opinions from others for both suggestions?

f-brinkmann: I would be fine with moving them up but keep them the way they are otherwise.

pyfar/constants/frequency_weighting.py
f-brinkmann: Having a function to convert from decibels sounds good. A different possibility would be to ad something like an `inverse` flag to `pyfar.dsp.decibels`.

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,203 @@
...
+
+    Returns
+    -------
+    weights_with_nominals: FrequencyData
+        The correction values in dB for the specific frequency weighting
+        over their nominal frequencies as a Frequency Data object.
+
+    Examples
+    --------
+    Plot the band correction weights.
```
f-brinkmann:
```suggestion
    Plot the band correction levels.
```

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,203 @@
...
+                label="A weighting in third bands")
+        plt.plot(weights_C.frequencies, weights_C.freq[0], "--",
+                c=(0.5, 0.5, 0.5, 0.5))
+        plt.plot(weights_C.frequencies, weights_C.freq[0], "go",
+                label="C weighting in octave bands")
+        plt.legend()
+        ticks = [10, 30, 100, 300, 1000, 3000, 10000]
+        plt.semilogx()
+        plt.xticks(ticks, ticks)
+        plt.grid()
```
f-brinkmann: This would be shorter if using `pf.plot.freq`. You could also set the ticks to `weights_C.frequencies` to get octave ticks.

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,370 @@
...
+
+    Parameters
+    ----------
+    signal : Signal, None
+        The signal to be filtered. Pass ``None`` to create the filter without
+        applying it.
+
+    target_weighting: str, optional
+        Specifies which frequency weighting curve to approximate.
+        Must be either "A" or "C". The default is "A".
```
f-brinkmann: Sorry, I missed that before. We have the convention to highlight code in the docstring:
```suggestion
        Must be either ``"A"`` or ``"C"``. The default is ``"A"``.
```
Variables (as in the notes above) should be italic (which they are due to the single ` quotes)

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,370 @@
...
+
+    n_frequencies: int, optional
+        At how many frequencies to evaluate the filter coefficients during
+        optimization. Less frequencies means faster iterations, but
+        potentially worse results. The evaluation frequencies are
+        logarithmically spaced between 10 Hz and the Nyquist frequency.
+        The default is 100.
```
f-brinkmann:
```suggestion
        The default is ``100``.
```

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,370 @@
...
+        specific frequency ranges. The function should take a float array as
+        argument, specifying the normalized frequencies between 0 and 1
+        (where 1 is the Nyquist frequency) to weight, and returns a float
+        array as output containing the weights.
+        The default value is None, in which case the errors of all
+        (logarithmically spaced) frequencies are equally weighted. This
+        usually leads to larger errors for higher frequencies. By passing
+        a function that emphasizes high frequencies, it is possible to reduce
+        this effect and get a filter potentially closer to the target curve.
+        Example: `error_weighting=lambda nf: 100**nf`. This example often
```
f-brinkmann:
```suggestion
        Example: ``error_weighting=lambda nf: 100**nf``. This example often
```

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,370 @@
...
+        leads to better results for typical sampling rates, but much worse
+        for very high rates. The default is `None`.
```
f-brinkmann:
```suggestion
        for very high rates. The default is ``None``.
```

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,370 @@
...
+
+    sampling_rate: float, conditionally optional
+        The sampling rate of the returned filter. The default is `None`.
```
f-brinkmann:
```suggestion
        The sampling rate of the returned filter. The default is ``None``.
```

tests/test_frequency_weighting.py
```diff
@@ -0,0 +1,111 @@
...
+@pytest.mark.parametrize("fs", [48000, 44100])
+@pytest.mark.parametrize("error_weighting", [lambda _: 1,
+                                             lambda nf: 100**nf])
+def test_frequency_weighting_filter_weighted(weighting, fs,
+                                    error_weighting):
+    filt = pffilt.frequency_weighting_filter(None, weighting, sampling_rate=fs,
+                                             error_weighting=error_weighting)
+    is_class_1, _, _ = pffilt.frequency_weighting. \
+        _check_filter(filt.sampling_rate, filt.coefficients[0], weighting)
+    assert is_class_1
```
f-brinkmann: This is not really checking if `error_weighting` is working - which is a little hard in general. A simple way to test something might be, to remove the parameterization for `error_weighting` and check if the resulting filter coefficients (or IR) differ (in certain frequency ranges)

jf-som: True. I expanded the testing to make it more strict and clear, using something akin to your suggested approach

pyfar/constants/frequency_weighting.py
jf-som: I just tried that, but it seems using pf.plot.freq() is not shorter after all, or does not even work.
The calculated weights are in dB, but pf.plot requires pressure values. I could do `db=False`, but then it still takes the absolute values, which breaks the graph. Perhaps it could be done with the `mode="real"` from #817 later, but might still be a bit hacky. 
So I'll leave it with plt.plot() and only adjust the ticks and added axis labels.

pyfar/constants/frequency_weighting.py
jf-som: I moved them up now. If I understand correctly, this will be replaced by a dedicated function for nominal frequency bands later anyway.

pyfar/constants/frequency_weighting.py
jf-som: These values seem a bit arbitrary to me. I don't think we need to fire errors, since the code won't break from far-away limits, but simply return only bands in range.
Perhaps instead a warning would be helpful if the range would include the next third-band that isn't defined in the norm, so if the limits are outside of (8 Hz, 25000 Hz)?

pyfar/constants/frequency_weighting.py
jf-som: An inverse flag for dsp.decibels is also an option. I don't really know how I would name a dedicated inverse function, so an inverse parameter might be easiest.
Either way, I think it best to return the dB values only, as that's how the standard defines them.

pyfar/constants/frequency_weighting.py
jf-som: reverted it to numpy arrays after today's discussion about dB values in FrequencyData (#825).

pyfar/constants/frequency_weighting.py
jf-som: reverted it to numpy arrays after today's discussion about dB values in FrequencyData (#825).

pyfar/constants/frequency_weighting.py
f-brinkmann: We decided to leave this as is for now.

pyfar/constants/frequency_weighting.py
f-brinkmann: we decided that only returning dB vales in a numpy array is fine.

pyfar/constants/frequency_weighting.py
jf-som: I think after today's meeting this discussion should go to #824. If possible, #824 should probably update the nominal frequency logic in `constants/frequency_weighting.py` once it rolls out, to to use the same logic everywhere.

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,200 @@
...
+    --------
+    Plot the band correction levels.
+
+    .. plot::
+
+        import pyfar as pf
+        import matplotlib.pyplot as plt
+
+        range = (10, 20000)
+        nominals_third, weights_A = pf.constants. \
```
hoyer-a: Code using `\` isn't properly displayed in the docs.

Probably better to put it on one line and use `# noqaE501`

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,200 @@
...
+        plt.plot(nominals_octave, weights_C, "go",
+                label="C weighting in octave bands")
+        plt.legend()
+        ticks = nominals_octave[::2].astype(int)
+        plt.semilogx()
+        plt.xticks(ticks, ticks)
+        plt.xlabel("f in Hz")
+        plt.ylabel("Corrections in dB")
+        plt.grid()
+    """
```
hoyer-a:
```suggestion
    """  # noqa E501
```

hoyer-a (commented):
Found a little bug in the docs, otherwise this looks really good to me.

pyfar/constants/frequency_weighting.py
jf-som: Good catch! Fixed it now.

pyfar/constants/frequency_weighting.py
jf-som: Thanks showing me the method, I didn't know the `noqa` trick. Implemented it now.

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,198 @@
...
+_F_2 = 107.65264864304629
+_F_3 = 737.8622307362901
+_F_4 = 12194.217147998012
+_A_1000 = -2
+_C_1000 = -0.062
+
+# Constants for nominal band corrections, taken from table 3 in IEC 61672-1.
+# Could be replaced/combined with a dedicated frequency_bands_nominal()
+# function later.
+_NOMINAL_THIRDBAND_FREQUENCIES = np.array([
```
ahms5: this can be replaced after #824 is merged.

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,198 @@
...
+    # formula (E.1) in the standard
+    bracket_term = (_F_4**2 * f**2) / ((f**2 + _F_1**2) * (f**2 + _F_4**2))
+    return 10 * np.log10(bracket_term**2) - _C_1000
+
+
+def frequency_weighting_curve(weighting: Literal["A", "C"],
+                              frequencies: list[float]):
+    """
+    Calculates the level correction in dB of a frequency component when using
+    the A or C weighting defined in IEC 61672-1.
```
ahms5: can we add a reference to the standard?

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,198 @@
...
+
```
ahms5, line 63: would it make sence to add the equation in the docstring?

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,198 @@
...
+        raise ValueError("Allowed literals for weighting are 'A' and 'C'")
+
+    if bands == "octave":
+        nominals = _NOMINAL_THIRDBAND_FREQUENCIES[2::3]
+        band_weights = all_weights[2::3]
+    elif bands == "third":
+        nominals =_NOMINAL_THIRDBAND_FREQUENCIES
+        band_weights = all_weights
+    else:
+        raise ValueError("Allowed literals for bands are 'octave' and 'third'")
```
ahms5: if we here use the new method from pr #824 we could simplify this, maybe we should than use the same input num_fraction here. Similar to #824

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,198 @@
...
+    Examples
+    --------
+    Plot the band correction levels.
+
+    .. plot::
+
+        import pyfar as pf
+        import matplotlib.pyplot as plt
+
+        range = (10, 20000)
```
ahms5: maybe use sth else than range here, due to the build in method.

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,370 @@
...
+        n_frequencies=100,
+        error_weighting: Optional[Callable[[
+            np.ndarray], np.ndarray]] = None,
+        sampling_rate: Optional[float] = None,
+        **kwargs,
+        ):
+    """
+    Create and apply an A or C weighting filter.
+    The created SOS filter approximates the A or C weighting defined
+    in IEC/DIN-EN 61672-1.
```
ahms5: maybe add reference here

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,370 @@
...
+from pyfar.constants.frequency_weighting import _F_1, _F_2, _F_3, _F_4
+
+
+_A_WEIGHTING_ZEROS = [0, 0, 0, 0]
+_A_WEIGHTING_POLES = [_F_1, _F_1, _F_4, _F_4, _F_2, _F_3]
+_C_WEIGHTING_ZEROS = [0, 0]
+_C_WEIGHTING_POLES = [_F_1, _F_1, _F_4, _F_4]
+
+
+def frequency_weighting_filter(
```
ahms5, line 21:
```suggestion
def frequency_weighting(
```
do we need filter here? it is already in the filter submodule

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,370 @@
...
+        signal,
+        target_weighting: Literal["A", "C"]="A",
+        n_frequencies=100,
+        error_weighting: Optional[Callable[[
+            np.ndarray], np.ndarray]] = None,
+        sampling_rate: Optional[float] = None,
+        **kwargs,
+        ):
+    """
+    Create and apply an A or C weighting filter.
```
ahms5, line 31:
```suggestion
    Create and apply an A or C weighting filter.
    
```

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,370 @@
...
+        specific frequency ranges. The function should take a float array as
+        argument, specifying the normalized frequencies between 0 and 1
+        (where 1 is the Nyquist frequency) to weight, and returns a float
+        array as output containing the weights.
+        The default value is None, in which case the errors of all
+        (logarithmically spaced) frequencies are equally weighted. This
+        usually leads to larger errors for higher frequencies. By passing
+        a function that emphasizes high frequencies, it is possible to reduce
+        this effect and get a filter potentially closer to the target curve.
+        Example: ``error_weighting=lambda nf: 100**nf``. This example often
```
ahms5: maybe shift the example as a new example below the current one. There might be a good example to demonstrate what you explan here

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,370 @@
...
+        leads to better results for typical sampling rates, but much worse
+        for very high rates. The default is ``None``.
+
+    sampling_rate: float, conditionally optional
```
ahms5, line 82: i like the ``conditionally optional``

tests/test_frequency_weighting.py
```diff
@@ -0,0 +1,148 @@
+from numpy import testing as npt
```
ahms5, line 1: could be maybe put the tests for constants and dsp.filter module in separate files?

tests/test_frequency_weighting.py
```diff
@@ -0,0 +1,148 @@
...
+            frequency_weighting_band_corrections(weighting, bands, freq_range)
+    calculated_weights = pyfar.constants. \
+            frequency_weighting_curve(weighting, nominals)
+
+    assert isinstance(iec_weights, np.ndarray)
+    assert isinstance(calculated_weights, np.ndarray)
+
+    # maybe there is a better way of testing. tolerance is 0.2, because
+    # we need 0.1 for rounding the weight and then larger error
+    # due to the difference in nominal vs exact frequency being evaluated
```
ahms5, line 26: I would personally test them independ of each other. first frequency_weighting_band_corrections and check against what I except, even if this is just vaules from the tablen.

same for frequency_weighting_curve just pick a couple of values and check, against manual, maybe additioanlly with frequency_weighting_band_corrections, what you are doing here.

ahms5 (changes requested):
Thank you, I'm sorry I still have some comments^^

pyfar/dsp/filter/frequency_weighting.py
mberz, line 21: I would argue that the naming is misleading. The filter operation does not weight/scale frequencies, but levels (with frequency dependent weights.)
I thing something like a `level_weighting(_filter)` would be more appropriate here.

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,370 @@
...
+_A_WEIGHTING_ZEROS = [0, 0, 0, 0]
+_A_WEIGHTING_POLES = [_F_1, _F_1, _F_4, _F_4, _F_2, _F_3]
+_C_WEIGHTING_ZEROS = [0, 0]
+_C_WEIGHTING_POLES = [_F_1, _F_1, _F_4, _F_4]
+
+
+def frequency_weighting_filter(
+        signal,
+        target_weighting: Literal["A", "C"]="A",
+        n_frequencies=100,
```
mberz: in `Signal` & `FrequencyData` we use `n_bins` instead of `n_frequencies`, which could be considered here.

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,370 @@
...
+        weighting_filter = pf.dsp.filter.frequency_weighting_filter(
+            None, "A", sampling_rate=noise.sampling_rate)
+        weighted_noise = weighting_filter.process(noise)
+
+        pf.plot.freq(noise, label="unweighted noise")
+        pf.plot.freq(weighting_filter.impulse_response(10000),
+                    label="weighting filter")
+        pf.plot.freq(weighted_noise, label="A weighted noise")
+        plt.ylim(-45, 5)
+        plt.legend()
```
mberz: Please see example formatting in other functions

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,370 @@
...
+    if signal is None:
+        return filt
+    else:
+        signal_filt = filt.process(signal)
+        return signal_filt
+
+
+def _design_frequency_weighting_filter(sampling_rate: float,
+                                       target_weighting: Literal["A", "C"]="A",
+                                       n_frequencies=100,
```
mberz: see comment above.

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -0,0 +1,370 @@
...
+    mags = np.abs(freq_resp)
+    mags_dB = 10 * np.log10(mags**2)
+    mags_target = pfc.frequency_weighting_curve(weighting, test_freqs)
+    mag_diffs = mags_dB - mags_target
+
+    upper, lower = _get_error_margins(test_freqs)
+    is_class_1 = np.all(mag_diffs < upper) and np.all(mag_diffs >= lower)
+    max_diff = np.max(np.abs(mag_diffs))
+    mean_diff = np.mean(np.abs(mag_diffs))
+    return is_class_1, max_diff, mean_diff
```
mberz, line 395: Maybe this function could also modified to allow user access to the error margins similar to the tolerance functions for fractional octave filters?

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,198 @@
...
+
+        plt.plot(plot_frequencies, weights_A, label="A weighting")
+        plt.plot(plot_frequencies, weights_C, label="C weighting")
+        plt.semilogx()
+        plt.xlabel("f in Hz")
+        plt.ylabel("Weights in dB")
+        ticks = [10, 30, 100, 300, 1000, 3000, 10000, 30000]
+        plt.xticks(ticks, ticks)
+        plt.grid()
+        plt.legend()
```
mberz: See formatting example code (https://pyfar--811.org.readthedocs.build/en/811/modules/pyfar.dsp.html#pyfar.dsp.find_impulse_response_delay)

pyfar/constants/frequency_weighting.py
mberz:
```suggestion
_NOMINAL_THIRD_OCTAVE_BAND_FREQUENCIES = np.array([
```
In english it's either a third octave or simply third, third band would be the third band not a third octave. ;)
I'll not mark this everywhere, but there are a number of occurences in this PR. (It's a very minor comment, indeed)

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,198 @@
...
+    ----------
+    weighting: str ('A' or 'C')
+        Which frequency weighting to use.
+
+    frequencies: array-like
+        The frequencies at which to evaluate the weighting curve.
+
+    Returns
+    -------
+    weights: NDArray
```
mberz:
```suggestion
    weights: numpy.ndarray[float]
```
I think something like this should render a link to the respecive numpy docs.
Again, there are a couple of occurences, where this could be changed below.

pyfar/dsp/filter/frequency_weighting.py
mberz: see general comment below

pyfar/dsp/filter/frequency_weighting.py
jf-som: changed as requested

pyfar/dsp/filter/frequency_weighting.py
jf-som: changed as requested

pyfar/dsp/filter/frequency_weighting.py
jf-som: changed as requested

pyfar/dsp/filter/frequency_weighting.py
jf-som: What for or in what way? We could edit the _get_error_margins() function and make it public, if that is what you have in mind. I don't really see the use case at the moment though.

pyfar/constants/frequency_weighting.py
jf-som: changed as requested

pyfar/constants/frequency_weighting.py
jf-som: changed as requested (only here though, not sure where else this happens)

pyfar/constants/frequency_weighting.py
jf-som: changed as requested

pyfar/constants/frequency_weighting.py
jf-som: mentioning the name of the standard like here is what I found in some other places in pyfar too. How would you suggest we reference it?

pyfar/constants/frequency_weighting.py
jf-som: true, changed it to f_range

pyfar/dsp/filter/frequency_weighting.py
jf-som: Added a code example and reworked the parameter description accordingly

pyfar/constants/frequency_weighting.py
jf-som: I wanted to do it, but the fractional octave function is not suitable since it still only goes down to 25 Hz whereas the frequency weighting standard requires nominal bands down to 10 Hz.
I'd have to do something hacky such as rounding the exact frequencies instead, which sounds like a poor choice for maintainers

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,192 @@
...
+        defined center frequency is 10 Hz and the highest is 20 kHz.
+
+    Returns
+    -------
+    nominal_frequencies: numpy.ndarray[float]
+        The nominal center frequencies included in the given range.
+
+    weights: numpy.ndarray[float]
+        The correction values in dB for the specific frequency weighting.
+
```
ahms5, line 150:
```suggestion
    References
    ----------
    .. [#] IEC 62672-1:2013 Electroacoustics - Sound level meters
             - Part 1: Specifications
```

pyfar/constants/frequency_weighting.py
```diff
@@ -0,0 +1,192 @@
...
+
+def frequency_weighting_band_corrections(
+    weighting: Literal["A", "C"],
+    bands: Literal["third", "octave"],
+    frequency_range: tuple[float, float],
+):
+    """
+    Returns the A or C frequency weighting band corrections as specified in
+    IEC 62672-1 for the given frequency range.
```
ahms5:
```suggestion
    Returns the standardized A or C frequency weighting band corrections.
    
    The implementations follows the IEC 62672-1 standard _[#].
```

pyfar/constants/frequency_weighting.py
jf-som: I don't think so, because it consists of multiple small formulas in the norm that are hard to read and aren't very informative to a user.

pyfar/constants/frequency_weighting.py
ahms5: the new method is desined to also return values from 10Hz, see https://github.com/pyfar/pyfar/pull/824/files#r2290390757
but it is not yet merged.

jf-som: Nevermind my previous comment, I was looking at the old fractional octaves function, not the new one from #824. I can adapt to the new function once it is merged.

pyfar/constants/frequency_weighting.py
jf-som: ah, I think I found what you mean. I added a References section to each public function comment with the reference

pyfar/constants/frequency_weighting.py
jf-som: It now uses num_fractions like #824

pyfar/constants/frequency_weighting.py
jf-som: When adapting to #824 I removed the frequencies array and replaced the two correction value arrays by dictionaries, which should be more intuitive to read.

PR #816, opened 2025-07-25
Improve docs for `fractional_octave_bands`

f-brinkmann:

- Mention IEC standard in docs
- Add second example to show that filter are in accordance with standard


ahms5 (commented):
Nice, thank you. Would it make sense to also add a test where we check if the filter is within the tolerance of the standard? Otherwise approved.

f-brinkmann:
> Would it make sense to also add a test where we check if the filter is within the tolerance of the standard? 

@ahms5 yes, I thought about that as well, when doing the example. I'll start drafting something for discussion outside this pull.

ahms5 (approved):

pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -240,6 +248,41 @@ def fractional_octave_bands(
...
+
+        >>> import pyfar as pf
+        >>> import numpy as np
+        >>> import matplotlib.pyplot as plt
+        >>>
+        >>> # get octave filter
+        >>> center_frequency = 1000
+        >>> octave_filter = pf.dsp.filter.fractional_octave_bands(
+        >>>     pf.signals.impulse(2**12), 1, None,
+        ...     (center_frequency, center_frequency))
```
mberz: This is a good use-case for the impulse response method I would say. 
It should get rid of the arbitrarily chosen length.

pyfar/dsp/filter/fractional_octaves.py
```diff
@@ -171,6 +171,9 @@ def fractional_octave_bands(
         order=14):
     """Create and/or apply an energy preserving fractional octave filter bank.
 
+    The filters are in accordance with IEC 61260-1:2014 [#]_ for octave and
+    third band widths and the default filter order of 14 (see examples below).
+
```
mberz: I'm actually not sure if our implementation with hp filters if the band spans across Nyquist is compliant with the standard. Could you please check/confirm if it complies. Otherwise we'd need to add a comment/note here.

mberz (changes requested):

pyfar/dsp/filter/fractional_octaves.py
f-brinkmann: Used impulse response as suggested, but still a length parameter to get a nicer frequency resolution (only for the plot).

ahms5:
for just improving the doc, we usually tread it as hotfix, why not here?

f-brinkmann:
See above, we amped it up:
> We decided to wait with this until the tolerances are added to pyfar.constants. Once this is done, this pull request will also implement checking if filters are standard conform.

f-brinkmann:
closing this in favor of #829 and a clean commit history. Comments above have been acountet for.

PR #818, opened 2025-07-30
Add octave band tolerance to `pyfar.constants`

f-brinkmann:

@ahms suggested to test if our fractional octave filters are in agreement with the tolerance from the IEC standard. To start this, I suggest to add the corresponding tolerances to the constants module. Chages are

- add octave tolerances to constants module
- generate test data
- add tests for new functionality


pyfar/constants/constants.py
```diff
@@ -434,3 +435,116 @@ def density_of_air(
...
+        The type of fractional octave band. Must be either 'octave' or 'third'.
+    tolerance_class : Literal[1, 2]
+        The tolerance class as defined in the standard. Must be 1 or 2.
+
+    Returns
+    -------
+    tolerance : FrequencyData
+        Tolerance limits in dB. The lower tolerance is given in the first
+        channel (``tolerance.freq[0]``) and the upper tolerance in the second
+        (``tolerance.freq[1]``).
```
ahms5: even if the standard is just defininig it in dB, the pyfar style would be e.g. for plotting to also return it as an energy FrequencyData object.

ahms5 (commented):
Thank you for taking care!
Looks fine, I just have one comment, also related to https://github.com/pyfar/pyfar/pull/811#issuecomment-3131072596

pyfar/constants/constants.py
f-brinkmann: We will do it as suggested but not energetic.

pyfar/constants/constants.py
```diff
@@ -434,3 +435,116 @@ def density_of_air(
...
+        raise ValueError(f"bands is '{bands}' but must be 'octave' or 'third'")
+    if tolerance_class not in [1, 2]:
+        raise ValueError(
+            f"tolerance_class is {tolerance_class} but must be 1 or 2")
+
+    # constants from DIN EN 61260-1:2014-10, Sec. 5.10 and Tab. 1
+    # (define only first half due to symmetry)
+    G = 10**(3/10)  # octave ratio according to Eq. (1)
+    eps = 1e-6      # small offset to model discontinuities in tolerance curve
+    inf = 60e3      # dB value for infinity (number required for plots)
```
mberz: please do not re-define symbols/functions/terms which have a common meaning by arbitrary values ;)
eps and especially infinity are not what you name them.

f-brinkmann: absolutely. are those fine?

mberz:
I'm moving this to hold until #825 is resolved.

pyfar/constants/constants.py
f-brinkmann: as discussed in the weekly, returning db values and frequencies in separate numpy arrays

hoyer-a (approved):

pyfar/constants/constants.py
```diff
@@ -434,3 +435,115 @@ def density_of_air(
...
+    Calculate the tolerance limits for fractional octave band filters.
+
+    Calculation is in accordance with IEC 61260-1:2014 [#]_ (Section 5.10 and
+    Table 1).
+
+    .. note ::
+        The standard defines some lower tolerance limits to :math:`-\infty`,
+        which is inconvenient for plotting. The returned tolerance is -60000 dB
+        in these cases, which is below the smallest possible value approx.
+        -6000 dB (``20*np.log10(np.finfo(float).tiny)``).
```
mberz:
```suggestion
        ``20*np.log10(np.finfo(float).tiny`` `:math: \approx -6000 \mathrm{dB}`).
```

Maybe something like this? Or improve the wording

pyfar/constants/constants.py
```diff
@@ -434,3 +435,115 @@ def density_of_air(
...
+    Parameters
+    ----------
+    exact_center_frequency : float
+        The exact center frequency of the band filter in Hz (see
+        :py:func:`~pyfar.dsp.filter.fractional_octave_frequencies`).
+    num_fractions : Literal[1, 3]
+        The number of bands an octave is divided into. ``1`` for octave bands
+        and ``3`` for third octave bands.
+    tolerance_class : Literal[1, 2]
+        The tolerance class as defined in the standard. Must be 1 or 2.
```
mberz: Please keep this consistent
```suggestion
        The tolerance class as defined in the standard. Must be ``1`` or ``2``.
```

pyfar/constants/constants.py
```diff
@@ -434,3 +435,115 @@ def density_of_air(
...
+
+    Returns
+    -------
+    lower_tolerance : numpy array
+        Lower tolerance limits in dB of shape (19, ).
+    upper_tolerance : numpy array
+        Upper tolerance limits in dB of shape (19, ).
```
mberz, line 473: We could also condense this down into a single 2x19 array to make it a bit more compact as for example in the lower and upper cutoff frequencies returned by the fractional octave frequency freq. function.

mberz: https://pyfar--818.org.readthedocs.build/en/818/modules/pyfar.dsp.filter.html#pyfar.dsp.filter.fractional_octave_frequencies

f-brinkmann: I had that before but then found it more explicit and it produces code that is easier to read when checking for the tolerance, e.g.

`np.all(pf.dsp.decibel(octave_filter) < upper_tolerance)`

instead of 

`np.all(pf.dsp.decibel(octave_filter) < tolerance[1])`

I'd vote to keep it like this, but have no strong feelings about it.

pyfar/constants/constants.py
f-brinkmann: Yes that looks better. I slightly modified it to make all numbers appear in the same font.

pyfar/constants/constants.py
f-brinkmann: done

pyfar/constants/constants.py
mberz: @ahms5 commented on #824 to change return two variables instead of an array, so let's discuss. I also do not have a strong opinion on it, but would prefer if we keep it consisten if possible.

PR #819, opened 2025-07-31
Feature/horn transmission matrices

RoubenRehman:
In the same vein as the `new create_transmission_line` class, this adds support for conical horns to the `TransmissionMatrix` class according to Lampton, 1978 Eq. 5-18.

#### Parameters
- **a, b:** Distances from the horn's narrow (a) and wide (b) end to the virtual point, where the cone's walls would converge
- **Omega ($\Omega$):** Area constant. The horn's cross-sectional area follows $A(l) = \Omega l^2 \quad \forall a\leq l \leq b$
- **k:** Wave number
- **medium_impedance:** The medium impedance. Crucially **not** the characteristic impedance, as due to the varying cross-section, I found relating the medium impedance to it correctly before calling the function to be too error-prone.

##### Medium Impedance
As explained above, I chose to take in the medium impedance and not the characteristic impedance. This is different from the current implementation of the `create_transmission_line` function. If these two should be consistent with each other, some other solution has to be found.

##### Propagation direction
In the standard form (as presented e.g. in Lampton), the horn matrix expresses $p_a$ and $u_a$ at the narrow end in terms of $p_b$ and $u_b$ at the wider end, so it models propagation through the horn "in reverse" (from wide to narrow). If one wants to model propagation from narrow to wide, `a` and `b` have to be switched when calling the function. I find this slightly unintuitive, but kept it like this for now as it corresponds to the standard definition in literature. 

##### Tests
I added tests for the proposed function. However, I checked the reciprocal case by interchanging `a` and `b` when calling and comparing that to the inverse of the desired matrix as these two cases have to be equal. This required changing `_compare_tmat_vs_abcd` to use `np.allclose` instead of `np.all`.

### DIFF
--- a/pyfar/classes/transmission_matrix.py
+++ b/pyfar/classes/transmission_matrix.py
@@ -876,15 +876,14 @@ class TransmissionMatrix(FrequencyData):
             >>> import numpy as np
             >>> # Horn parameters
             >>> a = 0.1
-            >>> b = 0.2
-            >>> Omega = 0.4
+            >>> b = 1.7
+            >>> Omega = 2.4
             >>> frequencies = np.linspace(20, 20e3, 1000)
             >>> omega = 2 * np.pi * frequencies
             >>> k = pf.FrequencyData(omega / pf.constants.reference_speed_of_sound, frequencies)
             >>> Z0 = pf.constants.reference_air_impedance
-            >>> direction = 'forwards'
             >>> # Create the transmission matrix
-            >>> T = pf.TransmissionMatrix.create_conical_horn_section(a, b, Omega, k, Z0, direction)
+            >>> T = pf.TransmissionMatrix.create_conical_horn(a, b, Omega, k, Z0)
             >>> # Plot the transmission matrix
             >>> pf.plot.freq(T.input_impedance(np.inf))
             
--- a/pyfar/classes/transmission_matrix.py
+++ b/pyfar/classes/transmission_matrix.py
@@ -870,6 +870,8 @@ class TransmissionMatrix(FrequencyData):
 
         Example
         -------
+        Arbitrarily sized lossless conical horn section.
+        
         .. plot::
 
             >>> import pyfar as pf
@@ -884,9 +886,10 @@ class TransmissionMatrix(FrequencyData):
             >>> Z0 = pf.constants.reference_air_impedance
             >>> direction = 'forwards'
             >>> # Create the transmission matrix
-            >>> T = conical_horn_section(a, b, Omega, k, Z0, direction)
+            >>> T = pf.TransmissionMatrix.create_conical_horn_section(a, b, Omega, k, Z0, direction)
             >>> # Plot the transmission matrix
             >>> pf.plot.freq(T.input_impedance(np.inf))
+            
         """
         if not isinstance(a, Number):
             raise TypeError("The input a must be a number.")
--- a/pyfar/classes/transmission_matrix.py
+++ b/pyfar/classes/transmission_matrix.py
@@ -831,7 +831,6 @@ class TransmissionMatrix(FrequencyData):
             Omega: Number,
             k: FrequencyData,
             medium_impedance: Number | FrequencyData,
-            direction: str = 'backwards'
             ) -> TransmissionMatrix:
         r"""Create a transmission matrix representing a conical horn.
 
@@ -861,9 +860,6 @@ class TransmissionMatrix(FrequencyData):
             Wave number.
         medium_impedance : float | FrequencyData
             The impedance of the medium filling the horn.
-        direction : str, optional
-            Direction of propagation. Must be either 'backwards' (from b to a)
-            or 'forwards' (from a to b). Default is 'backwards'.
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +825,117 @@ def create_transmission_line(
...
+            Distance from the horn's narrow end to the virtual apex of the underlying cone.
+            This is not the physical length of the horn but the extrapolated length to the point
+            where the conical walls would converge.
+        b : float
+            Distance from the horn's wide end to the same virtual apex of the cone.
+            The difference (b - a) gives the actual physical length of the horn segment.
+        Omega : float
+            Area constant, such that the horn's cross-sectional area at any length
+            :math:`a\leq l \leq b` is given by :math:`S(l) = \Omega l^2`.
+        k : np.ndarray
```
ahms5, line 859:
```suggestion
        k : float | FrequencyData
```

pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +825,117 @@ def create_transmission_line(
         return TransmissionMatrix.from_abcd(
             A, B, C, D, kl.frequencies)
 
+    @staticmethod
+    def create_conical_horn(a: Number,
```
ahms5, line 829: as already discussed, maybe we could replace ``a`` and ``b`` by sth more intuitive and interanlly calculate a and b based on this. e.g. S_0 or r_0, the radius or surface at the beginning of the horn. and e.g. a parameter for the direction of the wave. 
and l being the length from S_0.

pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +825,117 @@ def create_transmission_line(
...
+        Parameters
+        ----------
+        a : float
+            Distance from the horn's narrow end to the virtual apex of the underlying cone.
+            This is not the physical length of the horn but the extrapolated length to the point
+            where the conical walls would converge.
+        b : float
+            Distance from the horn's wide end to the same virtual apex of the cone.
+            The difference (b - a) gives the actual physical length of the horn segment.
+        Omega : float
```
ahms5, line 856: maybe call it sth else than Omega, since it could be confusing with the angular frequency.

pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +825,117 @@ def create_transmission_line(
...
+            Area constant, such that the horn's cross-sectional area at any length
+            :math:`a\leq l \leq b` is given by :math:`S(l) = \Omega l^2`.
+        k : np.ndarray
+            Wave number.
+        medium_impedance : float | FrequencyData
```
ahms5, line 861: maybe as a default value it could be the impedance of air:
pf.constants.reference_air_impedance.

pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +825,117 @@ def create_transmission_line(
...
+    @staticmethod
+    def create_conical_horn(a: Number,
+            b: Number,
+            Omega: Number,
+            k: FrequencyData,
+            medium_impedance: Number | FrequencyData,
+            ) -> TransmissionMatrix:
+        r"""Create a transmission matrix representing a conical horn.
+
+        The transmission matrix is calculated based on the starting point :math:`a`,
```
ahms5, line 837: maybe we could also add a sketch with from the measrues of the horn like here: https://pyfar--819.org.readthedocs.build/en/819/classes/pyfar.transmission_matrix.html#pyfar.classes.transmission_matrix.TransmissionMatrix.input_impedance

tests/test_transmission_matrix.py
```diff
@@ -328,6 +328,202 @@ def test_create_transmission_line_frequency_matching():
         ValueError, match="The frequencies do not match"):
         TransmissionMatrix.create_transmission_line(kl, Z)
 
+def test_create_conical_horn_number():
+    """Test `create_transmission_line` with impedance as number."""
```
ahms5, line 332:
```suggestion
    """Test `create_transmission_line` with impedance as number with the wave propaging into both directions."""
```
or sth like this

tests/test_transmission_matrix.py
```diff
@@ -328,6 +328,202 @@ def test_create_transmission_line_frequency_matching():
...
+    b = 0.5
+    Omega = 0.4
+    
+    tmat_backwards = TransmissionMatrix.create_conical_horn(a, b, Omega, k, Z)
+    tmat_forwards = TransmissionMatrix.create_conical_horn(b, a, Omega, k, Z)
+
+    A = b/a * np.cos(k.freq*(b-a)) - 1/(k.freq*a) * np.sin(k.freq*(b-a))
+    B = 1j * Z / (a*b*Omega) * np.sin(k.freq*(b-a))
+    C = 1j * Omega / (k.freq*k.freq*Z) * ((1 + k.freq*k.freq*a*b) * np.sin(k.freq*(b-a)) - k.freq*(b-a)*np.cos(k.freq*(b-a)))
+    D = a/b * np.cos(k.freq*(b-a)) + 1/(k.freq*b) * np.sin(k.freq*(b-a))
```
ahms5, line 346:
```suggestion
    A_desired = 3
    B_desired = 7+1j
    C_desired = 3
    D_desired = 42
```

maybe for testing it would be better to calcualte it for inputs

tests/test_transmission_matrix.py
```diff
@@ -328,6 +328,202 @@ def test_create_transmission_line_frequency_matching():
...
+    b = 0.5
+    Omega = 0.4
+    
+    tmat_backwards = TransmissionMatrix.create_conical_horn(a, b, Omega, k, Z)
+    tmat_forwards = TransmissionMatrix.create_conical_horn(b, a, Omega, k, Z)
+
+    A = b/a * np.cos(k.freq*(b-a)) - 1/(k.freq*a) * np.sin(k.freq*(b-a))
+    B = 1j * Z.freq / (a*b*Omega) * np.sin(k.freq*(b-a))
+    C = 1j * Omega / (k.freq*k.freq*Z.freq) * ((1 + k.freq*k.freq*a*b) * np.sin(k.freq*(b-a)) - k.freq*(b-a)*np.cos(k.freq*(b-a)))
+    D = a/b * np.cos(k.freq*(b-a)) + 1/(k.freq*b) * np.sin(k.freq*(b-a))
```
ahms5, line 375: same here, see above

tests/test_transmission_matrix.py
```diff
@@ -328,6 +328,202 @@ def test_create_transmission_line_frequency_matching():
...
+    b = 0.5
+    Omega = 0.4
+
+    tmat_backwards = TransmissionMatrix.create_conical_horn(a, b, Omega, k, Z)
+    tmat_forwards = TransmissionMatrix.create_conical_horn(b, a, Omega, k, Z)
+
+    A = b/a * np.cos(k.freq*(b-a)) - 1/(k.freq*a) * np.sin(k.freq*(b-a))
+    B = 1j * Z.freq / (a*b*Omega) * np.sin(k.freq*(b-a))
+    C = 1j * Omega / (k.freq*k.freq*Z.freq) * ((1 + k.freq*k.freq*a*b) * np.sin(k.freq*(b-a)) - k.freq*(b-a)*np.cos(k.freq*(b-a)))
+    D = a/b * np.cos(k.freq*(b-a)) + 1/(k.freq*b) * np.sin(k.freq*(b-a))
```
ahms5, line 404: same here, see above.

ahms5 (changes requested):
Thank you for implementing it. It already looks very helpfull. I just some suggestions about the input parameters.

PR #823, opened 2025-08-07
Fix issue 822: Creating matrix from ABCD doesn't work with scalar values

### LINKED ISSUE #822: [BUG] Creating transmission matrices with Numbers instead of FrequencyData Obejcts for singular frequencies doesn't work
### Error Descritpion
When looking at the [Pyfar documentation](file:///C:/Users/rehman/sciebo/Doktorstudium/Programming/pyfar/docs/_build/html/classes/pyfar.transmission_matrix.html#pyfar.classes.transmission_matrix.TransmissionMatrix.from_abcd) for `TransmissionMatrix.from_abcd`, the function signature allows all parameters to be of type `double`.

Image

When trying to create a matrix like this however, the following error is thrown:

```
ValueError                                Traceback (most recent call last)
Cell In[3], [line 1](vscode-notebook-cell:?execution_count=3&line=1)
----> [1](vscode-notebook-cell:?execution_count=3&line=1) t = pf.TransmissionMatrix.from_abcd(A, B, C, D, fr)

File c:\...\pyfar\classes\transmission_matrix.py:215, in TransmissionMatrix.from_abcd(cls, A, B, C, D, frequencies)
    212 order = np.append(order, data.ndim-1)
    213 data = np.transpose(data, order)
--> [215](file:///C:/Users/rehman/sciebo/Doktorstudium/HAPPAA/Simulations/Occlusion%20Effect%20Simulation/.venv/Lib/site-packages/pyfar/classes/transmission_matrix.py:215) return cls(data, frequencies)

File c:\...\pyfar\classes\transmission_matrix.py:92, in TransmissionMatrix.__init__(self, data, frequencies, comment)
     90 n_dim = len(shape)
     91 if n_dim < 3 or shape[-3] != 2 or shape[-2] != 2:
---> [92](file:///C:/.../pyfar/classes/transmission_matrix.py:92)     raise ValueError("'data' must have a shape like "
     93                      "(..., 2, 2, n_bins), e.g. (2, 2, 100).")
     95 super().__init__(data, frequencies, comment)

ValueError: 'data' must have a shape like (..., 2, 2, n_bins), e.g. (2, 2, 100).
```

### Steps to reproduce
Run the following code:

```python
import pyfar as pf

A = 1
B = 1
C = 1
D = 1
fr = 1

t = pf.TransmissionMatrix.from_abcd(A, B, C, D, fr)
```

### Version
Tested with pyfar version `0.7.1` and `0.7.3` on Python 3.12.3

RoubenRehman:

Closes #822 

Inside `from_abcd` the code now checks for instances of type `Number` and converts them to 1D np-arrays using:
```python
A, B, C, D = [np.atleast_1d(np.asarray(param)) if isinstance(param, Number) else param for param in (A, B, C, D)]
```

I've also added tests that test for the correct creation of matrices with all accepted input parameter types as this was missing previously.

### DIFF
--- a/pyfar/classes/transmission_matrix.py
+++ b/pyfar/classes/transmission_matrix.py
@@ -136,7 +136,7 @@ class TransmissionMatrix(FrequencyData):
     @staticmethod
     def create_frequency_independent_abcd(A, B, C, D):
         """Create a T-matrix from frequency-independentA-, B-, C-, D-data.
-        
+
         Parameters
         ----------
         A : scalar (real or complex)
@@ -152,11 +152,6 @@ class TransmissionMatrix(FrequencyData):
         -------
         np.ndarray
             A 2x2 numpy array representing the transmission matrix.
-            
-        Examples
-        --------
-        >>> import pyfar as pf
-        >>> tmat = pf.TransmissionMatrix.create_frequency_independent_abcd(1, 2, 3, 4)
         """
         if not (
             (isinstance(A, Number) or isinstance(A, complex))
--- a/tests/test_transmission_matrix.py
+++ b/tests/test_transmission_matrix.py
@@ -277,7 +277,7 @@ def test_tmatrix_create_frequency_independent_abcd():
     assert isinstance(tmat, np.ndarray)
     assert tmat.shape == (2, 2)
     npt.assert_allclose(tmat, [[1, 2], [3, 4]], atol=1e-15)
-    
+
 @pytest.mark.parametrize("wrong_input", ["input", [1,2], (1,2), { 'A':1 }])
 def test_tmatrix_create_frequency_independent_abcd_wrong_input(wrong_input):
     """Test whether creation of frequency-independent ABCD matrix raises
@@ -285,13 +285,17 @@ def test_tmatrix_create_frequency_independent_abcd_wrong_input(wrong_input):
     """
     error_msg = "A, B, C, and D must be scalars"
     with pytest.raises(TypeError, match=error_msg):
-        TransmissionMatrix.create_frequency_independent_abcd(wrong_input, 2, 3, 4)
+        TransmissionMatrix.create_frequency_independent_abcd(
+            wrong_input, 2, 3, 4)
     with pytest.raises(TypeError, match=error_msg):
-        TransmissionMatrix.create_frequency_independent_abcd(1, wrong_input, 3, 4)
+        TransmissionMatrix.create_frequency_independent_abcd(
+            1, wrong_input, 3, 4)
     with pytest.raises(TypeError, match=error_msg):
-        TransmissionMatrix.create_frequency_independent_abcd(1, 2, wrong_input, 4)
+        TransmissionMatrix.create_frequency_independent_abcd(
+            1, 2, wrong_input, 4)
     with pytest.raises(TypeError, match=error_msg):
-        TransmissionMatrix.create_frequency_independent_abcd(1, 2, 3, wrong_input)
+        TransmissionMatrix.create_frequency_independent_abcd(
+            1, 2, 3, wrong_input)
 
 def test_tmatrix_create_identity(frequencies):
     """Test whether creation of identity matrix with frequencies."""


pyfar/classes/transmission_matrix.py
```diff
@@ -139,16 +140,17 @@ def from_abcd(cls, A, B, C, D, frequencies = None):
 
         Parameters
         ----------
-        A : FrequencyData, array_like, double
+        A : FrequencyData | array_like | scalar
```
ahms5: usually we use the ``,`` notation, is there any advantage if using ``|``?

pyfar/classes/transmission_matrix.py
```diff
@@ -191,10 +193,13 @@ def from_abcd(cls, A, B, C, D, frequencies = None):
             raise ValueError(
                         "If using FrequencyData objects, all matrix entries "
                         "A, B, C, D, must be FrequencyData objects.")
-
         if frequencies is None:
             raise ValueError("'frequencies' must be specified if not using "
                              "'FrequencyData' objects as input.")
+        
+        # Convert all parameters of type Number to 1d np arrays for further computation
+        A, B, C, D = [np.atleast_1d(np.asarray(param)) if isinstance(param, Number) else param for param in (A, B, C, D)]
```
ahms5: i think you can skip the if check, convert an array into an array would not change anything.

pyfar/classes/transmission_matrix.py
```diff
@@ -191,10 +193,13 @@ def from_abcd(cls, A, B, C, D, frequencies = None):
...
+        
         # broadcast shapes
         shape = np.broadcast_shapes(
             np.array(A).shape, np.array(B).shape,
```
ahms5:
```suggestion
            A.shape, B.shape,
```
then we can remove the arry conversion here

ahms5 (changes requested):
Thank you for your quick fix. 
I did not check the tests yet.

pyfar/classes/transmission_matrix.py
RoubenRehman: Thanks for pointing that out. Until now I've oriented my docstrings on the `create_transmission_line` docstring which has the `|` notation without realizing that every other function uses the `,` :D Will change this to `,` in every instance.

ahms5 (changes requested):
Thank you, Rouben. It seems there are a lot of changes from the formatter. Could you please remove the changes caused by the formatter? This would make it much simpler to review and would also prevent merge conflicts.

tests/test_transmission_matrix.py
```diff
@@ -73,6 +80,86 @@ def test_tmatrix_from_abcd_input_types(frequencies, A_list, A_np, A_FreqDat):
     _expect_error_abcd_same_type(A_np, A_FreqDat, A_np, A_np)
     _expect_error_abcd_same_type(A_FreqDat, A_np, A_np, A_np)
 
+def test_tmatrix_from_abcd_broadcasting(frequencies):
```
ahms5: this might be a good example for using parameters, see e.g. [here](https://github.com/pyfar/pyfar/blob/89f9ea1925756498ce751022fcab621139544c58/tests/test_coordinates_new.py#L16C14-L16C25). 
i would be also god practive to split the working and expected errors in separate tests (even we often missed it).
the output should also be tested.

tests/test_transmission_matrix.py
```diff
@@ -73,6 +80,86 @@ def test_tmatrix_from_abcd_input_types(frequencies, A_list, A_np, A_FreqDat):
...
+    D = 1
+    TransmissionMatrix.from_abcd(A, B, C, D, frequencies)
+
+    D = [1, 2]
+    _expect_error_abcd_broadcast(A, B, C, D, frequencies)
+
+    D = [1, 2, 3, 4]
+    _expect_error_abcd_broadcast(A, B, C, D, frequencies)
+
+def test_tmatrix_from_abcd_single_frequency():
```
ahms5: same here

tests/test_transmission_matrix.py
```diff
@@ -73,6 +80,86 @@ def test_tmatrix_from_abcd_input_types(frequencies, A_list, A_np, A_FreqDat):
...
+    TransmissionMatrix.from_abcd(A, B, C, D, frequencies)
+
+    D = [[1], [2]]
+    TransmissionMatrix.from_abcd(A, B, C, D, frequencies)
+
+    D = 1
+    TransmissionMatrix.from_abcd(A, B, C, D, frequencies)
+
+    D = [1, 2]
+    _expect_error_abcd_broadcast(A, B, C, D, frequencies)
```
ahms5: if it is just used here once, I would also paremtrize it, and remove the dedicated method ``_expect_error_abcd_broadcast``

tests/test_transmission_matrix.py
```diff
@@ -121,6 +208,18 @@ def abcd_data_3x3x1():
     D = A + 3
     tmat = TransmissionMatrix.from_abcd(A, B, C, D)
     return tmat, A, B, C, D
+@pytest.fixture(scope="module")
+def abcd_data_complex():
```
ahms5, line 219: where do we need it?

ahms5 (changes requested):
Thanks Rouben, looks fine, I just have a few recommendations for the testing.

tests/test_transmission_matrix.py
RoubenRehman: This is used in `def test_tmatrix_abd_entries` alongside the others

RoubenRehman:
@ahms5 thanks for the comments, I've integrated them now.  Regarding this section:

```python
@pytest.fixture(scope="module")
def abcd_data_complex():
```

I've added this, because I noticed issues with my initial implementation when handling complex numbers. The class should be able to do this (and is), but it was previously not tested at all.

RoubenRehman:
> Thank you for these efforts! I have a conceptual question. What is the idea behing putting functionality in `_tmat_from_abcd` in a separate function rather than directly `from_abcd`? Methods such as `create_shunt_admittance` could then call `from_abcd`. Please correct me if I understand something wrong here.

@sikersten my initial proposal was to extract the `tmat` logic from the member functions into `from_abcd`, since functions like `create_shunt_admittance` call `from_abcd` anyways. But in the friday meeting, where this was discussed, the consensus was to not do that, because it'd be weird for a constructor function like `from_abcd` to **not** return a class instance. By extracing the logic into its own function, `from_abcd` continues to always return a `TransmissionMatrix` instance.

pyfar/classes/transmission_matrix.py
```diff
@@ -132,21 +133,34 @@ def from_tmatrix(cls, data, frequencies, comment = ""):
         """
         return cls(data, frequencies, comment)
 
+    def _tmat_from_abcd(A, B, C, D):
+        """Create a T-matrix from A-, B-, C-, D-data."""
```
sikersten: Maybe something like?
```suggestion
    def _tmat_from_scalar_abcd(A, B, C, D):
        """Create a array representing a frequeny-independent T-matrix from A-, B-, C-, D-data."""
```

sikersten: Alternatively, what to you think of making this function public and calling it something like `create_frequency_independent`?
I fully agree that the `from_ ` functions should not return something else than a TransmissionMatrix object. This could be here circumvented by callling it `create_(...)`.

sikersten (commented):
Thanks for this clarification, @RoubenRehman, and sorry that I forgot! With this in mind, the functionality looks fine!
I would only suggest some naming/documentation for making the distinction between the functions more clear.
Alternatively, you may consider making it a public function (see below). Please let me know if you would like to follow this idea within this PR, or if we should implement this separately after merging this one.

pyfar/classes/transmission_matrix.py
RoubenRehman: Thanks @sikersten for your review! I agree with the idea of making the function public. Currently, the `create_` functions allow for frequency independent `tmat` returns, but don't allow to use arbitrary matrices. The `from_` methods allow arbitrary matrices, but require explicit frequencies. What do you think about making this new function a public method too and naming it `create_frequency_independent_abcd`?

sikersten: Yes, sounds good!

pyfar/classes/transmission_matrix.py
```diff
@@ -132,21 +133,52 @@ def from_tmatrix(cls, data, frequencies, comment = ""):
         """
         return cls(data, frequencies, comment)
 
+    @staticmethod
+    def create_frequency_independent_abcd(A, B, C, D):
+        """Create a T-matrix from frequency-independentA-, B-, C-, D-data.
```
sikersten, line 138:
```suggestion
        """Create a T-matrix from frequency-independent A-, B-, C-, D-data.
```

Issue #731, opened 2024-12-06
Potential breaking changes for pyfar 1.0.0


f-brinkmann:
## Description

these breaking changes could be implemented at some point

- [ ] Order of input arguments for fractional filter banks `pyfar.dsp.filter.fractional_octave_bands` gets sampling rate not at the end, which is the case everywhere else.

ahms5:
maybe also the Orientation class, #639

sikersten:
Replace type checks with `isinstance(...)` by adding type hints to function declaration and corresponding automatic input checks.

mberz:
I think Simon is referring to libraries such as pydantic.

mberz:
I've removed the mention of revising the Orientations class in the description, since we introduced the Rotation class in #904 without breaking changes.

PR #826, opened 2025-08-08
Add impulse_repsonse-method to GammatoneBands

hoyer-a:
Add `impulse_response()` and `minimum_impulse_response_length()` methods to `GammatoneBands`.

`impulse_response()` returns the impulse reponses with length of the longest filter / lowest band (similar to how channels are handled in `Filter.impulse_reponse()`)

For now, two Signal objects for real and imaginary part are returned, see #731.

### DIFF
--- a/pyfar/dsp/filter/gammatone.py
+++ b/pyfar/dsp/filter/gammatone.py
@@ -527,7 +527,7 @@ class GammatoneBands():
             The impulse response of the imaginary part of the filterbank of
             ``cshape = (:py:func:`~n_bands`, )``.
         """
-        # set or check impulse_response length
+        # set or check minimum impulse_response length
         minimum_impulse_response_length = \
             int(np.max(self.minimum_impulse_response_length(unit='samples')))
 
@@ -575,7 +575,7 @@ class GammatoneBands():
 
         # create a FilterSOS object
         SOS_filt = pf.FilterSOS(sos_sections, self.sampling_rate)
-        # get the minimum impulse response length
+        # get the minimum impulse response length per band / filter-channel
         estimated_length = \
             SOS_filt.minimum_impulse_response_length(unit=unit,
                                                      tolerance=tolerance)


mberz (commented):
Thanks for implementing this, I have a fundamental conceptual comment for this though:
I think it would make more sense to derive GammatoneBands from Filter instead of copying the implementation here. Are there any required changes to the original implementation from FilterSOS to the implementation here?

tests/test_filter_gammatone.py
```diff
@@ -206,3 +206,43 @@ def test_erb_frequencies_assertions():
...
+def test_impulse_response():
+    """Test the impulse response of GammatoneBands."""
+    GFB = pf.dsp.filter.GammatoneBands([0, 22050])
+    irs = GFB.impulse_response()
+    ir_real, ir_imag = irs
+
+    assert isinstance(irs, tuple)
+    assert len(irs) == 2
+    assert isinstance(ir_real, pf.Signal)
+    assert isinstance(ir_imag, pf.Signal)
```
f-brinkmann: This is stricter and maybe more future proof
```suggestion
    assert type(ir_real) == pf.Signal
    assert type(ir_imag) == pf.Signal
```

tests/test_filter_gammatone.py
```diff
@@ -206,3 +206,43 @@ def test_erb_frequencies_assertions():
...
+    assert ir_real.cshape == (GFB.n_bands, 1)
```
f-brinkmann, line 221: Also add this for ir_imag?

tests/test_filter_gammatone.py
```diff
@@ -206,3 +206,43 @@ def test_erb_frequencies_assertions():
...
+            match="n_samples should be at least as long as the filter"):
+        GFB.impulse_response(n_samples=100)
+
+
+def test_impulse_response_length():
+    """Test the minimum impulse response length."""
+    GFB = pf.dsp.filter.GammatoneBands((100, 200))
+    coefficients = GFB.coefficients
+    sos_coeffs = np.tile([[1, 0, 0, 1, -coefficients[0], 0]], (4, 1))
+    sos_coeffs[3, 0] = GFB.normalizations[0]
```
f-brinkmann: This is actually more elegant than the solution in `minimum_impulse_response_length` and could be used here as well. I think checking values computed using identical coefficients is not necessary here. I would rather manually check the length of a few bands for plausibility and tests against hard coded values. E.g. if a minimum length of 2000 samples is returned and the IR looks plausible test `min_length[some_index] == 2000`. Apart from that I would test the shape.

f-brinkmann (changes requested):
Thanks for the implementation, I only had small comments.

tests/test_filter_gammatone.py
hoyer-a: The idea was to just check if `GFB.minimum_impulse_response_length()` is working correctly. But yeah, it's more or less comparing the code in the class to itself and therefore probably not necessary.

Since internally the FilterSOS class is used for calculation, could we rely on the filter-class tests and not test results here again? Otherwise I can add checks against some hard-coded values and at the very least, for shape.

f-brinkmann: Yes true. Maybe checking values is completely unnecessary and checking type and shape is enough.

f-brinkmann (approved):
Thanks, its now fine for me.

Issue #825, opened 2025-08-08
Handling of dB values in pyfar


f-brinkmann:
We started to discuss how we should handle dB values that could be returned as `TimeData` or `FrequencyData` objects. I thought it might be good to open an issue to collect possible solutions and information

## What do we want to achieve

- Consitency within pyfar
- Avoid obvious user errors (to some degree)
- Keep flexibility for specific use cases
- We do not want to make that no user errors can happen

## Possible solutions

1. Add this information to `TimeData` and `FrequencyData` using decorators if the data is in dB. General information about this can be found [here](https://refactoring.guru/design-patterns/decorator), along with a [code example](https://gist.github.com/pingelit/8026c2263c6f6f07a301dd58d7a70e32).
2. Inheritance, by having a private dB-class and new classes, e.g., `TimeData` and TimeDataDB` classes.
3. As 2, but not adding new classes, e.g., the information is in `TimeData` and `FrequencyData`.
4. Always return time/frequency-dependend values in `TimeData` and `FrequencyData`. They can contain any values, its up to the users to cope with that.
5. As 4 but add the information to the comment to help the user figure out. This would be backwards incompatable because of (b).
6. We always use array likes instead of `TimeData` or `FrequencyData` to return non pressure values. 
7. Zero bits given - don't change anything.

Notes on the solutions

## Note a: Effort, backwards compatability, and effect

- Solutions 1-3 break backwards compatability because old data that might be saved to disc, does not have this functionality. They could be combined/extended for general units handling
- Solutions 4-6 break backwards compatibility because we have to change return values of current functions
- Solution 7 is backwards compatible but dirty.
- Solutions 1-3 require most changes (see note c). They would solve the dB problem, but there might be others (see note d)
- Solutions 4-6 require less changes (see note b). They would be consistent but shift the problem to the user.

## Note b: Current inconsistencies 

We are currently inconsitent about this. These functions currently return dB values or other things as `FrequencyData`:

- dB values `constants.air_attenuation`
- group delay in `signals.magnitude_spectrum_weighted_sweep`

Places where we return frequency/time depeneden values as numpy array. They were at least partly written before we had `FrequencyData`:

- `dsp.decibel`
- `dsp.group_delay`
- `dsp.phase`
- `dsp.sepctrogram` (special case, since its time-frequency data)
- `dsp.wrap_to_2pi` (also gets array like)

## Note c: Required changes if accounting for dB values

[...]

PR #827, opened 2025-08-11
BUG: return decible air attenuation as numpy array

ahms5:

kind of followed from the discussion in #825
- no decible values are allowed in FrequencyData

- return decible values ``alpha`` as a numpy array instead of FrequencyData

### DIFF
--- a/pyfar/constants/constants.py
+++ b/pyfar/constants/constants.py
@@ -30,7 +30,7 @@ def air_attenuation(
 
     Returns
     -------
-    alpha : :py:class:`~pyfar.FrequencyData`
+    alpha : np.ndarray[float]
         Pure tone air attenuation coefficient in decibels per meter for
         atmospheric absorption.
     m : :py:class:`~pyfar.FrequencyData`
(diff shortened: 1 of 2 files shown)


pyfar/constants/constants.py
```diff
@@ -30,7 +30,7 @@ def air_attenuation(
 
     Returns
     -------
-    alpha : :py:class:`~pyfar.FrequencyData`
+    alpha : np.ndarray[float]
         Pure tone air attenuation coefficient in decibels per meter for
         atmospheric absorption.
```
f-brinkmann, line 36: since its not frequency data anymore, maybe add a description of the shape to indicate frequencies are in the last dimension.

pyfar/constants/constants.py
```diff
@@ -147,17 +147,15 @@ def air_attenuation(
...
+    m = pf.FrequencyData(
+        air_attenuation / (10*np.log10(np.exp(1))), frequencies=frequencies)
 
     # calculate accuracy
     accuracy = _air_attenuation_accuracy(
-        h, temperature, atmospheric_pressure, frequencies, alpha.freq.shape)
+        h, temperature, atmospheric_pressure, frequencies, m.freq.shape)
 
-    return alpha, m, accuracy
+    return air_attenuation, m, accuracy
```
f-brinkmann: It can be a little confusing if the variable name does not equal the return parameter name from the docs, maybe harmonize this to ease maintanance?

pyfar/constants/constants.py
ahms5: i think it is mentioned above. we just talk about shape, so not specific to the objects.

mberz (approved):

hoyer-a (approved):

PR #828, opened 2025-08-15
fix bug in case caxis is not iteratable

f-brinkmann:

An error occurred in `dsp.average` when an integer value was passed for `caxis` and the corresponding axis had only one channel.

- force iterable to avoid the error
- expand tests

### DIFF
--- a/tests/test_dsp_average.py
+++ b/tests/test_dsp_average.py
@@ -106,10 +106,7 @@ def test_error_raises():
     with pytest.raises(ValueError,
                 match="mode must be 'linear', 'magnitude_zerophase',"):
         pf.dsp.average(signal, mode='invalid_mode')
-    with pytest.warns(UserWarning,
-                      match="Averaging one dimensional caxis"):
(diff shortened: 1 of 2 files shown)


hoyer-a (approved):
Thanks for fixing!

I think there's something wrong with our TU-server. Tests for pf.signals.files don't start and pyfar.org also doesn't load.

mberz (approved):

tests/test_dsp_average.py
```diff
@@ -122,3 +119,14 @@ def test_error_raises():
     with pytest.raises(ValueError,
                 match="'power' is not implemented for complex time signals."):
         pf.dsp.average(signal, 'power')
+
+
+@pytest.mark.parametrize('caxis', [2, (1, 2)])
```
f-brinkmann, line 124: > looks fine, We could add a test which would have failed before the fix, then this will never happen again.

@ahms This test is half new and half old. The new part is `caxis=2`, which raised an error before the fix.

ahms5 (approved):

Issue #835, opened 2025-08-27
One abstract parent class for all pyfar classes


ahms5:
> ## further ideas
> One could also go further and define a abstract parent class for all pyfar classen, this class forces all childs to implement certain methods such as ``broadcast_cshape``, ``broadcast_cshape`` and also io releated methods such as ``_encode`` and ``_decode`` and others. This would also make sure, that all our classes shares common properties, such as the cshape concept. 

 _Originally posted by @ahms5 in [#832](https://github.com/pyfar/pyfar/issues/832#issuecomment-3228055045)_

this would have some advantages:
- testing for pyfar object for io stuff could be simplified: https://github.com/pyfar/pyfar/blob/463f37ab6dfe9f1ea842562ab8f4cec97e5abfc5/pyfar/io/_codec.py#L386
- io: other pyfar packages could simply use this class and we dont need a reference in the core package https://github.com/pyfar/spharpy/issues/118
- make sure all our classes shares common properties, such as the cshape concept

sikersten:
# Suggestions

- Separate base classes for different aspects
- Harmonize Audio and Coordinates in a first step

## Pull Request structure

- Branch from develop for both implementation of base classes and application #850 
- Subbranch: Implementation
- One subsubbranch per base class 
- Subbranches for each new method
- Subbranches: one per application

## _PyfarBase(ABC) #836 
Implements the requirements for all pyfar classes.
Properties:
- comment

Methods:
- copy
- _encode
- _decode
- __eq__

## _PyfarMultichannel #837 
Implements equivalent functionality to numpy arrays.

Properties:
- cshape
- cdim
- csize

Methods
- reshape
- flatten
- transpose, T
- broadcast_cshape
- broadcast_cdim
- slicing

## _PyfarArithmetics #845 
Implements arithmetic operations for the objects.

- add, +
- substract, -
- multiply, *
- divide, /
- power, **
- _assert_match_for_arithmetics

# Ideas for new methods:
- abs
- sum
- resize
- swap_caxis
- squeeze
- Methods for objects containing complex-valued data.
  - conj
  - real
  - imag

h-chmeruk:
_PyfarBase is the abstract parent class for _Audio, Coordinates (Orientations, etc.). 
_NumpyLike, _PyfarArithmetics, _PyfarComplex are mixin classes. Proposal regarding dependencies:

class _Audio(_PyfarBase, _NumpyLike, _PyfarArithmetics, _PyfarComplex):
    ...

[...]

PR #836, opened 2025-08-28
_PyfarBase class

h-chmeruk:
As proposed in the [Issue #835](https://github.com/pyfar/pyfar/issues/835), here is a draft implementation of the _PyfarBase class. This class should be an abstract parent class for other pyfar classes such as _Audio and Coordinates. 

If you have any ideas about which properties or methods should also be included, please feel free to write them in the comments :)

### DIFF
--- a/tests/test_pyfar_base.py
+++ b/tests/test_pyfar_base.py
@@ -1,6 +1,5 @@
 from pyfar.classes._pyfar_base import _PyfarBase
 from unittest.mock import patch
-import pytest
 
 @patch.multiple(_PyfarBase, __abstractmethods__=set())
 def test_comment():


pyfar/classes/pyfar_base.py
sikersten, line 1: Maybe name it _pyfar_base.py to indicate it is private?

pyfar/classes/pyfar_base.py
```diff
@@ -0,0 +1,78 @@
+"""
+The following documents the pyfar base class.
```
sikersten: I would suggest implementing all private pyfar base classes in this file

pyfar/classes/pyfar_base.py
```diff
@@ -0,0 +1,78 @@
...
+
+from abc import ABC, abstractmethod
+from copy import deepcopy
+
+class _PyfarBase(ABC):
+    """
+    Abstract base class for Audio and Coordinates classes in pyfar.
+
+    This class defines the interface that Audio and Coordinates classes
+    must follow.
```
sikersten: I know that the plan is to first use this class for Audio and Coordinates. However, there may be different other classes in the future which use _PyfarBase, so it makes sense to implement these docs more general
```suggestion
    Abstract base class for classes in pyfar.

    This class defines the interface that pyfar classes
    must follow.
```

pyfar/classes/pyfar_base.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+
+    def __init__(self, comment=""):
+            self._comment = comment
+
+    @property
+    def comment(self):
+        """Get comment."""
+
+        return self._comment
```
sikersten: What about the setter?

pyfar/classes/pyfar_base.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @property
+    @abstractmethod
+    def cshape(self):
+        """
+        Subclasses must define a channel shape.
+
+        The channel shape gives the shape of the data excluding the last
+        dimension.
```
sikersten: I would suggest already providing the default doc string. The info regarding the last dimension does not hold for TransmissionMatrix.
```suggestion
        Return channel shape.
```

pyfar/classes/pyfar_base.py
```diff
@@ -0,0 +1,78 @@
...
+        pass
+
+    @property
+    @abstractmethod
+    def cdim(self):
+        """
+        Subclasses must define a channel dimension.
+
+        The channel dimension (`cdim`) gives the number of dimensions of the
+        audio data excluding the last dimension.
```
sikersten: I would suggest already providing the default doc string:
```suggestion
        Return channel dimension.

        The channel dimension (`cdim`) is the length of the channel
        shape (`cshape`) (e.g. ``self.cshape = (2, 3)``; ``self.cdim = 2``).
```

pyfar/classes/pyfar_base.py
```diff
@@ -0,0 +1,78 @@
...
+        """
+        pass
+
+    @property
+    @abstractmethod
+    def csize(self):
+        """
+        Subclasses must define a channel size.
+
+        The channel size gives the number of points stored in the object.
```
sikersten: I would suggest already providing the default doc string:
```suggestion
        Return channel size.

        The channel size is the total number of channels
        (e.g. ``self.cshape = (2, 3)``; ``self.csize = 6``)
```

pyfar/classes/pyfar_base.py
```diff
@@ -0,0 +1,78 @@
...
+        """
+        Subclasses must define a channel dimension.
+
+        The channel dimension (`cdim`) gives the number of dimensions of the
+        audio data excluding the last dimension.
+        """
+        pass
+
+    @property
+    @abstractmethod
```
sikersten: Does this property need to be abstract our can it already be implemented based on `cshape`?

pyfar/classes/pyfar_base.py
```diff
@@ -0,0 +1,78 @@
...
+        """
+        Subclasses must define a channel shape.
+
+        The channel shape gives the shape of the data excluding the last
+        dimension.
+        """
+        pass
+
+    @property
+    @abstractmethod
```
sikersten: Does this property need to be abstract our can it already be implemented based on `cshape`?

pyfar/classes/pyfar_base.py
sikersten: Changed my mind

pyfar/classes/pyfar_base.py
h-chmeruk: Thank you very much for your comments!

I also thought about implementing `cdim` and `csize` based on `cshape`, but I wasn't sure if this would work for every class that inherits from _PyfarBase (or _NumPyLike, because we now want to move `cshape`, `cdim`and `csize` in the _NumPyLike). 

But I'm almost certain that it can't work with `csize`, because `csize` needs information about all dimensions, and `cshape` doesn't provide the last dimension for the most classes.

pyfar/classes/_pyfar_base.py
```diff
@@ -0,0 +1,54 @@
...
+
+    @abstractmethod
+    def _decode(self):
+        """Decode an object based on its respective `_encode` counterpart."""
+        pass
+
+    @abstractmethod
+    def __eq__(self, other):
+        """Check for equality of two objects."""
+        pass
```
ahms5:
```suggestion
    def __eq__(self, other):
        """Check for equality of two objects."""
        return self._encode() == other._encode()
```

maybe sth like this could be a good default

pyfar/classes/_pyfar_base.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def copy(self):
+        """Return a deep copy of the object."""
+
+        return deepcopy(self)
+
+    @abstractmethod
+    def _encode(self):
+        """Return a dictionary for the encoding."""
+        pass
```
ahms5:
```suggestion
    def _encode(self):
        """Return a dictionary for the encoding."""
        return self.__dict__
```

could be a good default. Maybe you can also have a look on different pyfar classes, how they implemented ``_encode``

h-chmeruk: Thank you for the review!
I have already looked through the files of most of our classes and found that the function ‘_encode’ differs slightly for Signal (since we have set the domain of the signal to time) and for Orientations (since we first convert the object to a quaternion). For most classes, `return self.copy().__dict__` can surely be used.

pyfar/classes/pyfar_base.py
sikersten, line 1: This file needs to be removed, doesn't it?

pyfar/classes/__init__.py
```diff
@@ -5,6 +5,7 @@
 from . import filter
 from . import transmission_matrix
 from . import warnings
+from . import pyfar_base
```
sikersten: Like this?
```suggestion
from . import _pyfar_base
```

pyfar/classes/__init__.py
```diff
@@ -13,4 +14,6 @@
     'filter',
     'transmission_matrix',
     'warnings',
+    'pyfar_base',
```
sikersten: Like this?
```suggestion
    '_pyfar_base',
```

pyfar/classes/_pyfar_base.py
```diff
@@ -0,0 +1,54 @@
+"""
+The following documents the pyfar base class.
```
sikersten, line 2:
```suggestion
The following documents the pyfar base class.

It serves the purpose to implement minimum requirements for all pyfar classes.
```

pyfar/classes/_pyfar_base.py
sikersten: `self.__dict__`  or `self.copy().__dict__`?

pyfar/classes/_pyfar_base.py
sikersten: From Audio:
```suggestion
    @abstractmethod
    def __eq__(self, other):
        """Check for equality of two objects."""
        return not deepdiff.DeepDiff(self.__dict__, other.__dict__)
```

sikersten: From Filter:
```suggestion
    @abstractmethod
    def __eq__(self, other):
        """Check for equality of two objects."""
        return not deepdiff.DeepDiff(self, other)
```

sikersten (commented):
Looks already good! I changed this PR from draft to ready for review :)

pyfar/classes/_pyfar_base.py
h-chmeruk: Thank you for the suggestion!
 I'm not sure we really need a default for `__eq__`, since most pyfar classes have their own way to check objects equality. But perhaps we could use `not deepdiff.DeepDiff(self.__dict__, other.__dict__)`? It is used in _Audio. Or just  `not deepdiff.DeepDiff(self, other)`, which is used in the Filter class.

pyfar/classes/_pyfar_base.py
h-chmeruk: Until now, `self.copy().__dict__` has always been used in the code, although I don't see the purpose of it, since `__dict__` is only read here and cannot be changed.

pyfar/classes/pyfar_base.py
h-chmeruk: Oh, yes, my mistake, thank you!

pyfar/classes/_pyfar_base.py
sikersten: Then let's improve existing code here to `self.__dict__` :)

pyfar/classes/_pyfar_base.py
ahms5: I would also go for a default implementation, I guess both are doing the same, and we could remove the double implentation from the other classes

pyfar/classes/_pyfar_base.py
ahms5: imo self.__dict__  seems to be a good default, other methods can sill overwrite the method if they require sth specific.

pyfar/classes/_pyfar_base.py
ahms5, line 1: if we have one class per file, i would call the file ``_PyfarBase.py``.

tests/test_pyfar_base.py
ahms5, line 1: rename ``test__PyfarBase.yp``

pyfar/classes/_pyfar_base.py
```diff
@@ -0,0 +1,50 @@
...
+    def __init__(self, comment=""):
+            self._comment = comment
+
+    @property
+    def comment(self):
+        """Get comment."""
+        return self._comment
+
+    @comment.setter
+    def comment(self, comment):
```
ahms5:
```suggestion
    def comment(self, comment: str):
```

pyfar/classes/_pyfar_base.py
```diff
@@ -0,0 +1,50 @@
...
+
+class _PyfarBase(ABC):
+    """
+    Abstract base class for classes in pyfar.
+
+    This class defines the interface that pyfar classes
+    must follow.
+    """
+
+    def __init__(self, comment=""):
```
ahms5:
```suggestion
    def __init__(self, comment:str=""):
```

tests/test_pyfar_base.py
```diff
@@ -0,0 +1,45 @@
...
+    """Test the comment property."""
+
+    instance = Dummy("Dummy #1")
+    assert instance.comment == "Dummy #1"
+
+def test_comment_setter():
+    """Test the comment.setter."""
+
+    instance = Dummy()
+    instance.comment = "Dummy #2"
```
ahms5, line 15:
```suggestion
    assert instance.comment == ''
    instance.comment = "Dummy #2"
```
in this way we could check the default

ahms5 (changes requested):
thank you, just very small comments

tests/test_PyfarBase.py
```diff
@@ -0,0 +1,46 @@
+from pyfar.classes._PyfarBase import _PyfarBase
+
+class Dummy(_PyfarBase):
+    """
+    Minimal dummy class for testing _PyfarBase.
+    """
+
+    def _decode(self):
+        pass
```
mberz: I assume you're doing this to avoid the un-set abstract method.
However, this makes reading the tests a bit opaque to me, since below you're always testing the dummy and not `_PyfarBase`.

Instead you can patch the abstract methods using the snippet below, which makes reading the code easier in my opinion.

``` python

@patch.multiple(_AbstractClassImplementation, __abstractmethods__=set())
def test_abstract_class
```

pyfar/classes/__init__.py
```diff
@@ -5,6 +5,7 @@
 from . import filter
 from . import transmission_matrix
 from . import warnings
+from . import _PyfarBase
```
mberz: For file naming we usually follow the snake case convention, not capitalized camel case.

I would further suggest to collect the abstract base classes in a single file to make things a bit more compact and avoid jumping around between files when working on the base classes.

Since all base classes are not part of the public API, I would also suggest to remove the import here altogether.

``` suggestion
```

pyfar/classes/__init__.py
```diff
@@ -13,4 +14,6 @@
     'filter',
     'transmission_matrix',
     'warnings',
+    '_PyfarBase',
+
```
mberz:
```suggestion
```

mberz (changes requested):
Thanks for the work on this.
I have a couple of comments on the file/import structure and testing implementation.
If need be, we can discuss during the weekly.

tests/test_PyfarBase.py
h-chmeruk: Thank you for your review! 

I'm not really sure whether this dummy class is so bad, because in the end we won't be using the functionality of `_PyfarBase` directly, but with subclasses, so testing with a dummy class is actually more realistic than with the trick you suggest.

Also, I will definitely need a dummy class for `_PyfarMultichannel`, because this dummy class has to define `_data`. Without `_data`, there would be no test for the class at all. 

So personally, I would be consistent and if we have to use a dummy class to test one of the base classes, I would  also use dummy classes for other base classes.

But in my opinion, that is actually something we should discuss in the meeting :)

pyfar/classes/__init__.py
sikersten: Just to be sure: your comment on the case convention refers to the file naming, right? Not the name of the classes themselves, where  follow capitalized camel case (TimeData, FrequencyData, FilterFIR ...): https://pyfar.readthedocs.io/en/stable/api_reference.html#classes.

tests/test_PyfarBase.py
ahms5: The Dummy class is basically doing the same. It is also Mocking the Abstract class for testing.
Here you can find a Mock section in our [contribution guidelines](https://pyfar-gallery.readthedocs.io/en/latest/contribute/contribution_packages.html#fixtures). 

The advantage of using the ``unittest.mock`` is about readability and maintainability.

pyfar/classes/__init__.py
h-chmeruk: Am I understanding correctly that I should change the name of the `_PyfarBase` class files back to `_pyfar_base.py` and `test_pyfar_base.py`, and name all future ones according to the snake case?

mberz: Yes, and I would also add all future abstract base (like the multichannel, etc.) classes to the same file `_pyfar_base.py` to avoid fragmentation of related code.

pyfar/classes/__init__.py
```diff
@@ -13,4 +13,5 @@
     'filter',
     'transmission_matrix',
     'warnings',
+
```
ahms5, line 16:
```suggestion
```

tests/test_pyfar_base.py
```diff
@@ -0,0 +1,41 @@
+from pyfar.classes._pyfar_base import _PyfarBase
+from unittest.mock import patch
+import pytest
+
+@pytest.fixture(autouse=True)
+def _disable_abstract_methods():
+    """Automatically disable abstract methods for tests."""
+    with patch.multiple(_PyfarBase,__abstractmethods__=set()):
+            yield
```
mberz: I know this seems convenient, but please apply the patch to the test where it's required explicitly.
If additional tests are added to the file it may happen that this is not wanted or overlooked which can cause problems.

mberz (commented):
Largely good to go.
I very much would urge you to change the implicit handling of patching to a more explicit way. See comment below.

PR #839, opened 2025-09-04
Reader for .ita files and test functions for it

### LINKED ISSUE #206: Reader for *.ita files
## Description

For ease of compatibility we should have a reader function for *.ita files created using the ITA-Toolbox.

h-chmeruk:
#### Which issue(s) will be closed by this pull request?
Replaces #720
Closes #206

Here is a draft implementation of test functions for the .ita Reader. If you have any suggestions or comments regarding the test functions or the ` read_ita`  function itself, please do not hesitate to write them.


pyfar/io/io.py
```diff
@@ -962,3 +963,112 @@ def _read_comsol_get_headerline(filename):
...
+        weights = ita_coordinates.weights if ita_contains_weights else None
+    else:
+        weights = None
+
+    # import as cartesian coordinates: x, y, z
+    if hasattr(ita_coordinates, 'cart'):
+        if ita_coordinates.cart.size == 0:
+            return pf.Coordinates()
+        if ita_coordinates.cart.shape[-1] != 3:
+            raise ValueError(
```
h-chmeruk, line 1070: I also thought about implementing a function to test this ValueError, but I was unable to create an .ita file with an invalid number of dimensions (as there are enough exceptions for this case in MATLAB). 
If anyone knows how to test this, I would appreciate some help :)

ahms5: seems to be difficult, in this case you can skip the test, but we should keep the error handling here

pyfar/io/io.py
```diff
@@ -962,3 +963,112 @@ def _read_comsol_get_headerline(filename):
     is_complex = 'i' in line
     delimiter = ',' if ',' in line else None
     return header, is_complex, delimiter
+
+
+def read_ita(filename):
+    """Read a *.ita file from the ITA-toolbox.
```
sikersten: It may be nice to link to the toolbox here (adding it as reference below): https://www.ita-toolbox.org/

pyfar/io/io.py
```diff
@@ -962,3 +963,112 @@ def _read_comsol_get_headerline(filename):
...
+
+    Parameters
+    ----------
+    filename : str
+        The filename. The file extension must be *.ita.
+
+    Returns
+    -------
+    data : pyfar.Signal, pyfar.TimeData, pyfar.FrequencyData
+        The data contained in the *.ita file.
```
sikersten:
```suggestion
        The audio data contained in the *.ita file.
```

pyfar/io/io.py
sikersten:
```suggestion
    """Read an *.ita file created using the ITA-toolbox.
```

pyfar/io/io.py
```diff
@@ -962,3 +963,112 @@ def _read_comsol_get_headerline(filename):
...
+            data = pf.FrequencyData(
+                data_in, matlab_data.abscissa,
+                comment=comment)
+
+    metadata = _matlab_to_dict(matlab_data)
+
+    return data, object_coordinates, channel_coordinates, metadata
+
+
+def _to_coordinates(ita_coordinates):
```
sikersten:
```suggestion
def _ita_to_pyfar_coordinates(ita_coordinates):
```

pyfar/io/io.py
```diff
@@ -962,3 +963,112 @@ def _read_comsol_get_headerline(filename):
...
+    if hasattr(matlab_data, "channelCoordinates"):
+        channel_coordinates = _to_coordinates(matlab_data.channelCoordinates)
+    if hasattr(matlab_data, "objectCoordinates"):
+        object_coordinates = _to_coordinates(matlab_data.objectCoordinates)
+
+    data_in = np.ascontiguousarray((matlab_data.data.T).astype(float))
+
+    # checks if *.ita object is itaAudio or itaResult
+    if (hasattr(matlab_data, 'signalType')):  # itaAudio
+        fft_norm = 'none' if matlab_data.signalType == 'energy' else 'rms'
```
sikersten, line 1021: Maybe add a comment here, that this results from the ITA-toolbox logic?

pyfar/io/io.py
```diff
@@ -962,3 +963,112 @@ def _read_comsol_get_headerline(filename):
...
+    if hasattr(ita_coordinates, 'weights'):
+        ita_contains_weights = ita_coordinates.weights.size > 0
+        weights = ita_coordinates.weights if ita_contains_weights else None
+    else:
+        weights = None
+
+    # import as cartesian coordinates: x, y, z
+    if hasattr(ita_coordinates, 'cart'):
+        if ita_coordinates.cart.size == 0:
+            return pf.Coordinates()
```
sikersten: This object won't be of any use, right? So maybe return None instead? T

pyfar/io/io.py
```diff
@@ -962,3 +963,112 @@ def _read_comsol_get_headerline(filename):
...
+                "ITA coordinates must have three "
+                "columns at the last dimension.")
+        return pf.Coordinates(
+            ita_coordinates.cart[..., 0],
+            ita_coordinates.cart[..., 1],
+            ita_coordinates.cart[..., 2],
+            weights=weights)
+
+    # import as spherical coordinates: r, theta, phi
+    if hasattr(ita_coordinates, 'sph'):
```
sikersten: Is it possible that ita_coordinates has both attributes "cart" and "sph"? If not, this would need to be an elif connected to line 1035

pyfar/io/io.py
```diff
@@ -962,3 +963,112 @@ def _read_comsol_get_headerline(filename):
...
+            raise ValueError(
+                "ITA coordinates must have three "
+                "columns at the last dimension.")
+        return pf.Coordinates.from_spherical_colatitude(
+            ita_coordinates.sph[..., 1],
+            ita_coordinates.sph[..., 2],
+            ita_coordinates.sph[..., 0],
+            weights=weights)
+    else:
+        return pf.Coordinates()
```
sikersten: I would suggest returning None here, since this object won't be of any use

tests/test_io.py
```diff
@@ -710,6 +711,34 @@ def test_default_audio_subtype(default_audio_subtype_mock):
...
+    ('freq_itaResult.ita', 'signal'),
+    ('freq_itaResult_mult.ita', 'data'),
+    ('freq_itaResult_mult_ww.ita', 'data'),
+    ('time_itaResult.ita', 'data'),
+    ('time_itaResult_mult.ita', 'data')])
+def test_read_ita(filename, data_type):
+    """Tests correct exception raise for example *.ita files."""
+    filepath = os.path.join('tests', 'test_io_data', filename)
+    if (filename == 'freq_itaResult.ita' and data_type == 'signal'):
+        message = "The itaResult object can't contain a signal."
```
sikersten: This message is not defined in the function?

tests/test_io_read_ita_ahe.py
sikersten, line 1: I think this file should not be here

tests/test_io.py
sikersten, line 1: Please move the added content to test_io_read_ita.py

h-chmeruk: Thank you for your review! 
I just noticed that this file `tests/test_io.py` contains an outdated test function `test_read_ita(filename, data_type)` that we don't need. Therefore, I would simply remove this function from the file.

pyfar/io/io.py
h-chmeruk: I am not an expert on the ITA-Toolbox, but I think that ita_coordinates can have both attributes at the same time.

pyfar/io/io.py
```diff
@@ -962,3 +963,121 @@ def _read_comsol_get_headerline(filename):
...
+        return pf.Coordinates(
+            ita_coordinates.cart[..., 0],
+            ita_coordinates.cart[..., 1],
+            ita_coordinates.cart[..., 2],
+            weights=weights)
+
+    # import as spherical coordinates: r, theta, phi
+    if hasattr(ita_coordinates, 'sph'):
+        if ita_coordinates.cart.size == 0:
+            return pf.Coordinates()
```
f-brinkmann: See comment above from sikersten. I would suggest to return `None` here, too.

pyfar/io/io.py
```diff
@@ -962,3 +963,121 @@ def _read_comsol_get_headerline(filename):
...
+        if domain == 'time':
+            data = pf.TimeData(
+                data_in, matlab_data.abscissa,
+                comment=comment)
+        elif domain == 'freq':
+            data = pf.FrequencyData(
+                data_in, matlab_data.abscissa,
+                comment=comment)
+
+    metadata = _matlab_to_dict(matlab_data)
```
f-brinkmann: This return redundant information, e.g., `matlab_data.sampling_rate` and `matlab_data.data` that are already returned in the Signal. I would suggest to remove all data that is already returned elsewehre.

pyfar/io/io.py
```diff
@@ -962,3 +963,121 @@ def _read_comsol_get_headerline(filename):
...
+        The filename. The file extension must be `*.ita`.
+
+    Returns
+    -------
+    data : pyfar.Signal, pyfar.TimeData, pyfar.FrequencyData
+        The audio data contained in the `*.ita` file.
+    object_coordinates : pyfar.Coordinates
+        The object coordinates of the `*.ita` file.
+    channel_coordinates : pyfar.Coordinates
+        The channel coordinates of the `*.ita` file.
```
f-brinkmann: In case we follow sikerstens suggestion below, this should become
```suggestion
    object_coordinates : pyfar.Coordinates, None
        The object coordinates of the `*.ita` file if they exist, ``None`` otherwise.
    channel_coordinates : pyfar.Coordinates, None
        The channel coordinates of the `*.ita` file if they exist, ``None`` otherwise.
```

f-brinkmann: Maybe add a note on the difference of object and channel coordinates? Do we need both? Just asking because I'm not familiar with this.

pyfar/io/io.py
f-brinkmann: @mberz can you clarify? The files I saw, that contained both had identical data stored in different coordinate conventions, I think.

pyfar/io/io.py
```diff
@@ -962,3 +963,121 @@ def _read_comsol_get_headerline(filename):
...
+    else:
+        return None
+
+
+def _matlab_to_dict(mat_obj):
+    """Recursively transform MATLAB struct objects to nested dictionaries."""
+    mat_dict = {}
+    for fieldname in mat_obj._fieldnames:
+        elem = mat_obj.__dict__[fieldname]
+        if isinstance(elem, spio.matlab.mio5_params.mat_struct):
```
f-brinkmann: I think this line causes the deprecation errors and should be changed accordingly:

> DeprecationWarning: Please import `mat_struct` from the `scipy.io.matlab` namespace; the `scipy.io.matlab.mio5_params` namespace is deprecated and will be removed in SciPy 2.0.0.

f-brinkmann (changes requested):
Thanks! I found data that is returned twice. Apart from that its mostly small suggestions.

pyfar/io/io.py
```diff
@@ -962,3 +963,121 @@ def _read_comsol_get_headerline(filename):
     is_complex = 'i' in line
     delimiter = ',' if ',' in line else None
     return header, is_complex, delimiter
+
+
+def read_ita(filename):
+    """
+    Read an `*.ita` file created using the ITA-toolbox.
+
+    ITA-Tollbox is a MATLAB Toolbox for Acoustics [#]_.
```
ahms5:
```suggestion
    ITA-Toolbox is a MATLAB Toolbox for Acoustics [#]_.
```

pyfar/io/io.py
```diff
@@ -962,3 +963,121 @@ def _read_comsol_get_headerline(filename):
...
+                data_in, matlab_data.abscissa,
+                comment=comment)
+
+    metadata = _matlab_to_dict(matlab_data)
+
+    return data, object_coordinates, channel_coordinates, metadata
+
+
+def _ita_to_pyfar_coordinates(ita_coordinates):
+    if hasattr(ita_coordinates, 'weights'):
```
ahms5, line 1059: maybe add a short docstring here.

pyfar/io/io.py
ahms5: I can either be ``'cart'`` or ``'sph'``, similar to what we has before in our coordinates class.

pyfar/io/io.py
```diff
@@ -962,3 +963,121 @@ def _read_comsol_get_headerline(filename):
...
+            ita_coordinates.sph[..., 1],
+            ita_coordinates.sph[..., 2],
+            ita_coordinates.sph[..., 0],
+            weights=weights)
+    else:
+        return None
+
+
+def _matlab_to_dict(mat_obj):
+    """Recursively transform MATLAB struct objects to nested dictionaries."""
```
ahms5: maybe also add documation about input and output of the method

tests/test_io_data/read_ita/dirac_itaAudio.ita
ahms5, line 1: maybe move to ``tests/test_io_data/read_ita/*`` in this way it would be cleaner and easier to distinguish where it bolongs.

tests/test_io_read_ita.py
```diff
@@ -0,0 +1,86 @@
...
+    conversion of this data to pf.Coordinates.
+    """
+
+    filepath = os.path.join('tests', 'test_io_data', filename)
+    _, _, channel_coords, metadata = pf.io.read_ita(filepath)
+    metadata_channel_coords_reshaped = \
+        metadata["channelCoordinates"]["cart"].reshape(channel_coords.cartesian.shape)
+    assert np.array_equal(metadata_channel_coords_reshaped,
+                            channel_coords.cartesian) and \
+        np.array_equal(channel_coords.cartesian, channelCoordinates)
```
ahms5:
```suggestion
    assert np.array_equal(metadata_channel_coords_reshaped,
                            channel_coords.cartesian)
    assert np.array_equal(channel_coords.cartesian, channelCoordinates)
```

tests/test_io_read_ita.py
```diff
@@ -0,0 +1,86 @@
+import pyfar as pf
+import os
+import pytest
+import numpy as np
+
```
ahms5, line 5: is there any test about the metadata which is not part of other returned data?

pyfar/io/io.py
h-chmeruk: Thank you for your review!

I'm not quite sure what exactly I should do.
Are you suggesting including less redundant data in the `metadata`? For example, removing the `matlab_data.domain` or `matlab_data.sampling_rate` from the `metadata`? 

Wouldn't that be too confusing? Because the variable `data` stores an instance of `pf.Signal/pf.TimeData/pf.FrequencyData`, and in the variable `metadata` we only have the raw data for this instance, not the instance itself.

ahms5: Good point Fabian, since there is not straight forward solution for this, I would still leave it as it is.

pyfar/io/io.py
h-chmeruk: The ITA documentation in MATLAB contains some brief descriptions: 
```
channelCoordinates: coordinates for each channel (must be itaCoordinates)
objectCoordinates: the coordinates of the measured object
```
Would that be enough for us? I am also not very familiar with the purpose and functionality of `channelCoordinates` and `objectCoordinates`, so I would appreciate the opinion of someone who worked with this :)

tests/test_io_read_ita.py
h-chmeruk: Thank you for the review!

Regarding your question: No, not really. 
Which data from the `metadata` should I also test?

In the `metadata` dictionary there are also keys such as "classname", "classversion", "comment", "history", "fileName", "dateCreated", "dateModified", "dateSaved", "userData", "plotAxesProperties", "plotLineProperties", "userName", "channelNames", "channelUnits", "channelSensors", "channelUserData", "dataTypeOutput", but they are mostly empty in most files we are testing.

pyfar/io/io.py
f-brinkmann: Would it make sense to exclude at least the `matlab_data.data` from `metadata`? This bascially doubles the RAM requirements and might be annyoing if reading large files.

pyfar/io/io.py
mberz: Same concept as in sofa with listener and receiver.
One object can have multiple channels with relative positioning to the object.

pyfar/io/io.py
sikersten: +1 for @f-brinkmann's suggestion

PR #842, opened 2025-09-17
Feature/conical transmission line

RoubenRehman:
This PR adds a function `create_conical_horn` to `pf.TransmissionMatrix` that returns the matrix representation of a conical horn transmission line section as stated e.g. in *Lampton - 1978 - Transmission Matrices in Electroacoustics.pdf*.

The function takes the arguments:
- `S0: float` - The surface area of the narrow end
- `S1: float` - The surface area of the wider end
- `L: float` - The section's length
- `k: FrequencyData, scalar` - the wave number
- `medium_impedance: FrequencyData, scalar` - the impedance of the medium filling the horn
- `propagation_direction: str = {'forwards', 'backwards'}` - The direction of sound propagation through the horn

Distances from the horns virtual apex are calculated internally using a private function `_calculate_horn_geometry_parameters`.

#### It is important to note that `medium_impedance` is **not** the same as `characteristic_impedance` taken in by `create_transmission_line`:

The latter is referenced to the cross-sectional area of the transmission line, but the conical horn has no constant cross-section so these parameters have to be distinct.

### DIFF
--- a/pyfar/classes/transmission_matrix.py
+++ b/pyfar/classes/transmission_matrix.py
@@ -34,7 +34,6 @@ import numpy as np
 import numpy.testing as npt
 from pyfar.classes.audio import FrequencyData
 from pyfar.constants import reference_air_impedance, reference_speed_of_sound
-from numbers import Number
 
 
 class TransmissionMatrix(FrequencyData):
--- a/pyfar/classes/transmission_matrix.py
+++ b/pyfar/classes/transmission_matrix.py
@@ -941,6 +941,8 @@ class TransmissionMatrix(FrequencyData):
         propagation_direction: str = "expanding",
     ) -> np.ndarray | TransmissionMatrix:
         r"""Create a transmission matrix representing a conical horn.
+        It expresses sound pressure and volume velocity at one end of the horn
+        as a function of sound pressure and volume velocity at the other end.
 
         The transmission matrix is determined based on the surface
         area of the horn's narrow end, the surface area of the
@@ -953,19 +955,19 @@ class TransmissionMatrix(FrequencyData):
 
         .. math::
             \begin{bmatrix}
-                p_1\quad[Pa] \\
-                q_1\quad[m^3/s]
+            p_1\quad[Pa] \\
+            q_1\quad[m^3/s]
             \end{bmatrix}
             =
             \begin{bmatrix}
-                \frac{b}{a}\cos(kl) - \frac{1}{ka}\sin(kl) &
-                \frac{jZ_0}{ab\Omega}\sin(kl) \\[6pt]
-                \frac{j\Omega}{k^2Z_0}\left( (1 + k^2ab)\sin(kl) - kl\cos(kl) \right) &
-                \frac{a}{b}\cos(kl) - \frac{1}{kb}\sin(kl)
+            \frac{b}{a}\cos(kl) - \frac{1}{ka}\sin(kl) &
+            \frac{jZ_0}{ab\Omega}\sin(kl) \\[6pt]
+            \frac{j\Omega}{k^2Z_0}\left( (1 + k^2ab)\sin(kl) - kl\cos(kl) \right) &
+            \frac{a}{b}\cos(kl) - \frac{1}{kb}\sin(kl)
             \end{bmatrix}
             \begin{bmatrix}
-                p_2\quad[Pa] \\
-                q_2\quad[m^3/s]
+            p_2\quad[Pa] \\
+            q_2\quad[m^3/s]
             \end{bmatrix}
 
         Parameters
--- a/pyfar/classes/transmission_matrix.py
+++ b/pyfar/classes/transmission_matrix.py
@@ -893,7 +893,7 @@ class TransmissionMatrix(FrequencyData):
 
         Returns
         -------
-        tuple[float, float, float]
+        Omega, a, b: tuple[float, float, float]
             A tuple containing the area constant Omega, the distance a from the
             narrow end to the virtual apex of the cone, and the distance b from
             the wide end to the same virtual apex.
@@ -914,23 +914,11 @@ class TransmissionMatrix(FrequencyData):
 
         r0 = np.sqrt(S0 / np.pi)
         r1 = np.sqrt(S1 / np.pi)
-
-        d0 = 2 * r0
-        d1 = 2 * r1
-
-        a = r0 * (2 * L) / (d1 - d0)
+        
+        a = r0 * L / (r1 - r0)
         b = a + L
 
-        Omega0 = S0 / a**2
-        Omega1 = S1 / b**2
-
-        if not np.isclose(Omega0, Omega1, atol=1e-15):
-            raise ValueError(
-                f"S0/a² is not equal to S1/b²: {Omega0} != {Omega1}."
-                "This should never happen.",
-            )
-
-        Omega = Omega0  # or Omega1, they are equal
+        Omega = S0 / a**2   # equal to S1 / b**2
 
         return Omega, a, b
 
@@ -947,7 +935,7 @@ class TransmissionMatrix(FrequencyData):
 
         The transmission matrix is determined based on the surface
         area of the horn's narrow end, the surface area of the
-        horn's wide end, the horn's length, wave number, and
+        horn's wide end, the horn's length, wavenumber, and
         medium impedance.The surface areas are internally
         transformed into starting point :math:`a`,
         end point :math:`b`, and area constant :math:`\Omega`
@@ -972,7 +960,7 @@ class TransmissionMatrix(FrequencyData):
             The distance between the narrow and wide end of the horn,
             i.e. the horn's length.
         wave_number : FrequencyData, scalar
-            Wave number.
+            Wavenumber.
         medium_impedance : FrequencyData, scalar
             The impedance of the medium filling the horn. Default is
             ``pyfar.constants.reference_air_impedance``
@@ -1014,7 +1002,7 @@ class TransmissionMatrix(FrequencyData):
         if not (isinstance(wave_number, FrequencyData)
                 or isinstance(wave_number, Number)):
             raise TypeError(
-                "The wave number k must be a float, complex, "
+                "The wavenumber k must be a float, complex, "
                 "or FrequencyData object.",
             )
         elif isinstance(wave_number, FrequencyData):
@@ -1079,8 +1067,8 @@ class TransmissionMatrix(FrequencyData):
             / (a * b * Omega) * np.sin(wave_number * horn_length))
         C = (
             1j * Omega
-            / (wave_number * wave_number * medium_impedance)
-            * ((1 + wave_number * wave_number * a * b)
+            / (wave_number ** 2 * medium_impedance)
+            * ((1 + wave_number ** 2 * a * b)
             * np.sin(wave_number * horn_length) - wave_number * horn_length
             * np.cos(wave_number * horn_length))
         )
--- a/pyfar/classes/transmission_matrix.py
+++ b/pyfar/classes/transmission_matrix.py
@@ -884,16 +884,16 @@ class TransmissionMatrix(FrequencyData):
 
         Parameters
         ----------
-        area_narrow_end : float
+        area_narrow_end : Number
             Cross-sectional area at the narrow end of the horn.
-        area_wide_end : float
+        area_wide_end : Number
             Cross-sectional area at the wide end of the horn.
-        horn_length : float
+        horn_length : Number
             Length of the horn.
 
         Returns
         -------
-        Omega, a, b: tuple[float, float, float]
+        Omega, a, b: tuple[Number, Number, Number]
             A tuple containing the area constant Omega, the distance a from the
             narrow end to the virtual apex of the cone, and the distance b from
             the wide end to the same virtual apex.
@@ -939,7 +939,7 @@ class TransmissionMatrix(FrequencyData):
         wave_number: Number | FrequencyData,
         medium_impedance: Number | FrequencyData = reference_air_impedance,
         propagation_direction: str = "forwards",
-    ) -> TransmissionMatrix:
+    ) -> np.ndarray | TransmissionMatrix:
         r"""Create a transmission matrix representing a conical horn.
 
         The transmission matrix is determined based on the surface
@@ -961,11 +961,11 @@ class TransmissionMatrix(FrequencyData):
 
         Parameters
         ----------
-        area_narrow_end : float
+        area_narrow_end : Number
             The surface area of the horn's narrow end.
-        area_wide_end : float
+        area_wide_end : Number
             The surface area of the horn's wide end.
-        horn_length : float
+        horn_length : Number
             The distance between the narrow and wide end of the horn,
             i.e. the horn's length.
         wave_number : FrequencyData, scalar
@@ -1008,21 +1008,18 @@ class TransmissionMatrix(FrequencyData):
             >>> pf.plot.freq(T.input_impedance(np.inf))
 
         """
-        if not (isinstance(wave_number, FrequencyData)
-                or isinstance(wave_number, Number)):
+        if isinstance(wave_number, FrequencyData):
+            frequencies = wave_number.frequencies
+            wave_number = wave_number.freq
+            singular_frequency = False
+        elif isinstance(wave_number, complex) or isinstance(wave_number, Number):
+            frequencies = []
+            singular_frequency = True
+        else:
             raise TypeError(
                 "The wavenumber k must be a float, complex, "
                 "or FrequencyData object.",
             )
-        elif isinstance(wave_number, FrequencyData):
-            frequencies = wave_number.frequencies
-            wave_number = wave_number.freq
-        else:
-            if isinstance(wave_number, complex):
-                k_re = wave_number.real
-            else:
-                k_re = wave_number
-            frequencies = (k_re * reference_speed_of_sound) / (2 * np.pi)
         if isinstance(medium_impedance, FrequencyData):
             if not np.allclose(
                 medium_impedance.frequencies,
@@ -1084,7 +1081,10 @@ class TransmissionMatrix(FrequencyData):
         D = (a / b * np.cos(wave_number * horn_length)
             + 1 / (wave_number * b) * np.sin(wave_number * horn_length))
 
-        return TransmissionMatrix.from_abcd(A, B, C, D, frequencies)
+        if(singular_frequency == False):
+            return TransmissionMatrix.from_abcd(A, B, C, D, frequencies)
+        else:
+            return TransmissionMatrix.create_frequency_independent_abcd(A, B, C, D)
 
     def __repr__(self):
         """String representation of TransmissionMatrix class."""
--- a/pyfar/classes/transmission_matrix.py
+++ b/pyfar/classes/transmission_matrix.py
@@ -1014,6 +1014,7 @@ class TransmissionMatrix(FrequencyData):
             singular_frequency = False
         elif isinstance(wave_number, complex) \
             or isinstance(wave_number, Number):
+            k = wave_number
             frequencies = []
             singular_frequency = True
         else:
--- a/pyfar/classes/transmission_matrix.py
+++ b/pyfar/classes/transmission_matrix.py
@@ -912,10 +912,7 @@ class TransmissionMatrix(FrequencyData):
             or isinstance(horn_length, complex) or horn_length <= 0:
             raise ValueError("The input horn_length "
                              "must be a positive real number.")
-        if area_narrow_end > area_wide_end:
-            raise ValueError("area_narrow_end must be "
-                             "strictly smaller than area_wide_end.")
-        if area_narrow_end == area_wide_end:
+        if area_narrow_end >= area_wide_end:
             raise ValueError(
                 "For a conical horn area_narrow_end "
                 "must be strictly smaller than area_wide_end."
--- a/pyfar/classes/transmission_matrix.py
+++ b/pyfar/classes/transmission_matrix.py
@@ -952,12 +952,21 @@ class TransmissionMatrix(FrequencyData):
         matrix following Equation (5-18) of Reference [1]_:
 
         .. math::
-            T = \begin{bmatrix}
-                \frac{b}{a}cos(kl)-\frac{1}{ka}sin(kl) &
-                \frac{jZ_0}{ab\Omega} \sin{kl} \\
-                \frac{j\Omega}{k^2Z_0}\left( (1 + k^2ab) \sin{kl} -
-                kl \cos{kl} \right) & \frac{a}{b}cos(kl)-\frac{1}{kb}sin(kl)
-                \end{bmatrix}
+            \begin{bmatrix}
+                p_1\quad[Pa] \\
+                q_1\quad[m^3/s]
+            \end{bmatrix}
+            =
+            \begin{bmatrix}
+                \frac{b}{a}\cos(kl) - \frac{1}{ka}\sin(kl) &
+                \frac{jZ_0}{ab\Omega}\sin(kl) \\[6pt]
+                \frac{j\Omega}{k^2Z_0}\left( (1 + k^2ab)\sin(kl) - kl\cos(kl) \right) &
+                \frac{a}{b}\cos(kl) - \frac{1}{kb}\sin(kl)
+            \end{bmatrix}
+            \begin{bmatrix}
+                p_2\quad[Pa] \\
+                q_2\quad[m^3/s]
+            \end{bmatrix}
 
         Parameters
         ----------
--- a/pyfar/classes/transmission_matrix.py
+++ b/pyfar/classes/transmission_matrix.py
@@ -33,6 +33,8 @@ from numbers import Number
 import numpy as np
 import numpy.testing as npt
 from pyfar.classes.audio import FrequencyData
+from pyfar.constants import reference_air_impedance, reference_speed_of_sound
+from numbers import Number
 
 
 class TransmissionMatrix(FrequencyData):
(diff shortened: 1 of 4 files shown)


pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +854,210 @@ def create_transmission_line(
...
+            The surface area of the horn's wide end.
+        L : float
+            The distance between the narrow and wide end of the horn,
+            i.e. the horn's length.
+        k : FrequencyData, scalar
+            Wave number.
+        medium_impedance : FrequencyData, scalar
+            The impedance of the medium filling the horn. Default is
+            ``pyfar.constants.reference_air_impedance``
+        propagation_direction: str = {'forwards', 'backwards'}
```
h-chmeruk:
```suggestion
        propagation_direction : {'forwards', 'backwards'}
```
I am not 100% sure, but I think it is perfectly sufficient to simply write `{'forwards', 'backwards'}` without specifying the data type previously. In this context, it is also clear that this must be a string. 

At least, that is how I understood it from the [numpydoc style guide](https://numpydoc.readthedocs.io/en/latest/format.html#:~:text=When%20a%20parameter%20can%20only%20assume%20one%20of%20a%20fixed%20set%20of%20values%2C%20those%20values%20can%20be%20listed%20in%20braces%2C%20with%20the%20default%20appearing%20first)

pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +854,210 @@ def create_transmission_line(
...
+        elif not isinstance(medium_impedance, Number):
+            raise TypeError(
+                "The input medium_impedance must be a number"
+                "or a FrequencyData object.",
+            )
+        if not isinstance(propagation_direction, str):
+            raise TypeError(
+                "The input propagation_direction must be a string"
+                "with value 'forwards' or 'backwards'.",
+            )
```
h-chmeruk, line 1059:
```suggestion
```

As far as I know, there is no possibility that `propagation_direction is in ('forwards', 'backwards')` but not a string of the data type. 
Therefore, `propagation_direction in ('forwards', 'backwards') & not isinstance(propagation_direction, str)` will always be `False`. 
So I would say you can remove that check and just leave `if propagation_direction not in (“forwards”, ‘backwards’):`.

pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +854,210 @@ def create_transmission_line(
...
+            frequencies = (k_re * reference_speed_of_sound) / (2 * np.pi)
+        if isinstance(medium_impedance, FrequencyData):
+            if not np.allclose(
+                medium_impedance.frequencies,
+                frequencies,
+                atol=1e-15,
+            ):
+                raise ValueError(
+                    "The frequencies of characteristic_impedance must"
+                    "match those of k.",
```
h-chmeruk:
```suggestion
                raise ValueError(
                    "The frequencies of characteristic_impedance must "
                    "match those of k.",
```

Another missing blank between two separate words

pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +854,210 @@ def create_transmission_line(
...
+            )
+        if not isinstance(propagation_direction, str):
+            raise TypeError(
+                "The input propagation_direction must be a string"
+                "with value 'forwards' or 'backwards'.",
+            )
+        if propagation_direction not in ("forwards", "backwards"):
+            raise ValueError(
+                "The string propagation_direction must either be"
+                "'forwards' or 'backwards'.",
```
h-chmeruk:
```suggestion
                "The string propagation_direction must either be "
                "'forwards' or 'backwards'.",
```
And another missing blank between two separate words

pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +854,210 @@ def create_transmission_line(
...
+            >>> T = pf.TransmissionMatrix.create_conical_horn(
+            >>>    S0, S1, L, k, Z0, 'backwards')
+            >>> # Plot the transmission matrix
+            >>> pf.plot.freq(T.input_impedance(np.inf))
+
+        """
+        if not (isinstance(k, FrequencyData) or isinstance(k, Number)):
+            raise TypeError(
+                "The wave number k must be a float, complex,"
+                "or FrequencyData object.",
```
h-chmeruk, line 1034:
```suggestion
            raise TypeError(
                "The wave number k must be a float, complex, "
                "or FrequencyData object.",
```

Just a missing blank between ',' and a new word

tests/test_transmission_matrix.py
```diff
@@ -73,6 +81,86 @@ def test_tmatrix_from_abcd_input_types(frequencies, A_list, A_np, A_FreqDat):
...
+        ValueError,
+        match="Number of frequency values does not match the number",
+    ):
+        TransmissionMatrix.from_abcd(
+            A_list,
+            A_list,
+            A_list,
+            A_list,
+            frequencies,
+        )
```
h-chmeruk, line 168: This is another good example where you could save a few lines with  `@pytest.mark.parametrize()`

```suggestion
@pytest.mark.parametrize("frequencies", [[100, 200], [100], 100])
def test_tmatrix_from_abcd_frequencies_length_mismatch(A_list, frequencies):
    """Test from_abcd throws error if frequencies length does not match."""
    
    with pytest.raises(
        ValueError,
        match="Number of frequency values does not match the number",
    ):
        TransmissionMatrix.from_abcd(A_list, A_list, A_list, A_list, frequencies)

```

tests/test_transmission_matrix.py
```diff
@@ -73,6 +81,86 @@ def test_tmatrix_from_abcd_input_types(frequencies, A_list, A_np, A_FreqDat):
...
+    TransmissionMatrix.from_abcd(A, B, C, D, frequency)
+    TransmissionMatrix.from_abcd(A, B, C, D, [frequency])
+
+    A = 1
+    B = 2
+    C = 3
+    D = 4
+    frequency = 1000
+    TransmissionMatrix.from_abcd(A, B, C, D, frequency)
+    TransmissionMatrix.from_abcd(A, B, C, D, [frequency])
```
h-chmeruk: You could also use `@pytest.mark.parametrize()` here

tests/test_transmission_matrix.py
```diff
@@ -73,6 +81,86 @@ def test_tmatrix_from_abcd_input_types(frequencies, A_list, A_np, A_FreqDat):
...
+    TransmissionMatrix.from_abcd(A, B, C, D, frequencies)
+
+    D = 1
+    TransmissionMatrix.from_abcd(A, B, C, D, frequencies)
+
+    D = [1, 2]
+    _expect_error_abcd_broadcast(A, B, C, D, frequencies)
+
+    D = [1, 2, 3, 4]
+    _expect_error_abcd_broadcast(A, B, C, D, frequencies)
```
h-chmeruk:
```suggestion

@pytest.mark.parametrize("D", [[1], [[1]], [[1], [2]], 1])
def test_tmatrix_from_abcd_broadcasting_correct(frequencies, D):
    A, B, C = [1, 2, 3], [1], [1]
    TransmissionMatrix.from_abcd(A, B, C, D, frequencies)

@pytest.mark.parametrize("D", [ [1, 2], [1, 2, 3, 4] ])
def test_tmatrix_from_abcd_broadcasting_error(frequencies, D):
    A, B, C = [1, 2, 3], [1], [1]
    _expect_error_abcd_broadcast(A, B, C, D, frequencies)
```

The function would be shorter and easier to read if you would use `@pytest.mark.parametrize()` from pytest. You would also need to split the function into a function with correct behavior and a function with raised errors.

P.S.: Of course, you don't have to use my suggestion exactly; there may be something wrong with the proposed implementation, as I haven't tested it directly...

pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +854,210 @@ def create_transmission_line(
...
+
+        A = b / a * np.cos(k * L) - 1 / (k * a) * np.sin(k * L)
+        B = 1j * medium_impedance / (a * b * Omega) * np.sin(k * L)
+        C = (
+            1j
+            * Omega
+            / (k * k * medium_impedance)
+            * ((1 + k * k * a * b) * np.sin(k * L) - k * L * np.cos(k * L))
+        )
+        D = a / b * np.cos(k * L) + 1 / (k * b) * np.sin(k * L)
```
h-chmeruk: Maybe just add a brief comment that _this_ is the calculation of the matrix elements according to equation (5-18) from the reference

pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +854,210 @@ def create_transmission_line(
...
+            A tuple containing the area constant Omega, the distance a from the
+            narrow end to the virtual apex of the cone, and the distance b from
+            the wide end to the same virtual apex.
+        """
+        if not isinstance(S0, Number) or isinstance(S0, complex) or S0 <= 0:
+            raise ValueError("The input S0 must be a positive number.")
+        if not isinstance(S1, Number) or isinstance(S1, complex) or S1 <= 0:
+            raise ValueError("The input S1 must be a positive number.")
+        if not isinstance(L, Number) or isinstance(L, complex) or L <= 0:
+            raise ValueError("The input L must be a positive number.")
```
h-chmeruk:
```suggestion
        if not isinstance(S0, Number) or isinstance(S0, complex) or S0 <= 0:
            raise ValueError("The input S0 must be a positive real number.")
        if not isinstance(S1, Number) or isinstance(S1, complex) or S1 <= 0:
            raise ValueError("The input S1 must be a positive real number.")
        if not isinstance(L, Number) or isinstance(L, complex) or L <= 0:
            raise ValueError("The input L must be a positive real number.")
```
Maybe you could add "positive **real** number" here so that the error message clearly indicates that the error may have been caused by the complex-valued parameter.

h-chmeruk (changes requested):
Thank you very much for the implementation!

Since there is a lot of new code here, it is possible that I missed something. 

I couldn't evaluate whether the calculation part of the implementation is correct and understandable, as I am not familiar with conical horns, but I tried to look for pure logic and style issues in the code. 

Most of my comments relate to minor details; I only have one general suggestion regarding tests:

1. Some of the test functions are unnecessarily long and contain a lot of repetitive code. You could improve this with `@pytest.mark.parametrize`. 

2. In some of the test functions, you also manually calculate all the parameters for the elements of the matrix and the elements themselves. For example, in the functions `test_create_conical_horn_k_number()`, `test_create_conical_horn_imp_frequency_data()`, `test_create_conical_horn_imp_number()`, `test_create_conical_horn_broadcasting()`, and `test_create_conical_horn_default_parameters()`. 
 The calculations are quite complex, at least to me at first glance. I can't say for sure, but you could probably skip the intermediate steps of the calculations and just use the finished results if you're not using @pytest.mark.parametrize anyway. 
What I mean is, to compare correct matrix elements with the calculated matrix elements `_compare_tmat_vs_abcd(tmat_backwards, A, B, C, D)` in the function `test_create_conical_horn_k_frequency_data()`, it is sufficient to set `A, B, C, D = [correct values]` , and you don't need to calculate A, B, C, and D within the test function. 

Here's an example of what I mean:
```
def test_create_conical_horn_k_frequency_data():
    """Test create_conical_horn with impedance as FrequencyData."""
    k = FrequencyData([1j, 2, 3j], [1, 2, 3])
    Z = FrequencyData([1 + 1j, 2 + 2j, 3 + 1j], [1, 2, 3])

    a = 0.3
    b = 0.5
    Omega = 0.4

    S0 = Omega * a**2
    S1 = Omega * b**2
    L = b - a

    tmat_backwards = TransmissionMatrix.create_conical_horn(
        S0, S1, L, k, Z, "backwards",
    )
    tmat_forwards = TransmissionMatrix.create_conical_horn(
        S0, S1, L, k, Z, "forwards",
    )

    A = correct_A_for_given_parameters
    B = correct_B_for_given_parameters
    C = correct_C_for_given_parameters
    D = correct_B_for_given_parameters
    inv_prefix = 1 / (A * D - B * C)

    assert isinstance(tmat_backwards, TransmissionMatrix)
    _compare_tmat_vs_abcd(tmat_backwards, A, B, C, D)

    assert isinstance(tmat_forwards, TransmissionMatrix)
    _compare_tmat_vs_abcd(tmat_forwards, inv_prefix * D, -1 * inv_prefix * B,  -1 * inv_prefix * C, inv_prefix * A )
```

pyfar/classes/transmission_matrix.py
RoubenRehman: Thanks for highlighting that. You're right, the second check would be enough. I'd however still keep the extra check, because it yields slightly more explicit error messages. You'll get a `TypeError` on incorrect types and a `ValueError` on strings with incorrect values. Also, if you remove the `str`-check and pass e.g. a numpy array with length > 1, you'll get a value error about the uncertain truth value of such an array, which would be a bit confusing in my opinion. What do you think?

h-chmeruk: Oh, of course, I didn't think about the difference in error types...
I agree, it's better to keep both checks!

pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +854,211 @@ def create_transmission_line(
...
+        wave number :math:`k`, and medium impedance :math:`Z_0`
+        following Equation (5-18) of Reference [1]_:
+
+        .. math::
+            T = \begin{bmatrix}
+                \frac{b}{a}cos(kl)-\frac{1}{ka}sin(kl) &
+                \frac{jZ_0}{ab\Omega} \sin{kl} \\
+                \frac{j\Omega}{k^2Z_0}\left( (1 + k^2ab) \sin{kl} -
+                kl \cos{kl} \right) & \frac{a}{b}cos(kl)-\frac{1}{kb}sin(kl)
+                \end{bmatrix}
```
sikersten: This is a bit confusing, since the input parameters are not a and b, but S0 and S1. Maybe reformulate the docstring to something like?

```suggestion
        The transmission matrix is determined based on the surface
        area of the horn's narrow end, the surface area of the
        horn's wide end, the horn's length, wave number, and
        medium impedance.The surface areas are internally
        transformed into starting point, end point, and area
        constant of the horn to calculate the transmission
        matrix following Equation (5-18) of Reference [1]_.
```

pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +854,211 @@ def create_transmission_line(
...
+        ----------
+        S0 : float
+            The surface area of the horn's narrow end.
+        S1 : float
+            The surface area of the horn's wide end.
+        L : float
+            The distance between the narrow and wide end of the horn,
+            i.e. the horn's length.
+        k : FrequencyData, scalar
+            Wave number.
```
sikersten: I would suggest having more descriptive names such as `area_narrow_end`, `area_wide_end`, `length`, `wavenumber`.

pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +854,211 @@ def create_transmission_line(
...
+        medium_impedance : FrequencyData, scalar
```
sikersten, line 984: It would be nice to be consistent with [create_transmission_line method](https://pyfar--842.org.readthedocs.build/en/842/classes/pyfar.transmission_matrix.html#pyfar.classes.transmission_matrix.TransmissionMatrix.create_transmission_line), where the equivalent parameter is named `characteristic_impedance`.

pyfar/classes/transmission_matrix.py
sikersten: For the [create_transmission_line method](https://pyfar--842.org.readthedocs.build/en/842/classes/pyfar.transmission_matrix.html#pyfar.classes.transmission_matrix.TransmissionMatrix.create_transmission_line), the product kl is a parameter rather than wavenumber and length separately. This does not work here due to the calculation of a, b, and Omega, right?

pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +854,211 @@ def create_transmission_line(
...
+            The impedance of the medium filling the horn. Default is
+            ``pyfar.constants.reference_air_impedance``
+        propagation_direction : {'forwards', 'backwards'}
```
sikersten: I personally would prefer "forward" and "backward" without "s", but I think both is grammatically correct.

pyfar/classes/transmission_matrix.py
sikersten:
```suggestion
            Wavenumber.
```

pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +854,211 @@ def create_transmission_line(
...
+            >>> # Create the transmission matrix
+            >>> T = pf.TransmissionMatrix.create_conical_horn(
+            >>>    S0, S1, L, k, Z0, 'backwards')
+            >>> # Plot the transmission matrix
+            >>> pf.plot.freq(T.input_impedance(np.inf))
+
+        """
+        if not (isinstance(k, FrequencyData) or isinstance(k, Number)):
+            raise TypeError(
+                "The wave number k must be a float, complex, "
```
sikersten:
```suggestion
                "The wavenumber k must be a float, complex, "
```

pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +854,211 @@ def create_transmission_line(
...
+        Parameters
+        ----------
+        S0 : float
+            The surface area of the horn's narrow end.
+        S1 : float
+            The surface area of the horn's wide end.
+        L : float
+            The distance between the narrow and wide end of the horn,
+            i.e. the horn's length.
+        k : FrequencyData, scalar
```
sikersten: If I understand correctly, the use of the wavenumber as FrequencyData Object needs to be enforced, otherwise the frequencies for which to calculate the TransmissionMatrix not explicitely given. Am I wrong?

pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +854,211 @@ def create_transmission_line(
...
+            )
+        elif isinstance(k, FrequencyData):
+            frequencies = k.frequencies
+            k = k.freq
+        else:
+            if isinstance(k, complex):
+                k_re = k.real
+            else:
+                k_re = k
+            frequencies = (k_re * reference_speed_of_sound) / (2 * np.pi)
```
sikersten: I would refrain from this implicit declaration of evaluation frequencies and rather enforce the user to use a FrequencyData object for the wavenumber, see comment above.

pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +854,211 @@ def create_transmission_line(
...
+        S0 : float
+            Cross-sectional area at the narrow end of the horn.
+        S1 : float
+            Cross-sectional area at the wide end of the horn.
+        L : float
+            Length of the horn.
+
+        Returns
+        -------
+        tuple[float, float, float]
```
sikersten:
```suggestion
        Omega, a, b: tuple[float, float, float]
```

pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +854,211 @@ def create_transmission_line(
...
+            )
+
+        r0 = np.sqrt(S0 / np.pi)
+        r1 = np.sqrt(S1 / np.pi)
+
+        d0 = 2 * r0
+        d1 = 2 * r1
+
+        a = r0 * (2 * L) / (d1 - d0)
+        b = a + L
```
sikersten: could be shortened to

```suggestion
        r0 = np.sqrt(S0 / np.pi)
        r1 = np.sqrt(S1 / np.pi)

        a = r0 * L / (r1 - r0)
        b = a + L
```

pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +854,211 @@ def create_transmission_line(
...
+        Omega0 = S0 / a**2
+        Omega1 = S1 / b**2
+
+        if not np.isclose(Omega0, Omega1, atol=1e-15):
+            raise ValueError(
+                f"S0/a² is not equal to S1/b²: {Omega0} != {Omega1}."
+                "This should never happen.",
+            )
+
+        Omega = Omega0  # or Omega1, they are equal
```
sikersten: If there can't be any case where the error happends, why not

```suggestion
        Omega = S0 / a**2  # equivalent to S1 / b**2
```

pyfar/classes/transmission_matrix.py
```diff
@@ -825,6 +854,211 @@ def create_transmission_line(
...
+            )  # Error handling inside the helper function
+            L = -1 * L
+
+        # Calculate T-matrix entries according to Equation (5-18) of [1]
+        A = b / a * np.cos(k * L) - 1 / (k * a) * np.sin(k * L)
+        B = 1j * medium_impedance / (a * b * Omega) * np.sin(k * L)
+        C = (
+            1j
+            * Omega
+            / (k * k * medium_impedance)
```
sikersten: Why not use `k**2` here?

sikersten (changes requested):
Thanks Rouben! I reviewed the code related to the new feature. Looks all in all fine to me, mostly minor comments.

pyfar/classes/transmission_matrix.py
RoubenRehman: I don't think those two are equivalent. The `create_transmission_line` method can be used with specific and acoustic impedances. If a relation between two volume velocities is desired, you pass the acoustic impedance (p/q), if you want to relate two particle velocities, you pass the specific/characteristic impedance (p/v).

The `create_conical_horn` method in its current implementation only works with volume velocities. However, because there is no constant cross-sectional area in a horn, only the medium's specific impedance is passed and the translation into an acoustic impedance happens internally. 

Both implementations (for volume velocities or particle velocities) are of course possible, but the issue is that due to the changing cross-section, there is in my understanding no general implementation where this is decided based on what impedance the user passes. At the very least, a general implementation would require a fixed reference area (S0 or S1) for the passed impedance as a convention. But for such an implementation I think it'd be much more explicit and clear to just pass a specific medium impedance and decide between volume or particle velocity via an additional binary flag.

pyfar/classes/transmission_matrix.py
RoubenRehman: Agreed, I will do it like this

pyfar/classes/transmission_matrix.py
RoubenRehman: Thank you for the input Simon, I agree! I'd prefer, however, to keep the matrix in the docstring. That way one can see at a glance, whether this function yields the correct transmission matrix for the application. What do you think?

pyfar/classes/transmission_matrix.py
RoubenRehman: yes, exactly

pyfar/classes/transmission_matrix.py
RoubenRehman: In the current implementation yes, because `from_abcd` only works for frequency-dependent inputs. With one exception, the other `create_` functions, however, also work independently of frequency/for a singular frequency. In that case, they return a `np.ndarray`. I think this should be consistent for all `create_` functions. Currently, `create_transmission_line` is the only one of those, that doesn't support singular frequencies - maybe that could be adjusted there too?

pyfar/classes/transmission_matrix.py
RoubenRehman: As noted above, I think all `create_` functions should support singular frequencies returning a `np.ndarray` in that case. Currently (on develop), `create_transmission_line` is the only one not doing that. 

But this implicit calculation of `frequencies` can be removed nontheless, because the new `create_frequency_independent_abcd` doesn't need it.

pyfar/classes/transmission_matrix.py
RoubenRehman: That was meant as a safety net and to make the code more explicit, but the comment in your suggested line does this too, thanks!

pyfar/classes/transmission_matrix.py
```diff
@@ -973,7 +973,7 @@ def create_conical_horn(
         medium_impedance : FrequencyData, scalar
             The impedance of the medium filling the horn. Default is
             ``pyfar.constants.reference_air_impedance``
-        propagation_direction : {'forward', 'backwards'}, optional
+        propagation_direction : {'forward', 'backward'}, optional
```
mberz: I would suggest to use more explicit terminology here like "tapering" and "expanding" to describe the horn geometry. Forward and backward are not really absolute terms and depend on how one looks at the geometry.

RoubenRehman:
Thanks @sikersten for your comments! I've implemented most of them now:
- Dependence on #823 is solved by rebasing on develop to get `create_frequency_independent_abcd`
- Docstring of the function changed to your suggestion but I've kept the matrix as stated above. Happy to hear thoughts on this
- Parameter names are now more descriptive
- Its now "forward" and "backward" without an "s"
- Calculations are simplified

I've refrained from renaming `medium_impedance` to `characteristic_impedance` due to the isue with the changin cross-section. I think that was the consensus on this when discussed in a Friday Meeting, too.

I've also kept the functionality of passing scalars for `k` to represent frequency independent / singular frequency tmats. This was reworked to use the new `create_frequency_independent_abcd` now. As stated above, I think all `create_` functions should support this, including `create_transmission_line` which currently doesn't.

pyfar/classes/transmission_matrix.py
mberz: To me the difference between acoustic and characteristic impedance is a bit confusing. 
Specifically, because the term characteristic impedance is used in different ways in different literature.
I think the general use of `medium_impedance` makes a lot of sense here in general (also in the function @sikersten mentioned).
To avoid further confusion I'd suggest to add the quantities contained in the lhs and rhs vectors (maybe even including their units) to the docstring.

mberz (changes requested):

pyfar/classes/transmission_matrix.py
```diff
@@ -873,6 +874,222 @@ def create_transmission_line(
...
+            Cross-sectional area at the wide end of the horn.
+        horn_length : Number
+            Length of the horn.
+
+        Returns
+        -------
+        Omega, a, b: tuple[Number, Number, Number]
+            A tuple containing the area constant Omega, the distance a from the
+            narrow end to the virtual apex of the cone, and the distance b from
+            the wide end to the same virtual apex.
```
h-chmeruk:
```suggestion
        Omega : Number
            The area constant.
        a : Number
            The distance from the narrow end to the virtual apex of the cone.
        b : Number
            The distance from the wide end to the same virtual apex.
```
I'm not insisting on it, but I would say that the proposed code is more appropriate because it follows the [numpy style guide](https://numpydoc.readthedocs.io/en/stable/format.html).

pyfar/classes/transmission_matrix.py
```diff
@@ -873,6 +874,222 @@ def create_transmission_line(
...
+        if area_narrow_end > area_wide_end:
+            raise ValueError("area_narrow_end must be "
+                             "strictly smaller than area_wide_end.")
+        if area_narrow_end == area_wide_end:
+            raise ValueError(
+                "For a conical horn area_narrow_end "
+                "must be strictly smaller than area_wide_end."
+                "If area_narrow_end == area_wide_end, "
+                "use :fun:`create_transmission_line` instead.",
+            )
```
h-chmeruk:
```suggestion
        if area_narrow_end >= area_wide_end:
            raise ValueError(
                "For a conical horn area_narrow_end "
                "must be strictly smaller than area_wide_end."
                "If area_narrow_end == area_wide_end, "
                "use :fun:`create_transmission_line` instead.",
            )
```
I would say that one if statement is sufficient, as they raise the same ValueError and can be combined into one if statement.

pyfar/classes/transmission_matrix.py
```diff
@@ -873,6 +874,222 @@ def create_transmission_line(
...
+        wave_number: Number | FrequencyData,
+        medium_impedance: Number | FrequencyData = reference_air_impedance,
+        propagation_direction: str = "forward",
+    ) -> np.ndarray | TransmissionMatrix:
+        r"""Create a transmission matrix representing a conical horn.
+
+        The transmission matrix is determined based on the surface
+        area of the horn's narrow end, the surface area of the
+        horn's wide end, the horn's length, wavenumber, and
+        medium impedance.The surface areas are internally
```
h-chmeruk:
```suggestion
        medium impedance. The surface areas are internally
```

tests/test_transmission_matrix.py
```diff
@@ -467,6 +467,501 @@ def test_create_transmission_line_frequency_matching():
...
+    [{"a": 1, "b": 2}, np.array([1, 2, 3]), "term", 0, -5, 2j],
+)
+def test_calculate_horn_geometry_parameters_L_value_error(horn_length):
+    area_narrow_end = 0.2
+    area_wide_end = 0.3
+
+    with pytest.raises(ValueError, match="The input horn_length"):
+        TransmissionMatrix.\
+            _calculate_horn_geometry_parameters\
+            (area_narrow_end, area_wide_end, horn_length)
```
h-chmeruk:
```suggestion
@pytest.mark.parametrize("param", [
    "area_narrow_end",
    "area_wide_end",
    "horn_length",
])
@pytest.mark.parametrize(
    "invalid_value",
    [{"a": 1, "b": 2}, np.array([1, 2, 3]), "term", 0, -5, 2j],
)
def test_calculate_horn_geometry_parameters_value_error(param, invalid_value):
    parameters = {
        "area_narrow_end": 0.2,
        "area_wide_end": 0.3,
        "horn_length": 0.35,
    }
    parameters[param] = invalid_value

    with pytest.raises(ValueError, match=f"The input {param} must be"
                                            " a positive real number."):
        TransmissionMatrix._calculate_horn_geometry_parameters(
            parameters["area_narrow_end"],
            parameters["area_wide_end"],
            parameters["horn_length"])

```
You could also combine these tests.

tests/test_transmission_matrix.py
```diff
@@ -467,6 +467,501 @@ def test_create_transmission_line_frequency_matching():
...
+def test_calculate_horn_geometry_parameters_S0_equal_S1_value_error():
+    area_narrow_end = 0.35
+    area_wide_end = 0.35
+    horn_length = 0.2
+
+    with pytest.raises(ValueError, match=\
+        "For a conical horn area_narrow_end must be"):
+        TransmissionMatrix.\
+            _calculate_horn_geometry_parameters\
+            (area_narrow_end, area_wide_end, horn_length)
```
h-chmeruk: If you are going to remove one of the if statements in lines 913-922 in the classes/transmission_matrix.py file, you should also modify one of these tests and delete the other. For example:

```suggestion
@pytest.mark.parametrize("value", [0.3, 0.35])
def test_calculate_horn_geometry_parameters_S0_larger_S1_value_error(value):
    area_narrow_end = 0.35
    area_wide_end = value
    horn_length = 0.35

    with pytest.raises(
        ValueError,
        match="area_narrow_end must be strictly smaller than area_wide_end.",
    ):
        TransmissionMatrix._calculate_horn_geometry_parameters(
            area_narrow_end, area_wide_end, horn_length)
```

h-chmeruk (changes requested):
Thank you very much for implementing the previous changes!
 I have noticed a few more potential improvements, but these are just my personal comments, nothing you necessarily need to change.

RoubenRehman:
Thanks @h-chmeruk and @mberz for your comments! I've implemented all of them now.

h-chmeruk (approved):
Thank you for implementing the changes!
As I said earlier, I haven’t checked the calculations in the new function or the tests, but the logic, syntax and documentation all look fine to me

PR #843, opened 2025-09-23
Add `StateSpaceModel` class

### LINKED ISSUE #780: Feature request: State-space Systems

artpelling:
## Overview
This pull request introduces a new class for LTI systems (discrete-time only) using the state-space formulation. This allows efficient and numerically stable modelling of multi-input, multi-output (MIMO) systems.

- Added the `StateSpaceModel` class to `pyfar.classes.filter`, with support for MIMO, custom initial states, and impulse response computation.
- The `_process` method is a bit cryptic but removes quite some overhead by calling BLAS routines directly. I also benchmarked memory layouts and I am using Fortran style here. The resulting performance now is pretty good for small to medium-large models and not too far from a compiled solver :racing_car: 

Closes #780 

- adds new class with rudimentary state-space system support.

~~Waiting for #798 to be merged first.~~

### TODO
- [x] Add tests
- [ ] Add method for estimation of IR length (need to think about theory first)
- [x] Decide whether to add a tutorial
- [x] Discuss handling of multichannel inputs and outputs


artpelling:
The current implementation works as follows. The state-space system has a fixed number of in- and outputs and accepts input signals with conforming number of channels, i.e.

```python
#!/usr/bin/env python3

import numpy as np
from pyfar.classes.filter import StateSpaceModel
from pyfar import Signal

A = np.random.rand(3, 3)
B = np.random.rand(3, 2)
C = np.random.rand(4, 3)
D = np.random.rand(4, 2)

sys = StateSpaceModel(A, B, C, D, sampling_rate=44100)

print(sys)

input_signal = Signal(np.random.randn(2, 100), sampling_rate=44100)

print(input_signal)

output_signal = sys.process(input_signal)

print(output_signal)

```
Outputs:
```
Order 3 state-space model with 2 inputs and 4 outputs @ 44100 Hz sampling rate (float64).
time domain energy Signal:
(2,) channels with 100 samples @ 44100 Hz sampling rate and none FFT normalization

time domain energy Signal:
(4,) channels with 100 samples @ 44100 Hz sampling rate and none FFT normalization

```

BTW: I noticed that the `__repr__` are inconsistent, sometimes having trailing `\n` and sometimes not...

mberz:
> BTW: I noticed that the `__repr__` are inconsistent, sometimes having trailing `\n` and sometimes not...

Please feel free to contribute to the `__repr__` discussion or report occurrences here:
https://github.com/pyfar/pyfar/issues/903

pyfar/classes/filter.py
```diff
@@ -1140,3 +1141,241 @@ def __repr__(self):
...
+    sampling_rate : number, optional
+        The sampling rate of the system in Hz. The default is ``None``.
+    state : array, double, optional
+        The initial state of the system.
+    dtype : np.dtype, optional
+        The data type of the system matrices. Can be used to set the precision
+        of the response calculation. If ``None``, ``np.promote_types`` is used.
+    comment : str, optional
+        A comment. The default is ``''``.
+    """
```
ahms5, line 1228: how about adding a small example?

pyfar/classes/filter.py
```diff
@@ -1140,3 +1141,241 @@ def __repr__(self):
...
+                y=gemv(1., self._A, self._state, beta=0, y=self._state),
+            )
+        return y
+
+    def impulse_response(self, n_samples):
+        """Compute the impulse response of the system.
+
+        Parameters
+        ----------
+        n_samples : int, None
```
ahms5:
```suggestion
        n_samples : int
```

looks like None is not allowed

HISTORY.rst
```diff
@@ -2,6 +2,12 @@
 History
 =======
 
+0.8.1 (2026-XX-XX)
```
ahms5:
```suggestion
0.9.0 (2026-XX-XX)
```

I think this would go into the next minor release.

ahms5 (commented):
nice, just minor comments

artpelling:
I've added small example that renders like this:

image

artpelling:
@mberz I've updated the implementation to ensure correct memory layout of the output signals. I managed to keep the speed by using Fortran layout for the state-space matrices. Blue=before, orange=`56687ae`. If that is unwanted, too, I can update but this will lead to performance losses, see green.

state_space_model_layout_benchmark

pyfar/classes/filter.py
```diff
@@ -1140,3 +1141,266 @@ def __repr__(self):
...
+
+    A state-space model is defined by the matrices :math:`A`, :math:`B`,
+    :math:`C`, and :math:`D`. Contrary to an impulse response or filter model,
+    a state-space model can represent multi-input multi-output (MIMO) linear
+    time-invariant (LTI) systems in a numerically stable and efficient way. The
+    system is described by the following equations (discrete-time):
+
+    .. math:
+        x_{t+1} = A x_t + B u_t
+        y_t   = C x_t + D u_t,
```
f-brinkmann: These do not show up in the docs. Probably formatting (maybe a blank line required after `.. math:`).

pyfar/classes/filter.py
```diff
@@ -1140,3 +1141,266 @@ def __repr__(self):
...
+
+    where :math:`x_t` is the state, :math:`u` the input, and :math:`y` the
+    output at time step :math:`t`.
+
+    The matrix :math:`A` is the state matrix, defining the internal dynamics of
+    the system, i.e. how the internal state evolves on its own; :math:`B` is
+    the input matrix, defining the action of the input :math: `u` on the state;
```
f-brinkmann: Does not render correctly, possibly due to the additional space.
```suggestion
    the input matrix, defining the action of the input :math:`u` on the state;
```

pyfar/classes/filter.py
```diff
@@ -1140,3 +1141,266 @@ def __repr__(self):
...
+    C : array
+        The output matrix :math:`C` with dimensions ``(n_outputs, order)``.
+    D : array, optional
+        The feedthrough matrix :math:`D` with dimensions ``(n_outputs,
+        n_inputs)``. The default is ``None``, which initializes an all-zero
+        matrix.
+    sampling_rate : number, optional
+        The sampling rate of the system in Hz. The default is ``None``.
+    state : array, double, optional
+        The initial state of the system.
```
f-brinkmann: Can you please mention the shape of state ``(order,)``?

pyfar/classes/filter.py
```diff
@@ -1140,3 +1141,266 @@ def __repr__(self):
...
+        ...     label='Butter FilterSOS', marker='x')
+        >>> ax = pf.plot.time(ssm.impulse_response(100),
+        ...     label='Butter StateSpaceModel', ax=ax)
+        >>> ax.legend(loc='right')
+
+    """
+
+    def __init__(self, A, B, C, D=None, sampling_rate=None, state=None,
+            dtype=None, comment=""):
+        D = np.zeros((C.shape[0], B.shape[1])) if D is None else D
```
f-brinkmann: The below would fail if an array like is passed. Although not very likely, may add
`A = np.asarray(A)`, ... and so on?

pyfar/classes/filter.py
```diff
@@ -1140,3 +1141,266 @@ def __repr__(self):
...
+        ----------
+        state : array, self.dtype
+            The internal state of the system with dimensions ``(order,)``.
+        """
+        return self._state
+
+    @state.setter
+    def state(self, state):
+        assert state.shape == (self.order,)
+        assert state.dtype == self.dtype
```
f-brinkmann: The docs state that state can be a double as well. To make this work, probably a `state = np.atleast_1d(state)` is required. Can you also add a test for this please?

pyfar/classes/filter.py
```diff
@@ -1140,3 +1141,266 @@ def __repr__(self):
...
+        Returns
+        -------
+        out : Signal
+            The processed output signal.
+        """
+        assert signal.sampling_rate == self.sampling_rate, (
+            "The sampling rates of the signal and the state-space model"
+            "need to match."
+        )
+        u = np.ascontiguousarray(
```
f-brinkmann, line 1350: This is quite implicit. If a one dimensional signal is required (and also returned) and the number of channels must match I would recomend to document and enforce this with

`assert signal.cshape == (self.n_inputs, )`

to make the behaviour explicit.

pyfar/classes/filter.py
```diff
@@ -1140,3 +1141,266 @@ def __repr__(self):
...
+        obj.__dict__.update(obj_dict)
+        return obj
+
+    def __repr__(self):
+        """Representation of the state-space model."""
+        return (
+            f"Order {self.order} state-space model with {self.n_inputs} inputs"
+            f" and {self.n_outputs} outputs"
+            f" @ {self.sampling_rate} Hz sampling rate ({self.dtype})."
+        )
```
f-brinkmann: We started to make representations as short as possible. This will help to see the most important information, e.g., in a variable viewer. Would this be most important for handling it or would you rather include more information?

f'StateSpaceModel ({self.n_inputs} inputs, {self.n_outputs} outputs)'

f-brinkmann (changes requested):
Thanks for the effort and sorry for the wait. Looks quite good to me already. Had only minor comments and would ask you to add tests for all getters and setters.

pyfar/classes/filter.py
artpelling: missing colon :)

pyfar/classes/filter.py
artpelling: I am not sure what you mean. I have a check for compatibility in the following lines, after casting `signal` to `u`. This makes the check a bit easier, because the reshaping has taken place.

f-brinkmann: Sorry, maybe not clear enough. I would add to the docstring that the cshape of the input signal must be broadcastable to `(self.n_inputs, )` and that the cshape of the output signal is `(self.n_outputs, )`. For the input signal we could also force `signal.cshape = (self.n_inputs, )` to make it more explicit.

f-brinkmann (commented):
Clarified my previous comment and wondered if you saw my comment on adding tests for the getters?

artpelling:
> Clarified my previous comment and wondered if you saw my comment on adding tests for the getters?

Thanks! I did not and also cannot find it? Do you mean the getters of `A,B,C,D`?

f-brinkmann (changes requested):
> Thanks! I did not and also cannot find it? Do you mean the getters of `A,B,C,D`?

@artpelling yes, we usually test all getters (and setters). Although it's a little trivial and annyoing here I would suggest to do so. Sorry, mentioned that in the general review comment, where it most easy to miss it 🙈. You can for example use `pytest --cov=. --cov-report=html` to check your test coverage.

Other than that, things look good to me.

artpelling:
@f-brinkmann I've added tests as requested although all of these setters and getters were reported as covered before.

f-brinkmann (approved):
Thx for the effort. Good to go from my side!

artpelling:
@mberz I think this is ready now? :)

PR #845, opened 2025-09-30
First draft of the _PyfarArithmetics class

h-chmeruk:
As proposed in the [PR #838](https://github.com/pyfar/pyfar/pull/838), here is a draft implementation of the _PyfarArithmetics class. This abstract class should define the basic arithmetic operations.

### DIFF
--- a/pyfar/classes/_PyfarArithmetics.py
+++ b/pyfar/classes/_PyfarArithmetics.py
@@ -1,0 +1,77 @@
+"""
+The following documents the abstract class for pyfar classes,
+which defines arithmetic operations.
+"""
+
+from abc import ABC, abstractmethod
+
+class _PyfarArithmetics(ABC):
+    """
+    Internal abstract base class for pyfar classes.
+
+    This class defines the basic arithmetic operations (addition, subtraction,
+    multiplication, division and exponentiation) and their right-hand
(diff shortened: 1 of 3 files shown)


pyfar/classes/_PyfarArithmetics.py
```diff
@@ -0,0 +1,114 @@
...
+            raise TypeError("Incompatible object for additive operations")
+
+        if hasattr(other, "_data"):
+            data = other._data
+        else:
+            data = other
+
+        if self._data.shape != data.shape:
+            raise ValueError("Incompatible shapes for additive operations")
+        return data
```
h-chmeruk, line 35: So I'm not sure if this is the right approach, but it was just an idea that came to me while implementing the arithmetic operations. It's clear that most functions for arithmetic operations require some kind of input validation and then data extraction, so I added this function called `_check_data_add_sub`, which handles this for all additive operations. However, I can imagine that I might have overlooked something and not considered all possible cases. So please don’t hesitate to give me feedback on whether this is a good idea or if I’ve made a mistake in my reasoning. 

P.S.: The function's name could also be `_check_other_additive`, and there could be another function, `_check_other_multiplicative`, for multiplication and division.

PR #849, opened 2025-10-08
Feature: time-slicing function for pyfar.Signal objects

### LINKED ISSUE #751: Feature request: Time slicing and/or length modification for Signal objects
With the current implementation, there is no convenient way to change the length of a Signal or extract a part of it as a new Signal. As a regular pyfar user, I often encounter situations where this would be useful, however.
Use-cases:
- Splitting a signal into blocks of same size
- Bringing several different Signals of different lengths to the same length to add them
- Plotting parts of a signal without having to change the plot to use samples as units and then setting plt.xlim

Right now, the typical workflow to tackle these tasks is to get the raw time ndarray, modify the ndarray, then construct a new Signal. Here is an example to get a longer signal to the length of a shorter one:
```py
longer_signal = pf.signals.noise(300)
shorter_signal = pf.signals.noise(100)

trimmed_data = longer_signal.time[:, 0:shorter_signal.n_samples]
trimmed_signal = pf.Signal(trimmed_data, longer_signal.sampling_rate)
```

The idea here is to have a function - let's call it `extract` that does this in a shorter and more legible way, like:
`trimmed_signal = pf.dsp.extract(longer_signal, shorter_signal.n_samples)`

Such a function could also include functionality for slicing, or zero-padding with a target length rather than a target padding width. A very simple implementation could look like this:

```py
def extract(signal: pf.Signal, len: int, start=0, pad_zeros: str|None="end") -> pf.Signal:
    end = start + len
    time_data = signal[:, start:end]
    out = pf.Signal(time_data, signal.sampling_rate)
    
    # if pad_zeros is None, don't pad. Otherwise forward it as mode to pad zeros
    if out.n_samples < len and pad_zeros:
        out = pf.dsp.pad_zeros(out, len - out.n_samples, pad_zeros)
    return out
    
# usage example
trimmed = extract(longer_signal, shorter_signal.n_samples)
padded = extract(shorter_signal, longer_signal.n_samples)
middle_block = extract(shorter_signal, 20, 40)
```

Further considerations:
- a fitting name for the function (perhaps time_slice() would be more fitting?)
- should it be an instance method instead? (`sig.extract(len)` vs `pf.dsp.extract(sig, len)`)
- should we instead/additionally add a proper slice() method for pf.Signal (which would allow the [:, start:end] syntax directly. However, I find this confusing because it's not obvious whether we are slicing channels or samples, since the normal iterator method and the [] operator yield channels. Also, slice() uses start and end, whereas I believe for audio it's usually more useful to specify start and *length* instead)

h-chmeruk:

Closes #751

This is my implementation of the pyfar.dsp.time_crop function, which can be used to crop a signal in the time domain. I have also written some test functions, most of which are similar to the existing test functions for pyfar.dsp.time_window.

### DIFF
--- a/pyfar/dsp/dsp.py
+++ b/pyfar/dsp/dsp.py
@@ -566,9 +566,12 @@ def time_window(signal, interval, window='hann', shape='symmetric',
 def time_crop(signal, interval: Union[list[float], tuple[float, float],
                                        np.ndarray],
               unit: Literal["samples", "s"]='samples'):
-    """Crop a :py:class:`Signal <pyfar.Signal>` or a
+    r"""Crop a :py:class:`Signal <pyfar.Signal>` or a
     :py:class:`TimeData <pyfar.TimeData>` object in time.
-    Returns the signal within the specified `interval`.
+
+    Returns the signal :math:`x(t)` defined for all :math:`t` within the
+    interval :math:`interval[0] \le t \le interval[1]`, where :math:`t`
+    can be time or samples.
 
     Parameters
     ----------
@@ -591,6 +594,14 @@ def time_crop(signal, interval: Union[list[float], tuple[float, float],
     cropped_signal : pyfar.Signal, pyfar.TimeData
         Cropped :py:class:`Signal <pyfar.Signal>` or
         :py:class:`TimeData <pyfar.TimeData>` object.
+
+    Notes
+    -----
+    For :py:class:`Signal <pyfar.Signal>`, the starting point of
+    the cropped signal is always set to zero, while for
+    :py:class:`TimeData <pyfar.TimeData>`, the starting point is set to the
+    point where cropping begins in the original signal.
+
 
     Examples
     --------


pyfar/dsp/dsp.py
```diff
@@ -560,6 +560,64 @@ def time_window(signal, interval, window='hann', shape='symmetric',
     else:
         return signal_win
 
+def time_crop(signal, interval, unit='samples'):
+    """This function can be used to crop a signal object.
+
+    Parameters
+    ----------
+    signal : pyfar.Signal
```
ahms5: How about also allowing TimeData?

pyfar/dsp/dsp.py
```diff
@@ -560,6 +560,64 @@ def time_window(signal, interval, window='hann', shape='symmetric',
     else:
         return signal_win
 
+def time_crop(signal, interval, unit='samples'):
+    """This function can be used to crop a signal object.
```
ahms5:
```suggestion
    """Crop a Signal or TimeData in time.
```

pyfar/dsp/dsp.py
```diff
@@ -560,6 +560,64 @@ def time_window(signal, interval, window='hann', shape='symmetric',
...
+    Parameters
+    ----------
+    signal : pyfar.Signal
+        Signal object to be cropped.
+
+    interval : array_like
+        Must contain exactly two entries, these specify the beginning and
+        the end of the section to be cropped. The unit of `interval`
+        is specified by the parameter `unit`.
+        See below for more details.
```
ahms5:
```suggestion
```
might be redundant.

pyfar/dsp/dsp.py
```diff
@@ -560,6 +560,64 @@ def time_window(signal, interval, window='hann', shape='symmetric',
...
+        Time values are rounded to the nearest sample. The default is
+        ``'samples'``.
+
+    Returns
+    -------
+    cropped_signal : pyfar.Signal
+        Cropped signal object.
+
+    """
+    #Check the input
```
ahms5:
```suggestion
    # Check the input
```

pyfar/dsp/dsp.py
```diff
@@ -560,6 +560,64 @@ def time_window(signal, interval, window='hann', shape='symmetric',
...
+    # Convert to samples
+    interval = np.array(interval)
+    if unit == 's':
+        interval = np.round(interval*signal.sampling_rate).astype(int)
+    elif unit == 'samples':
+        interval = interval.astype(int)
+    else:
+        raise ValueError(f"unit is {unit} but has to be 'samples' or 's'.")
+
+    #Check the start and the end points
```
ahms5:
```suggestion
    # Check the start and the end points
```

tests/test_dsp.py
```diff
@@ -682,6 +682,50 @@ def test_time_window_return_window_error():
     with pytest.raises(TypeError, match="boolean"):
         dsp.time_window(sig, interval=(4, 8), return_window='a')
 
+def test_time_crop_signal_types():
+    """Raise TypeError if `signal` is not a pyfar.Signal instance."""
+    signal_1 = pyfar.Signal(np.ones(10), 2)
```
ahms5:
```suggestion
def test_time_crop_signal_types(sine_stub):
    """Raise TypeError if `signal` is not a pyfar.Signal instance."""
```

you can also use a fixture to get any example Signal. see [conftest.py](https://github.com/pyfar/pyfar/blob/develop/tests/conftest.py)

tests/test_dsp.py
```diff
@@ -682,6 +682,50 @@ def test_time_window_return_window_error():
...
+    dsp.time_crop(signal_1, interval=(1, 2))
+    signal_2 = "signal"
+    with pytest.raises(TypeError, match='has to be of type'):
+        dsp.time_crop(signal_2, interval=(1, 2))
+
+def test_time_crop_interval_types():
+    """Accept list or tuple for `interval`, raise TypeError otherwise."""
+    signal = pyfar.Signal(np.ones(10), 2)
+    dsp.time_crop(signal, interval=(1, 2))
+    dsp.time_crop(signal, interval=[1, 2])
```
ahms5: here you could use parameters as input for intervals of differnt type. [Here ](https://github.com/pyfar/pyfar/blob/cfafcf9fc4f52408dd9934b3f38afc5dfb90de6a/tests/test_io.py#L21)would be an example.

tests/test_dsp.py
```diff
@@ -682,6 +682,50 @@ def test_time_window_return_window_error():
...
+    with pytest.raises(TypeError, match='has to be of type'):
+        dsp.time_crop(signal, interval=5)
```
ahms5, line 709: this could be also shifted into a seperate test

pyfar/dsp/dsp.py
```diff
@@ -560,6 +560,64 @@ def time_window(signal, interval, window='hann', shape='symmetric',
...
+    unit : string, optional
+        Unit of `interval`. Can be set to ``'samples'`` or ``'s'`` (seconds).
+        Time values are rounded to the nearest sample. The default is
+        ``'samples'``.
+
+    Returns
+    -------
+    cropped_signal : pyfar.Signal
+        Cropped signal object.
+
```
ahms5, line 597: maybe we could add a small example here.

pyfar/dsp/dsp.py
```diff
@@ -560,6 +560,64 @@ def time_window(signal, interval, window='hann', shape='symmetric',
     else:
         return signal_win
 
+def time_crop(signal, interval, unit='samples'):
```
jf-som: I suggest adding types so people using VSCode etc can get helpful hints and autocompletion
```suggestion
def time_crop(signal, interval: tuple[float, float], unit: Literal["samples", "s"]='samples'):
```

pyfar/dsp/dsp.py
```diff
@@ -560,6 +560,64 @@ def time_window(signal, interval, window='hann', shape='symmetric',
...
+        raise TypeError(
+            "The parameter interval has to be of type list or tuple.")
+    if len(interval) != 2:
+        raise ValueError(
+            "The parameter interval must consist of 2 entries.")
+
+    # Convert to samples
+    interval = np.array(interval)
+    if unit == 's':
+        interval = np.round(interval*signal.sampling_rate).astype(int)
```
jf-som:
```suggestion
        interval = np.round(interval * signal.sampling_rate).astype(int)
```

pyfar/dsp/dsp.py
```diff
@@ -560,6 +560,64 @@ def time_window(signal, interval, window='hann', shape='symmetric',
     else:
         return signal_win
```
jf-som, line 564:
```suggestion


```
In python styling, we usually 2 lines of whitespace are recommended between functions

pyfar/dsp/dsp.py
```diff
@@ -560,6 +560,64 @@ def time_window(signal, interval, window='hann', shape='symmetric',
...
+    if start >= end:
+        raise ValueError("The start point in the parameter interval must" \
+        "be smaller than the end point.")
+
+    start_index = max(start, 0)
+    end_index = min(end, signal.n_samples)
+    time_data = signal.time[:, start_index:end_index]
+    cropped_signal = pyfar.Signal(time_data, signal.sampling_rate)
+
+    return cropped_signal
```
jf-som, line 703:
```suggestion


```

pyfar/dsp/dsp.py
```diff
@@ -560,6 +560,64 @@ def time_window(signal, interval, window='hann', shape='symmetric',
...
+
+    """
+    #Check the input
+    if not isinstance(signal, pyfar.Signal):
+        raise TypeError("The parameter signal has to be of type " \
+        "pyfar.Signal.")
+
+    if not isinstance(interval, (list, tuple)):
+        raise TypeError(
+            "The parameter interval has to be of type list or tuple.")
```
jf-som: In the comment you say the parameter is `array-like`, which usually means anything that can be used like an array or converted to one. This includes `np.ndarray` as well as many other data structures. 
If you want to explicitly check the type like you do now (which I think is a good thing!), I would mention the allowed types explicitly in the comment (like `tuple | list` and maybe add np.ndarray to the mix).

pyfar/dsp/dsp.py
h-chmeruk: Thank you for your review!
I'm not sure what kind of example I should implement. Maybe you could link to some references or briefly describe what the example should be about?

pyfar/dsp/dsp.py
```diff
@@ -561,6 +563,80 @@ def time_window(signal, interval, window='hann', shape='symmetric',
         return signal_win
 
 
+def time_crop(signal, interval: Union[list[float], tuple[float, float],
+                                       np.ndarray],
+              unit: Literal["samples", "s"]='samples'):
+    """This function can be used to crop a pf.Signal or pf.TimeData object.
```
ahms5:
```suggestion
    """Crop a pf.Signal or pf.TimeData object in time.
```

pyfar/dsp/dsp.py
ahms5: see e.g. [here](https://pyfar.readthedocs.io/en/stable/modules/pyfar.dsp.html#pyfar.dsp.time_window).
a simple example cout just be a sine, which is cropped, then both signals are plot.

tests/test_dsp.py
```diff
@@ -682,6 +682,83 @@ def test_time_window_return_window_error():
...
 
+def test_time_crop_signal_types():
+    """
+    Raise TypeError if `signal` is not a pyfar.Signal or pyfar.TimeData
+    instance.
+    """
+    signal_1 = pyfar.Signal(np.ones(10), 2)
+    dsp.time_crop(signal_1, interval = (1, 2))
+    signal_2 = pf.TimeData([1, 2], [1, 2])
+    dsp.time_crop(signal_2, interval = (1, 2))
```
ahms5:
```suggestion
```
is the above comment is the purpose of the test, these lines seems to be unnessesary.

tests/test_dsp.py
```diff
@@ -682,6 +682,83 @@ def test_time_window_return_window_error():
...
+
+@pytest.mark.parametrize("signal", [pyfar.Signal(np.ones(10), 2),
+                                     pf.TimeData([1, 2], [1, 2])])
+@pytest.mark.parametrize("interval", [(1, 2), [1, 2], np.array([1, 2])])
+
+def test_time_crop_interval_types(signal, interval):
+    """Accept list or tuple for `interval`, raise TypeError otherwise."""
+    dsp.time_crop(signal, interval)
+    dsp.time_crop(signal, interval)
+    dsp.time_crop(signal, interval)
```
ahms5:
```suggestion
```
isnt one call sufficiant?

tests/test_dsp.py
```diff
@@ -682,6 +682,83 @@ def test_time_window_return_window_error():
...
+    with pytest.raises(ValueError, match = '2 entries'):
+        dsp.time_crop(signal, interval=[1, 2, 3])
+
+@pytest.mark.parametrize("signal", [pyfar.Signal(np.ones(10), 2),
+                                     pf.TimeData([1, 2], [1, 2])])
+def test_time_crop_interval_size(signal):
+    """Allow `interval` boundaries outside signal range."""
+    cropped_signal = dsp.time_crop(signal, interval=[0, 40])
+    cropped_signal = dsp.time_crop(signal, interval=[-1, 2])
+    # If both boundaries are exceeded, the input signal is returned unchanged.
```
ahms5: I think we should return an error in this case. see [Design Considerations](https://pyfar-gallery.readthedocs.io/en/latest/contribute/packages/style_design.html#design-considerations). I would say these are invalid parameter combinations.
[time_window](https://pyfar.readthedocs.io/en/stable/_modules/pyfar/dsp/dsp.html#time_window) also raise an error.

tests/test_dsp.py
```diff
@@ -682,6 +682,83 @@ def test_time_window_return_window_error():
...
+    dsp.time_crop(signal_2, interval = (1, 2))
+    signal_3 = "signal"
+    with pytest.raises(TypeError, match='has to be of type'):
+        dsp.time_crop(signal_3, interval=(1, 2))
+
+
+@pytest.mark.parametrize("signal", [pyfar.Signal(np.ones(10), 2),
+                                     pf.TimeData([1, 2], [1, 2])])
+@pytest.mark.parametrize("interval", [(1, 2), [1, 2], np.array([1, 2])])
+
```
jf-som:
```suggestion
```

pyfar/dsp/dsp.py
```diff
@@ -561,6 +563,80 @@ def time_window(signal, interval, window='hann', shape='symmetric',
...
+    """
+    # Check the input
+    if not isinstance(signal, (pyfar.Signal, pyfar.TimeData)):
+        raise TypeError("The parameter signal has to be of type " \
+        "pyfar.Signal or pyfar.TimeData.")
+
+    if not isinstance(interval, (list, tuple, np.ndarray)):
+        raise TypeError(
+            "The parameter interval has to be of type" \
+            " list, tuple or numpy.array.")
```
jf-som:
```suggestion
            " list, tuple or numpy.ndarray.")
```
np.array is a function, you probably meant ndarray

pyfar/dsp/dsp.py
```diff
@@ -561,6 +563,116 @@ def time_window(signal, interval, window='hann', shape='symmetric',
...
+    interval = np.array(interval)
+    start, end = interval[0], interval[1]
+    if start >= end:
+        raise ValueError("The interval start point must be smaller " \
+        "than the end point.")
+    if start < 0 or end < 0:
+        raise ValueError("The interval boundaries must be non-negative.")
+
+    if unit == 's':
+            times = np.array(signal.times)
```
sikersten: Isn't this line obsolete, since Signal and TimeData always return an np.array?
```suggestion
```

pyfar/dsp/dsp.py
```diff
@@ -561,6 +563,116 @@ def time_window(signal, interval, window='hann', shape='symmetric',
...
+            mask = ((times >= interval[0]) & (
+                times <= interval[1]))
+            # Find the indices of the time instances that do not exceed
+            # the boundaries of the interval.
+            times_indices = np.nonzero(mask)[0]
+            if times_indices.size > 0 :
+                interval = (times_indices[0], times_indices[-1])
+            else:
+                raise ValueError("Interval is out of the boundaries" \
+                " of signal.times")
```
sikersten: Here, the function's behavior is different for seconds vs. samples. With `samples `out of the boundaries, an empty signal is returned, while an error is raised with seconds. I think both solutions are fine, it should just be the same behavior for both units. I would suggest raising an error for both, since an empty object is not really on any use.

pyfar/dsp/dsp.py
```diff
@@ -561,6 +563,116 @@ def time_window(signal, interval, window='hann', shape='symmetric',
...
+        Unit of `interval`. Can be set to ``'samples'`` or ``'s'`` (seconds).
+        Time values are rounded to the nearest sample. The default is
+        ``'samples'``.
+
+    Returns
+    -------
+    cropped_signal : pyfar.Signal, pyfar.TimeData
+        Cropped pf.Signal or pf.TimeData object.
+
+    Examples
```
sikersten, line 606: Very nice example! I would suggest to split it up into two, one for a Signal and one for TimeData (each with two panels in the plot). A short sentence before each example could highlight the difference.

pyfar/dsp/dsp.py
```diff
@@ -561,6 +563,116 @@ def time_window(signal, interval, window='hann', shape='symmetric',
         return signal_win
 
 
+def time_crop(signal, interval: Union[list[float], tuple[float, float],
+                                       np.ndarray],
+              unit: Literal["samples", "s"]='samples'):
+    """Crop a pf.Signal or pf.TimeData object in time.
```
sikersten:
```suggestion
    """Crop a Signal or TimeData object in time.
```

pyfar/dsp/dsp.py
```diff
@@ -561,6 +563,116 @@ def time_window(signal, interval, window='hann', shape='symmetric',
...
+
+    unit : string, optional
+        Unit of `interval`. Can be set to ``'samples'`` or ``'s'`` (seconds).
+        Time values are rounded to the nearest sample. The default is
+        ``'samples'``.
+
+    Returns
+    -------
+    cropped_signal : pyfar.Signal, pyfar.TimeData
+        Cropped pf.Signal or pf.TimeData object.
```
sikersten:
```suggestion
        Cropped Signal or TimeData object.
```

pyfar/dsp/dsp.py
```diff
@@ -561,6 +563,116 @@ def time_window(signal, interval, window='hann', shape='symmetric',
...
 
+def time_crop(signal, interval: Union[list[float], tuple[float, float],
+                                       np.ndarray],
+              unit: Literal["samples", "s"]='samples'):
+    """Crop a pf.Signal or pf.TimeData object in time.
+
+    Parameters
+    ----------
+    signal : pyfar.Signal, pyfar.TimeData
+        pf.Signal or pf.TimeData object to be cropped.
```
sikersten:
```suggestion
        Signal or TimeData object to be cropped.
```

pyfar/dsp/dsp.py
mberz:
```suggestion
            times = signal.times
```
I think what you mean would be this then?

pyfar/dsp/dsp.py
mberz: I think moving the check for the valid interval to below the if cases extracting the intervals makes sense here

pyfar/dsp/dsp.py
```diff
@@ -566,7 +566,7 @@ def time_window(signal, interval, window='hann', shape='symmetric',
 def time_crop(signal, interval: Union[list[float], tuple[float, float],
                                        np.ndarray],
               unit: Literal["samples", "s"]='samples'):
-    """This function can be used to crop a pf.Signal or pf.TimeData object.
+    """Crop a pf.Signal or pf.TimeData object in time.
```
mberz: I would suggest to add a sentence on the definition of the crop interval used by the function, i.e. that the function crops to the inside of the interval.

pyfar/dsp/dsp.py
```diff
@@ -597,41 +629,45 @@ def time_crop(signal, interval: Union[list[float], tuple[float, float],
...
             "The parameter interval must consist of 2 entries.")
 
-    # Convert to samples
     interval = np.array(interval)
+    start, end = interval[0], interval[1]
+    if start >= end:
+        raise ValueError("The interval start point must be smaller " \
+        "than the end point.")
+    if start < 0 or end < 0:
+        raise ValueError("The interval boundaries must be non-negative.")
```
mberz: Minor comment: This could be simplified to the following:

```suggestion
    if np.diff(interval) <= 0:
        raise ValueError(
            "The interval start point must be smaller than the end point.")
    if np.any(interval < 0):
        raise ValueError("The interval boundaries must be non-negative.")
```

pyfar/dsp/dsp.py
h-chmeruk: Thank you both for your reviews!
Unfortunately, I did not remove redundant code from the function in the previous commit. In the new commit, the interval is no longer updated after the mask is created, and exceeding the interval boundaries is checked in the same way for both cases, `unit='s'` and `unit='samples'`. 

```
    if unit == 's':
        indices = signal.times
    elif unit == 'samples':
        interval = interval.astype(int)
        indices = np.arange(len(signal.times))
    else:
        raise ValueError(f"unit is {unit} but has to be 'samples' or 's'.")

    mask = ((indices >= interval[0]) & (
            indices <= interval[1]))

    # If there are no True values in the mask, the interval lies
    # outside the boundaries of signal.times.
    if not mask.any():
         raise ValueError("Interval is out of the boundaries" \
                " of signal.times")
```

jf-som (approved):
As discussed in the weekly, I consider this ready for merge. 🙌 

As a last minor edit, it would be nice if you could add a short hint to the different `signal.times` behavior between Signal and TimeData input types before the merge, like we discussed in the weekly.

Thanks for implementation!

pyfar/dsp/dsp.py
```diff
@@ -561,6 +563,134 @@ def time_window(signal, interval, window='hann', shape='symmetric',
         return signal_win
 
 
+def time_crop(signal, interval: Union[list[float], tuple[float, float],
+                                       np.ndarray],
+              unit: Literal["samples", "s"]='samples'):
+    """Crop a :py:class:`Signal <pyfar.Signal>` or a
+    :py:class:`TimeData <pyfar.TimeData>` object in time.
+    Returns the signal within the specified `interval`.
```
ahms5:
```suggestion

    Returns the signal for all :math:`x` within the interval
    :math:`t_{beginning} \le t < t_{end}`, where :math:`x`
    can be time or samples.
```

maybe sth like this? we can leave the first sentence as short summary, and the remaining explainations is then providing more details, also see https://numpydoc.readthedocs.io/en/stable/format.html#short-summary

ahms5 (commented):
approved, execpt hint about the interval, feel free to ajust my recommendation.
Thank you Hanna.

PR #851, opened 2025-10-17
Feature/time weighting

### LINKED ISSUE #760: feature: time weighted sound pressure level

### LINKED ISSUE #925: sub-feature: time_weighted_level convenience function
add a function to compute the slow or fast time weighting according to the norm with A/C/Z frequency weighting and optionally in frequency bands.

jf-som:

Closes #760 and #925

- Adding functions for time weighted levels according to ISO 61672-1


jf-som:
Hey everyone. This is a draft implementing all of #675 except the calibration. The six high-level functions for calculating levels are in `standard_levels.py`.
While I'm still working on tests and documentation, there are a few things we could take a look at already, which are the following:
1) should these functions have some kind of `channel_handling` parameter? As it is currently, a multi-channel input will calculate levels for each channel independently and return them all. So far I didn't include it since it would be quite a bit of extra work/testing, and I assume actual SPL measurements will mostly be mono anyway. Also it's easy to do for the users themselves if they ever need to combine channels (like preprocessing with `pf.average` or combining the output with my new utility `pf.level.average_levels` in utils.py).

2) The functions `peak_level()` and `maximum_time_weighted_level()` currently have an `oversampling` parameter that I added. I think this makes sense, because an analog SPL meter ideally has infinite time resolution, and that is also what the standard aims at. 
The loudness normalization standard EBU R 128 finds inter-sample-peaks through oversampling, so this is what I used as reference. You can view it here: https://www.itu.int/dms_pubrec/itu-r/rec/bs/R-REC-BS.1770-5-202311-I!!PDF-E.pdf It also provides a table on page 23 showing the maximum error depending on the oversampling factor used. With a factor of 4, the theoretical maximum error is about 0.69 dB, which is probably acceptable, and can only happen with very loud frequency components near the Nyquist frequency in the first place.
For convenience, I also return the timestamp of the peak level found (per channel). With oversampling, this has to be a time in seconds rather than a sample index though.
Let me know what you think about the oversampling idea and which default value you might recommend instead (e.g. 8, or `None` for no oversampling)

f-brinkmann:
@jf-som thx for the implementation effort. It will be great to have the level functions in pyfar. However, the PR is quite big which might make it hard and long to review. Do you see possibilities to split this into smaller PRs?

jf-som:
Splitting the review is probably a good idea, but with my code structure it won't be possible to make independent branches/PRs. 
So I'll copy the core parts of this branch over to a new one and then gradually add features back in once the core is approved/merged.

jf-som:
I split the PR into smaller bits now. Please see #949, which we should use for discussion from now on.
I will still use this branch here as reference though, so please don't delete any branch or this PR yet.

PR #852, opened 2025-10-29
Feature/cache frequency weighting filter

jf-som:

none, but it was originally mentioned in #811 but then saved for later to declutter the already long discussion there.

- make the frequency weighting filter design cache its coefficients to skip iterative approach for subsequent calls with the same arguments.
- this speeds up the function when using it repeatedly to filter Signal objects directly

With the current implementation of `pf.dsp.filter.frequency_weighting_filter()`, code like the following will be slower than necessary, because the algorithm to iteratively approximate the filter coefficients is run for every call.

Example:

```py
import pyfar as pf
import time

signal = pf.signals.impulse(2048)

def cached():
    for i in range(100):
        _ = pf.dsp.filter.frequency_weighting_filter(signal)
        
def uncached():
    for i in range(100):
        # this is equivalent to the current implementation
        _ = pf.dsp.filter.frequency_weighting_filter(
            signal, bounds=(-1, 1))
    
def time_func(function):
    t0 = time.time()
    function()
    t1 = time.time()
    print(round(t1 - t0, ndigits=3), "s")
    
time_func(cached)
time_func(uncached)
```
Outputs on my machine:
```
0.055 s
4.715 s
```

Since `kwargs` and functions like `error_weighting` are not hashable, this solution only caches the results when neither argument was given by the user (which I expect to be the case most of the time). 
You can see the effect here, where giving the `bounds` kwarg disables caching.

### DIFF
--- a/pyfar/dsp/filter/frequency_weighting.py
+++ b/pyfar/dsp/filter/frequency_weighting.py
@@ -38,10 +38,9 @@ def frequency_weighting_filter(
         This function will run a least-squares algorithm to iteratively
         approach the filter coefficients for the target weighting curve.
         The coefficients are cached, but only when the `error_weighting`
-        and ``kwargs`` arguments are unused, which are unhashable.
-        Therefore, it may be much faster to create the filter once and
-        reuse that filter when calling this function repeatedly
-        with an `error_weighting` or `kwargs`.
+        and `kwargs` arguments are unused, which are unhashable.
+        When using these parameters, it is faster to create the filter once
+        and reuse it than calling this function repeatedly.
 
     .. note::
         When using default parameters for `n_bins` and `error_weighting`


pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -35,12 +36,15 @@ def frequency_weighting_filter(
 
     .. note::
         This function will run a least-squares algorithm to iteratively
-        approach the target weighting curve. Therefore, it may be much
-        faster to create the filter once and reuse that filter when
-        repeatedly weighting audio data of identical sampling rates.
+        approach the filter coefficients for the target weighting curve.
+        The coefficients are cached, but only when the ``error_weighting``
```
f-brinkmann: Our style guidelines suggest to use italic font for parameters. Can you please change this here and below?
```suggestion
        The coefficients are cached, but only when the `error_weighting`
```

tests/test_frequency_weighting.py
```diff
@@ -141,3 +141,26 @@ def test_frequency_weighting_filter_errors():
     with pytest.warns(UserWarning, match=re.compile(regex)):
         pffilt.frequency_weighting_filter(signal, "A",
                                           error_weighting=lambda _: 0)
+
+
+def test_frequency_weighting_filter_caching():
```
f-brinkmann, line 146: Can you add a brief docstring to explain what this does. Might help if this is revisited at some point.

mberz:
Thanks for your suggestion. 
I however have the same impression as @sikersten here.
The use case for filters which are re-used more than once is to return and store the filter object instead of re-creating it each time. The filter can then be applied to different signals or signals representing a continuous input stream, i.e. with support for resetting or keeping the filter state.

jf-som:
Your example is what I meant by the comment
```py
    # One could define the filter globally or pass it to every 
    # call to process_signal(), but both break the abstraction or encapsulation
    # that functions provide a little.
```
Of course it's *possible* to pass the filter and maybe a `reset` flag every time, but it doing so makes the function more complex for something that could happen automatically.

Anyway, I feel like we are getting lost in the details here. My intent was simply to add a performance upgrade to make the user experience more consistent with other function in dsp.filter (Other filters are created so quickly that reusing them effectively doesn't matter, which I wouldn't say about the frequency weighting filter unless we use caching), without any downsides for the user.

sikersten:
It appears to me that this is more a design question rather an implementation one, which would be easier to discuss in one of or weekly meetings (Fridays, 2 pm). @jf-som, would be available and interested in joining for a meeting in the upcoming weeks? ☺️

ahms5:
small correction: merge it as soon as it has at least 2 approves.

pyfar/dsp/filter/frequency_weighting.py
jf-som: Fixed it here and two more times in the same paragraph

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -37,11 +37,11 @@ def frequency_weighting_filter(
     .. note::
         This function will run a least-squares algorithm to iteratively
         approach the filter coefficients for the target weighting curve.
-        The coefficients are cached, but only when the ``error_weighting``
+        The coefficients are cached, but only when the `error_weighting`
         and ``kwargs`` arguments are unused, which are unhashable.
         Therefore, it may be much faster to create the filter once and
         reuse that filter when calling this function repeatedly
-        with an ``error_weighting`` or ``kwargs``.
+        with an `error_weighting` or `kwargs`.
```
f-brinkmann: This makes it more clear to me
```suggestionx
       When using these parameters, it is faster  to create the
       filter once and reuse it than calling this function repeatedly.
```

pyfar/dsp/filter/frequency_weighting.py
```diff
@@ -37,11 +37,11 @@ def frequency_weighting_filter(
     .. note::
         This function will run a least-squares algorithm to iteratively
         approach the filter coefficients for the target weighting curve.
-        The coefficients are cached, but only when the ``error_weighting``
+        The coefficients are cached, but only when the `error_weighting`
         and ``kwargs`` arguments are unused, which are unhashable.
```
f-brinkmann: Should also use single quotes

```suggestion
        and `kwargs` arguments are unused, which are unhashable.
```

pyfar/dsp/filter/frequency_weighting.py
jf-som: I like your wording and updated the comment!

pyfar/dsp/filter/frequency_weighting.py
jf-som: good catch

Issue #854, opened 2025-11-07
[Bug] Missing pyfar-errors for wrong unit input in `freq_group_delay_2d` and `group_delay_2d`


hoyer-a:
Came up in #853 

Entering a wrong unit in the plot-functions should raise an error like:
`ValueError: Unit is bla but must be s, ms, mus, samples, auto.`

But for `group_delay_2d` and `freq_group_delay_2d` this is not the case. 
- `group_delay_2d` raises a python-error for wrong list index in list.index() (`list.index(x): x not in list`)
- `freq_group_delay_2d` seems to raise no error at all

PR #859, opened 2025-11-07
Fix warnings in tests

f-brinkmann:

Fixes all warnings in tests except
- warnings for fractional octave frequencies are fixed in #856
- a bug in casting complex to real valued Signals and the related warning #858
- warnings in texting the TransmissionMatrix #860 

See comments for details.

I thought this would still be fine as one PR. I could also split it in 5 PRs. One for each commit.

### DIFF
--- a/tests/test_coordinates_arithmetic.py
+++ b/tests/test_coordinates_arithmetic.py
@@ -69,19 +69,18 @@ def test_arithmetic_coordinates(other, operator, order):
     ],
 )
 def test_arithmetic_coordinates_mul_div(other, operator, order):
-    coords = pf.Coordinates([0, 1], [0, 1], [0, 1])
+    coords = pf.Coordinates([2, 1], [5, 1], [10, 1])
     part1 = coords if order else other
     part2 = other if order else coords
     if operator == "*":
         new = part1 * part2
-        desired = pf.Coordinates([0, 5], [0, 5], [0, 5])
+        desired = pf.Coordinates([10, 5], [25, 5], [50, 5])
     elif operator == "/":
         new = part1 / part2
         if order:
-            desired = pf.Coordinates([0, .2], [0, .2], [0, .2])
+            desired = pf.Coordinates([2/5, .2], [1, .2], [2, .2])
         else:
-            desired = pf.Coordinates(
-                [np.inf, 5], [np.inf, 5], [np.inf, 5])
+            desired = pf.Coordinates([2.5, 5], [1, 5], [.5, 5])
     assert isinstance(new, pf.Coordinates)
     npt.assert_array_equal(new.cartesian, desired.cartesian)
 


tests/test_filter_classes.py
```diff
@@ -343,15 +343,15 @@ def test_filter_iir_process_complex(impulse_complex):
 
 def test_filter_iir_process_state(impulse):
     coeff = np.array([[1, 1/2, 0], [1, 0, 0]])
-    filt = fo.FilterIIR(coeff, impulse.sampling_rate, state=[0, 0])
+    filt = fo.FilterIIR(coeff, impulse.sampling_rate, state=[[[0, 0]]])
```
f-brinkmann, line 346: Here and below, there was a warning because the filter state was provided in the wrong shape.

tests/test_io.py
```diff
@@ -705,7 +705,7 @@ def test_audio_subtypes(audio_subtypes_mock):
 def test_default_audio_subtype(default_audio_subtype_mock):
     """Test correct call of the wrapped soundfile function."""
     audio_format = 'wav'
-    subtype_return = pyfar.io.default_audio_subtype(format=audio_format)
+    subtype_return = pyfar.io.default_audio_subtype(audio_format=audio_format)
```
f-brinkmann, line 708: The parameter `format` is deprecated but was still used.

tests/test_plot.py
```diff
@@ -132,8 +132,11 @@ def test_complex_freq_plots_2d(function, side_flag,
     functionality.
     """
 
-    signal = pf.utils.concatenate_channels((
-        handsome_complex_signal, handsome_complex_signal_v2))
+    if function == pf.plot.spectrogram:
+        signal = handsome_complex_signal
```
f-brinkmann, line 136: this is to avoid a warning if passing multi-channel signals to spectrogram

tests/test_deprecations.py
```diff
@@ -227,33 +227,42 @@ def test_deprecation_shelve_functions():
...
-                                 sampling_rate=44100)[0].coefficients,
-        rtol=0)
-
-    np.testing.assert_allclose(
-        pfilt.low_shelve_cascade(None, 250, "lower", -60, None, 4,
-                                 sampling_rate=44100)[0].coefficients,
-        pfilt.low_shelf_cascade(None, 250, "lower", -60, None, 4,
-                                sampling_rate=44100)[0].coefficients,
-        rtol=0)
+    with pytest.warns(PyfarDeprecationWarning,
```
f-brinkmann, line 230: This is for testing if the behavior of the new functions are identical of the old functions. But we did not catch the deprecation warnings before.

tests/test_audio_signal.py
```diff
@@ -205,10 +205,12 @@ def test_n_samples():
     signal = Signal([1, 2, 3], 44100, domain='time', is_complex=True)
     assert signal.n_samples == 3
 
-    signal = Signal([1, 2, 3], 44100, domain='freq')
+    with pytest.warns(UserWarning, match='Number of samples not given'):
```
f-brinkmann, line 208: This tests if `n_samples` is set correctly but did not catch the UserWarning

tests/test_frequency_weighting.py
```diff
@@ -8,8 +8,7 @@
 
 @pytest.mark.parametrize("weighting", ["A", "C"])
 @pytest.mark.parametrize("num_fractions", [1, 3])
-@pytest.mark.parametrize("freq_range", [(20, 20000), (0, 100),
-                                        (4000, 100_000)])
+@pytest.mark.parametrize("freq_range", [(20, 20000)])
```
f-brinkmann, line 11: Splitted these into two tests. One that tests a frequency range where no warnings are emitted and one that tests the warnings

tests/test_coordinates_arithmetic.py
```diff
@@ -76,10 +76,12 @@ def test_arithmetic_coordinates_mul_div(other, operator, order):
         new = part1 * part2
         desired = pf.Coordinates([0, 5], [0, 5], [0, 5])
     elif operator == "/":
-        new = part1 / part2
```
f-brinkmann, line 79: This causes a divide by that was not caught.

tests/test_coordinates_arithmetic.py
```diff
@@ -76,10 +76,12 @@ def test_arithmetic_coordinates_mul_div(other, operator, order):
...
         if order:
+            new = part1 / part2
             desired = pf.Coordinates([0, .2], [0, .2], [0, .2])
         else:
+            with pytest.warns(RuntimeWarning, match='divide by zero'):
```
ahms5: I think we should not check Warnings and errors from other packages, It would be better in this case to just prevent this warning by choosing Coordinates which are not zero, because the idea of the test is, to check the correct behaviour for valid operations, and dividing by 0 is not valid.

tests/test_deprecations.py
```diff
@@ -227,33 +227,42 @@ def test_deprecation_shelve_functions():
...
+            rtol=0)
+
+    with pytest.warns(PyfarDeprecationWarning,
+                      match="'low_shelve' will be deprecated in pyfar 0.9.0"):
+        np.testing.assert_allclose(
+            pfilt.low_shelve(None, 10e3, 1.5, 1, "I",
+                            sampling_rate=44100).coefficients,
+            pfilt.low_shelf(None, 10e3, 1.5, 1, "I",
+                            sampling_rate=44100).coefficients,
+            rtol=0)
```
ahms5, line 246:
```suggestion
        np.testing.assert_allclose(
            array([[[ 1.08735675,  1.10156456,  0.01420782],
                    [ 1.        ,  0.92685107, -0.07314893]]]),
            pfilt.low_shelf(None, 10e3, 1.5, 1, "I",
                            sampling_rate=44100).coefficients,
            rtol=0)
```

I would suggest to manually add the results from the privous implementation as knows truth values. At some point these need to be removed, and the test need to be changed anyways.

ahms5 (changes requested):
nice, thank you. I have 2 sugesstions to further improve it. Please comment, if there are other opintions.

tests/test_deprecations.py
hoyer-a: Not sure if this is necessary, since the functions are wrappers for the same private function and apart from the name exactly alike. I don't think we test any of the filters in dsp.filter for results.

ahms5: Agree, if the private method is already tested.
I would be careful with what we did in the past; sometimes we did not follow our own testing guidelines. It some cased it might make sense to rethink previous decisions.

tests/test_coordinates_arithmetic.py
ahms5: btw this was also the issue found in #853 and moved to #854

tests/test_coordinates_arithmetic.py
```diff
@@ -69,19 +69,18 @@ def test_arithmetic_coordinates(other, operator, order):
     ],
 )
 def test_arithmetic_coordinates_mul_div(other, operator, order):
-    coords = pf.Coordinates([0, 1], [0, 1], [0, 1])
+    coords = pf.Coordinates([2, 1], [5, 1], [10, 1])
```
f-brinkmann, line 72: Update coordinates object to avoid division by zero in tests

tests/test_coordinates_arithmetic.py
f-brinkmann: changed the coordinates to avoid the warning in the first place as suggested

tests/test_deprecations.py
f-brinkmann: Left this as this is according to the discussion

ahms5 (approved):

sikersten (approved):
Thank you for taking care!

PR #866, opened 2025-11-13
Use CircleCI trigger to create GitHub releases

mberz:

- Create a GitHub release from CircleCI on version tags after successful publication on PyPI.

Additional changes
- Separate building the package from uploading
- Use shared workspace folders to re-use the source and binary distributions from the build step


mberz:
related to #865

Issue #600, opened 2024-05-02
Move show functions to plot module


f-brinkmann:
## Description

The coordinates and orientations classes have show methods to plot their geometry. Those should be moved to the plot module.

## Checklist for Discussion

- [ ] Names: `pf.plot.coordinates` and `pf.plot.orientations()`?
- [ ] They should probably work as all pyfar plots and use the current axis. Since both plots require 3D axes, this might need some additional work when setting up the plot.

ahms5:
for the coordinates the show method is just a wrapper for ``pf.plot.scatter``, so maybe we can just deprecate it.
for the orientation we should shift it to pf.plot 
maybe related to #639

PR #872, opened 2025-12-01
BUG: scatter and quiver were hidden in the docs

ahms5:

- add scatter and quiver  to the docs.

Maybe there was a reason, why they were not added, but I suggest to add them the the documentation

### DIFF
--- a/docs/modules/pyfar.plot.rst
+++ b/docs/modules/pyfar.plot.rst
@@ -5,4 +5,3 @@ pyfar.plot
    :members:
    :undoc-members:
    :show-inheritance:
-   :exclude-members: [, scatter, quiver,]


f-brinkmann:
I think we removed them on purpose with #600 in mind but never followed up. I agree with hoyer-a that it might be better not to add them to the docs at this stage.

mberz:
> I would revise the method and deprecate `Coordiantes.show()` and `Orientations.show()`. In this way we can align it with the logic for Signal plotting. and scatter plot also supports Coordiantes, but not very elegantly.
> 
> lets collect ideas here and discuss it in the weekly.

Please do not plan features or changes in pull requests. That's what issues are for. 
Feel free to continue the discussion in issue #600 or open a new one if #600 is not fitting. Since it seems that the change suggested in this PR is not what we want to go for I'm closing it.

PR #874, opened 2026-01-01
Bugfix: Inconsistent returns for fractional_octave_frequencies_exact

### LINKED ISSUE #857: Inconsistent returns for new fractional octave frequencies functions
When working on #856, I noticed the following:

```python
frequncies = pf.constants.fractional_octave_frequencies_nominal(
    num_fractions=1, frequency_range=(20, 200))
```
returns `[31.5, 63, 125]`, which is consistent with `dsp.filter.fractional_octave_frequencies`. But

```python
f = pf.constants.fractional_octave_frequencies_exact(
    num_fractions=1, frequency_range=(20, 200))[0]
```
returns `[ 15.84  31.62   63.09 125.89 251.18]`, which includes frequencies below and above the specified range.

h-chmeruk:

Closes #857, #885

The exact center frequencies now correspond to the nominal frequencies, which means that for the same frequency range, the number of exact frequencies does not exceed the number of nominal frequencies. 

The nominal frequencies are now masked with the `lower_cutoff_frequencies` from `fractional_octave_frequencies_exact` for the lower limit and with the `upper_cutoff_frequencies` from `fractional_octave_frequencies_exact` for the upper limit. 

The calculation of the indices in the function `fractional_octave_frequencies_exact` has also been changed in order to achieve the most accurate assignment of limit frequencies `frequency_range[0]` and  `frequency_range[1]` to frequency bands.

I also added some additional tests to test more cases with the new behavior.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -48,6 +48,7 @@ Removed
 Fixed
 ^^^^^
 - Fix warnings and errors in tests (PR #859, #897)
+- Update the calculation of nominal and exact center frequencies in `pyfar.constants.fractional_octave_frequencies_exact` and `pyfar.constants.fractional_octave_frequencies_nominal` (PR #874)
 
 
 0.7.4 (2026-02-02)


pyfar/constants/constants.py
```diff
@@ -700,10 +707,23 @@ def fractional_octave_frequencies_exact(
...
+                                                    frequency_range)
+        if nominal.size > 0:
+            mask = (upper_cutoff_frequencies>=nominal[0]) & (
+                lower_cutoff_frequencies<=nominal[-1])
+        else:
+            mask = np.full(center_frequencies.shape, False)
+        center_frequencies = center_frequencies[mask]
+        lower_cutoff_frequencies = lower_cutoff_frequencies[mask]
+        upper_cutoff_frequencies = upper_cutoff_frequencies[mask]
+
```
ahms5:
```suggestion

    mask = (center_frequencies >= f_lims[0]) & (center_frequencies <= f_lims[1])
    center_frequencies = center_frequencies[mask]
    lower_cutoff_frequencies = lower_cutoff_frequencies[mask]
    upper_cutoff_frequencies = upper_cutoff_frequencies[mask]

```
I'm afraid this will lead to inconsistent behaviour because the behaviour may vary if num_faction is set to 1 or 2. Why don't we just use the same logic for the limit as in the nominal method?
I suggest adding an additinal mask at the end.

pyfar/constants/constants.py
```diff
@@ -700,10 +707,23 @@ def fractional_octave_frequencies_exact(
         # IEC 61260-1 Eq. (3)
         center_frequencies = ref_freq * (G)**((2*indices + 1)/
                                               (2*num_fractions))
-
     # IEC 61260-1 Eq. (5)
     upper_cutoff_frequencies = center_frequencies * G**(1/2/num_fractions)
     # IEC 61260-1 Eq. (4)
     lower_cutoff_frequencies = center_frequencies * G**(-1/2/num_fractions)
+
```
ahms5, line 752: To ensure that ``Nmin`` and ``Nmax`` are within the expected range, we increase ``Nmax`` and decrease ``Nmin`` by one.

tests/test_constants_center_frequencies.py
```diff
@@ -64,6 +64,22 @@ def test_exact_octave_subset():
...
+
+
+def test_exact_third_octave_subset():
+    exact = constants.fractional_octave_frequencies_exact(
+                                        3, (12, 62))[0]
+    expected = np.array([12.58925412, 15.84893192, 19.95262315, 25.11886432,
+                          31.6227766, 39.81071706, 50.11872336])
+    npt.assert_allclose(exact, expected)
+
+
```
ahms5, line 131: In addition we could also test directly if ``fractional_octave_frequencies_exact`` and ``fractional_octave_frequencies_nominal`` returns similar values for octave and one-third octave bands.

pyfar/constants/constants.py
h-chmeruk: Thank you for your review!

The problem with this implementation is the same as the one we discussed in the Issue #857. For certain frequencies, the nominal frequency is included in the output, but the corresponding exact frequency is no longer included because it is smaller or larger than one of the range points. For example, for `num_fractions=1` and `frequency_range=(16, 500)`, `fractional_octave_frequencies_nominal` would return` [16, 31.5, 63. 125, 250, 500]`, and with the proposed implementation, `fractional_octave_frequencies_exact` would return `[31.6, 63.1, 125.9, 251.2]`, since 15.8<16 and 501.2>500.

pyfar/constants/constants.py
h-chmeruk: Sorry, I didn't quite understand. Where should Nmin and Nmax be decreased and increased?
Changing Nmin and Nmax would also change the indices for calculating the frequencies, which would lead to incorrect results.

tests/test_constants_center_frequencies.py
h-chmeruk: Thank you for the suggestion, I'll add this :)

pyfar/constants/constants.py
ahms5: True, thanks for the clarification. 
sorry my bad, I did not see the latest update in the issue.

h-chmeruk:
[f-brinkmann](https://github.com/f-brinkmann), thank you for the suggestion! 

I wasn't entirely satisfied with the implementation either. I think it's a good idea to add another check so that the exact frequencies are also returned outside the nominal frequency range. The question is whether this would be the expected behavior and whether the implementation would still be maintainable and readable. 

I feel that by adapting the output of the `fractional_octave_frequencies_exact` function to the `fractional_octave_frequencies_nominal` function, we are making the `fractional_octave_frequencies_exact` function less basic and less universal, and too focused on special cases. Maybe we should just separate the functionality and leave the `fractional_octave_frequencies_exact` function as it is, and create a special function with exact frequencies that correspond to the nominal frequencies...

Should we mark this PR as a draft and return to the Issue #857 discussion?

pyfar/constants/constants.py
ahms5: this was related to the other comment, so it is obsolete

pyfar/constants/constants.py
```diff
@@ -575,6 +575,13 @@ def fractional_octave_frequencies_nominal(num_fractions:Literal[1,3]=1,
     nominal : numpy.ndarray of float
         The nominal center frequencies.
 
+    Notes
+    -----
+    The specified ``frequency_range`` is interpreted as frequencies lying
+    within octave bands, not as strict lower and upper center-frequency
+    limits. All bands whose frequency intervals overlap with the
+    specified frequency range are included in the result.
+
```
f-brinkmann, line 584: This would make it a more clear to me:
```suggestion
    Notes
    -----
    The specified ``frequency_range`` is interpreted as frequencies lying
    within (fractional) octave bands defined by their cut-off frequences
    (not their center frequencies). All bands that overlap with the
    specified frequency range are returned.

```

pyfar/constants/constants.py
```diff
@@ -685,6 +694,13 @@ def fractional_octave_frequencies_exact(
     .. [#] International Electrotechnical Commission, "IEC 61260-1:2014 -
         Electroacoustics - Octave-band and fractional-octave-band filters -
         Part 1: Specifications", IEC, 2014.
+
+    Notes
+    -----
+    The specified ``frequency_range`` is interpreted as frequencies lying
+    within octave bands, not as strict lower and upper center-frequency
+    limits. All bands whose frequency intervals overlap with the
+    specified frequency range are included in the result.
```
f-brinkmann: see above

tests/test_constants_center_frequencies.py
```diff
@@ -90,6 +106,22 @@ def test_exact_octave_subset():
...
+def test_exact_third_octave_subset():
+    exact = constants.fractional_octave_frequencies_exact(
+                                        3, (12, 62))[0]
+    expected = np.array([12.58925412, 15.84893192, 19.95262315, 25.11886432,
+                          31.6227766, 39.81071706, 50.11872336, 63.09573445])
+    npt.assert_allclose(exact, expected)
+
+
```
ahms5, line 131:
```suggestion

@pytest.mark.parametrize(
    "frac",
    [1, 2, 3])
@pytest.mark.parametrize(
    "frequency_range",
    [
        (500, 1000),
    ])
def test_exact_limits_within_bands(frac, frequency_range):
    exact, lower, upper = constants.fractional_octave_frequencies_exact(
        frac, frequency_range)

    # check if the frequency range limits are within the calculated bands
    npt.assert_allclose(
        lower[0] <= frequency_range[0] <= upper[0], True)
    npt.assert_allclose(
        lower[-1] <= frequency_range[1] <= upper[-1], True)
```

maybe I would add a test like this, were we can test different parameters, and see if the limits are within the bands

h-chmeruk: Thanks for mentioning this! I've already created an [Issue](https://github.com/pyfar/pyfar/issues/885) for the problem that the calculation of the indices is incorrect when the number of fractions is even. Should I perhaps first create a PR for the Issue #885 and merge it, and then we can merge this PR?

h-chmeruk: Or I could just add this new functionality to this PR, but that's kind of a different issue, so maybe a separate PR is better...

ahms5: up to you, you have the best overview, sounds like one fix depend on the other

ahms5 (approved):

pyfar/constants/constants.py
```diff
@@ -594,6 +601,8 @@ def fractional_octave_frequencies_nominal(num_fractions:Literal[1,3]=1,
     if f_lims[0] > f_lims[1]:
         raise ValueError(
             "The second frequency needs to be higher than the first.")
+    _, lower, upper = fractional_octave_frequencies_exact(num_fractions,
+                                                           frequency_range)
```
f-brinkmann: I think there too much indention

mberz:
@h-chmeruk I see that this PR has two approves. Please let me know when you finish working. (I didn't want to merge since it seems you're still working on it)

h-chmeruk:
@mberz, oh, sorry, I just added an entry to the HISTORY.rst and rebased the branch to the current state of develop. I consider it finished, unless someone else insists on reviewing it again.

PR #876, opened 2026-01-05
Add input validation for fractional_octave_frequencies_exact 

### LINKED ISSUE #875: Missing input validation in pf.constants.fractional_octave_frequencies_exact
While implementing the bug fix for Issue #857, I noticed that the
`pyfar.constants.fractional_octave_frequencies_exact` function lacks some input
validation. In particular, there are no checks to ensure that
`frequency_range` and `num_fractions` are provided with valid values.

Here is a suggestion for additional checks that could be added to the function:

```
    if not isinstance(frequency_range, (int, float)):
        raise ValueError('The frequency range must be a tuple of float '
                         'or integer numbers')
    if not isinstance(num_fractions, (int)):
        raise ValueError('Number of fractions must be an integer')

    f_lims = np.asarray(frequency_range)
    if f_lims.size != 2:
        raise ValueError(
            "You need to specify a lower and upper limit frequency.")
    if np.diff(f_lims) <= 0:
        raise ValueError(
    if np.any(f_lims <= 0):
        raise ValueError(
            "The frequencies must be positive numbers.")
    if num_fractions <= 0:
            raise ValueError(
            "Number of fractions must be a positive number.")
```

h-chmeruk:

Closes #875

- Added input checks for the parameter `frequency_range`
- Added input checks for the parameter `num_fractions`
- Added 4 new test functions to check whether TypeErrors and ValueErrors are raised

### DIFF
--- a/pyfar/constants/constants.py
+++ b/pyfar/constants/constants.py
@@ -687,7 +687,7 @@ def fractional_octave_frequencies_exact(
         Part 1: Specifications", IEC, 2014.
     """
     if not isinstance(frequency_range, (tuple, np.ndarray, list)):
-        raise TypeError("The frequency range must be a tuple, list or" \
+        raise TypeError("The frequency range must be a tuple, list or"
         " np.ndarray of float or integer values.")
     if not isinstance(num_fractions, (int)):
         raise TypeError("Number of fractions must be an integer.")


pyfar/constants/constants.py
```diff
@@ -686,6 +686,28 @@ def fractional_octave_frequencies_exact(
         Electroacoustics - Octave-band and fractional-octave-band filters -
         Part 1: Specifications", IEC, 2014.
     """
+    if not isinstance(frequency_range, (tuple, np.ndarray, list)):
+        raise TypeError("The frequency range must be a tuple, list or" \
```
f-brinkmann: I think the escape is not needed, since everything is in paraenthesis already
```suggestion
        raise TypeError("The frequency range must be a tuple, list or"
```

h-chmeruk: Thanks for the review! I will remove the escape character :)

PR #877, opened 2026-01-09
[ENH] Feature/ticker module

sikersten:
### Which issue(s) are closed by this pull request?

Closes #724 

### Changes proposed in this pull request:

The aim of this PR is to add a ticker submodule to the plot module. This submodule already exists privately. To make it public, several changes are necessary, which are collected by this PR:

- [x] Initial PR: Add existing module to API documentation
- [ ] Align FractionalOctaveFormatter and FractionalOctaveTicker with corresponding functions in constants module.
- [ ] Rename ITAToolbox formatter and locator to use descriptive names.
- [ ] Extend general documentation on new module.
- [ ] Include usage examples and defaults. Could be inspired by matplotlibs[ formatter](https://matplotlib.org/stable/gallery/ticks/tick-formatters.html) and [locator](https://matplotlib.org/stable/gallery/ticks/tick-locators.html) notebooks.
- [ ] Add tests based on comparing ticks and ticklabels.

PR #878, opened 2026-01-09
[ENH] Ticker module: Update frequency formatter

sikersten:
Part of adding the ticker module in #877.

- Renamed LogFormatterITAToolbox to FrequencyFormatter
- Added example to docs
- Replaced LogFormatter with Formatter: This make the formatter more generalizable. For example, it can be used on linear axes (see example) and for formatting fixed standardized fractional octave frequencies.
- This replacement made the FractionalOctaveFormatter obsolete, so removed it.

### DIFF
--- a/pyfar/plot/_line.py
+++ b/pyfar/plot/_line.py
@@ -3,7 +3,7 @@ from pyfar import Signal, TimeData, FrequencyData
 import pyfar.dsp as dsp
 from . import _utils
 from .ticker import (
-    LogFormatterITAToolbox,
+    UnitLogFormatter,
     LogLocatorITAToolbox,
     MultipleFractionLocator,
     MultipleFractionFormatter)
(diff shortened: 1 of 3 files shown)


ahms5 (changes requested):
looks fine, thank you, I might suggest a more descriptive name, like FrequencyTickLabelFormatter, or Labeler or sth similar. ``FrequencyFormatter`` is very general.

ahms5:
``FrequencyFormatter`` tells what it is used for, but not was it does. This could be used for frequencies, but also other stuff.
e.g for air attenuation vs distance in meters. If we make it public people could also use it for other axes. I would probably use it for other plots as well.
I would call it therefore after if does, something like ``MetricPrefixFormatter``

sikersten:
Fully agree on the replacement with matplotlib's formatter, as it does essentially the same.
PR is replaced by #934

PR #890, opened 2026-01-23
[Bugfix] Wrap scipy rotation methods

### LINKED ISSUE #886: [Bug] Breaking changes in scipy's Rotation Class (v 1.17)
Changes to scipy's Rotation class broke our Orientations class. 
Probably an issue with the inheritance due to a big part of the scipy Rotation being moved to a cython backend.

### Note

Same problem in `spharpy.transforms.RotationSH`

hoyer-a:
### Note
Tests are failing because of #905 

Closes #886, #905 

Wraps scipy Rotation methods in `pf.Orientations`. I only wrapped the `from_`-methods and everything that is necessary to fix the tests. More changes need to be done, all functions that return an instance of the class in scipy Rotation need to be wrapped and tested. I would create a separate pull to keep it manageable review-wise.

Remove scipy < 1.17.0 restriction

### Test
Type checking tests are added for all wrapper-methods. The dunder-methods `__mul__`, `__getitem__`, `__setitem__` are used in other tests (and broke the tests before) and therefore not explicitly tested.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,7 @@ classifiers = [
 ]
 dependencies = [
     "numpy>=1.23.0",
-    "scipy>=1.5.0,<1.17.0",
+    "scipy>=1.5.0",
     "matplotlib",
     "sofar>=0.1.2",
     "urllib3",


pyfar/classes/orientations.py
```diff
@@ -235,15 +477,66 @@ def __setitem__(self, idx, val):
...
         """
-        if isinstance(val, Orientations):
-            val = val.as_quat()
-        quat = np.atleast_2d(val)
+        if isinstance(val, Rotation):
+            quat = val.as_quat()
+        elif isinstance(val, Orientations):
+            quat = val.as_quat()
+        else:
+            quat = np.atleast_2d(val)
```
f-brinkmann: Could probably be shorter
```suggestion
        if isinstance(val, (Rotation, Orientation)):
            quat = val.as_quat()
```

pyfar/classes/orientations.py
```diff
@@ -235,15 +477,66 @@ def __setitem__(self, idx, val):
...
+        Rotation object.
+
+        Parameters
+        ----------
+        other : Orientations or Rotation
+            The object to multiply with. If an Orientations object is provided,
+            it will be converted to a Rotation object for the multiplication.
+        """
+        if isinstance(other, Orientations):
+            other = Rotation.from_quat(other.as_quat())
```
f-brinkmann, line 885: Why is there not return for this case here and in `__rmul__`?

hoyer-a: It seemed easier and more robust to convert Orientations to a Rotation (first `if`) and then use scipy Rotation's `__mul__` (second `if`).

f-brinkmann: Ah my bed, misread it as an `elif`

h-chmeruk (approved):
It looks good to me too! Thank you for handling this!
Maybe we should add a note somewhere that the docstrings for the new functions were copied from the [SciPy documentation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html) and that further information with examples can be found there?

mberz:
I've moved this to "On Hold" to avoid accidental merges against main.
How/when do you intend to move this to the develop branch?

pyfar/classes/orientations.py
```diff
@@ -130,7 +389,7 @@ def from_view_up(cls, views, ups):
         rotation_matrix = np.asarray([views, ups, rights])
         rotation_matrix = np.swapaxes(rotation_matrix, 0, 1)
 
-        return super().from_matrix(rotation_matrix)
+        return cls.from_matrix(rotation_matrix)
```
mberz, line 392: phew, that's another slumbering bug that we never spotted right there :)

pyfar/classes/orientations.py
```diff
@@ -235,15 +494,72 @@ def __setitem__(self, idx, val):
...
+            quat = np.atleast_2d(val)
+
         if quat.ndim > 2 or quat.shape[-1] != 4:
             raise ValueError(f"Expected assigned value to have shape"
                              f" or (1, 4), got {quat.shape}")
+
         quats = self.as_quat()
         quats[idx] = quat
-        self = super().from_quat(quats)
+        self = Orientations(quats)
```
mberz:
```suggestion
        self = Orientations.from_quat(quats)
```
Since using the constructor is generally not advised, I'd keep using the from_quat method here.

pyfar/classes/orientations.py
```diff
@@ -235,15 +494,72 @@ def __setitem__(self, idx, val):
...
+        Parameters
+        ----------
+        idx : indexes
+            see NumPy Indexing
+        """
+        # Methods inherited from scipy.spatial.transforms.Rotation return an
+        # instance of the parent class for scipy >= 1.17 and are therefore
+        # wrapped to return an Orientations-object instead.
+        quat = self.as_quat()
+        return Orientations(quat[idx])
```
mberz:
```suggestion
        return Orientations.from_quat(quat[idx])
```
see comment above.

mberz (commented):
I have a couple smaller comments.

PR #893, opened 2026-01-30
[Bugfix] Wrap scipy rotation methods 2

hoyer-a:
Merges into #890 

Tests are failing because of #905 

Wrap remaining methods from scipy Rotation.

### Note
Concerning the docstrings, i tried to keep 'rotation' whenever it is used as a term to describe geometric terms and methods, e.g. 'rotation matrix', 'rotation between vectors' etc. and use orientation when referring to the object / class.

Feel free to make suggestions

### DIFF
--- a/pyfar/classes/orientations.py
+++ b/pyfar/classes/orientations.py
@@ -913,6 +913,9 @@ class Orientations(Rotation):
 
     def __pow__(self, n):
         """Compose orientation with itself n times."""
+        # Methods inherited from scipy.spatial.transforms.Rotation return an
+        # instance of the parent class for scipy >= 1.17 and are therefore
+        # wrapped to return an Orientations-object instead.
         rot = super().__pow__(n)
         return Orientations(rot.as_quat())
 


pyfar/classes/orientations.py
```diff
@@ -443,6 +443,136 @@ def show(self, positions=None,
...
+                "Pointing in Real Euclidean Space", Journal of Guidance,
+                Control, and Dynamics, Vol. 20, No. 5, 1997, pp. 916-922.
+        """
+        if return_sensitivity:
+            rot, rssd, sens = Rotation.align_vectors(a, b, weights,
+                                                     return_sensitivity)
+            return Orientations.from_quat(rot.as_quat()), rssd, sens
+
+        rot, rssd = Rotation.align_vectors(a, b, weights, return_sensitivity)
+        return Orientations.from_quat(rot.as_quat()), rssd
```
mberz, line 576:
```suggestion

        rot, rssd, sens = Rotation.align_vectors(
            a, b, weights, return_sensitivity)
        if return_sensitivity:
            return Orientations.from_quat(rot.as_quat()), rssd, sens
        else:
            return Orientations.from_quat(rot.as_quat()), rssd
```

mberz: You can make this slightly more compact

pyfar/classes/orientations.py
```diff
@@ -468,10 +598,205 @@ def as_view_up_right(self):
             return np.swapaxes(vector_triple, 0, 1)
         return vector_triple
 
+    @staticmethod
```
mberz: Could this be implemented as a classmethod to avoid having to overload in inheriting classes?

pyfar/classes/orientations.py
```diff
@@ -468,10 +598,205 @@ def as_view_up_right(self):
...
+                     in orientations if isinstance(rotation, Orientations)]
+
+        rot = Rotation.concatenate(orientations)
+        return Orientations(rot.as_quat())
+
     def copy(self):
         """Return a deep copy of the Orientations object."""
         return self.from_quat(self.as_quat())
 
+    @staticmethod
```
mberz: Could this be implemented as a class method to avoid having to overload in inheriting classes?

pyfar/classes/orientations.py
```diff
@@ -468,10 +598,205 @@ def as_view_up_right(self):
...
+        .. [#] Hartley, Richard, et al.,
+                "Rotation Averaging", International Journal of Computer Vision
+                103, 2013, pp. 267-305.
+        """
+        rot = Rotation.from_quat(self.as_quat())
+        rot = rot.mean(weights, axis)
+        return Orientations(rot.as_quat())
+
+
+    @staticmethod
```
mberz: Could this be implemented as a class method to avoid having to overload in inheriting classes?

mberz (commented):
I have a couple of minor comments and one conceptual question.

Do you see a way of implementing methods which return an Orientation object such that they do not require overloading in child classes? to me it seems that this is not possible, but would be pretty great to avoid having to do all of that in spharpy.

pyfar/classes/orientations.py
h-chmeruk: I don't think `@staticmethod` necessarily need to be overloaded, right? Only `@abstractmethod`.

pyfar/classes/orientations.py
hoyer-a: `sens` is only returned if return sensitivity is true, so this doesn't work unless we always call `Rotation.align_vectors` with `return_sensitivity=True`, which i wouldn't do.

Issue #813, opened 2025-07-15
Implicit use of FFT introduces noise


tjueterb:
Hey all,

When multiplying a `Signal` with a scalar, this is done in the frequency domain by default and introduces noise. This is especially noticeable when working with single point precision.

Proposed fix: Use multiplication in the time domain by default when multiplying a `Signal` with a scalar.

<img width="552" height="376" alt="Image" src="https://github.com/user-attachments/assets/ef789cba-6306-4f04-8f6b-20ef09e62b50" />

Best, Tobi

---

Code to reproduce (pyfar 0.7.3, numpy 1.24.3):

```python
import pyfar as pf
import numpy as np
import matplotlib.pyplot as plt

data_double = np.logspace(0, -20, 10000)
data_single = np.logspace(0, -20, 10000, dtype=np.float32)
signal_double = pf.Signal(data_double, 44100)
signal_single = pf.Signal(data_single, 44100)

fig, ax = plt.subplots(2,1, figsize=(6, 4))
pf.plot.time(signal_single, dB=True, label='single', ax=ax[0])
pf.plot.time(signal_double, dB=True, label='double', ax=ax[0])
pf.plot.time(pf.multiply((signal_single, 1), domain='time'), dB=True, ax=ax[0], label='single*1, time domain')
pf.plot.time(pf.multiply((signal_double, 1), domain='time'), dB=True, ax=ax[0], label='double*1, time domain')
pf.plot.time(pf.multiply((signal_single, 1), domain='freq'), dB=True, ax=ax[1], label='single*1, freq domain')
pf.plot.time(pf.multiply((signal_double, 1), domain='freq'), dB=True, ax=ax[1], label='double*1, freq domain')
for a in ax:
    a.set_ylim(-400, 0)
    a.legend()
```

jf-som:
I just came across this issue. I'm pretty sure this is the expected numerical error caused by FFTs. In your case, you have time-domain signals and implicitly FFT them into frequency domain when using pf.multiply in domain="freq" mode.
The same happens the other way around, if you have a Signal based on frequency data and FFT it into time domain:

```py
import pyfar as pf
import matplotlib.pyplot as plt
import numpy as np

time_data = np.logspace(0, -20, 10000, dtype=np.float32)
freq_data = np.fft.rfft(time_data)
# time_data = np.array(time_data, dtype=np.float32)
# freq_data = np.array(freq_data, dtype=np.float32)
[...]

PR #900, opened 2026-02-04
Normalize in the signal's domain

tjueterb:
This is related to #813 but was easier to fix.

Normalizing a signal in the wrong domain introduces noise, depending on the numerical precision.
I fixed `pf.dsp.normalize`, so it uses `signal.domain` instead of always defaulting do multiplication and division in the frequency domain.

This plot reproduces the issue. The second row shows `signal*1`, which still defaults to frequency domain multiplication.

precision_normalization

Code to reproduce:
```python
import pyfar as pf
import numpy as np

data_double = np.logspace(0, -20, 10000)
data_single = np.logspace(0, -20, 10000, dtype=np.float32)
signal_double = pf.Signal(data_double, 44100, domain='time')
signal_single = pf.Signal(data_single, 44100, domain='time')

fig, ax = plt.subplots(3,1, figsize=(5, 6))
pf.plot.time(signal_single, dB=True, label='single', ax=ax[0])
pf.plot.time(signal_double, dB=True, label='double', ax=ax[0])
pf.plot.time(signal_single * 1, dB=True, ax=ax[1], label='single*1, time domain')
pf.plot.time(signal_double * 1, dB=True, ax=ax[1], label='double*1, time domain')
pf.plot.time(pf.dsp.normalize(signal_double), dB=True, label='norm double', ax=ax[2])
pf.plot.time(pf.dsp.normalize(signal_single), dB=True, label='norm single', ax=ax[2])

ax[0].set_title('Original Signals')
ax[1].set_title('Before: Normalization always in frequency domain')
ax[2].set_title('After: Normalization in signal.domain')
ax[0].legend()

for a in ax:
    a.set_ylim(-400, 0)
plt.tight_layout()
```

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -8,7 +8,7 @@ History
 
 Fixed
 ^^^^^
-* Apply normalization in the signal's native domain to prevent floating point errors (PR #900)
+* Make `pyfar.dsp.normalize` use the signal's native domain to avoid floating point errors (PR #900)
 
 0.7.4 (2026-02-02)
 ------------------
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -8,7 +8,7 @@ History
 
 Fixed
 ^^^^^
-* Make `pyfar.dsp.normalize` use the signal's native domain to avoid floating point errors (PR #900)
+* Make `pyfar.dsp.normalize` use the signal's current domain to avoid floating point errors (PR #900)
 
 0.7.4 (2026-02-02)
 ------------------
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,14 @@
 History
 =======
 
+
+[Unreleased]
+------------
+
+Fixed
+^^^^^
+* Apply normalization in the signal's native domain to prevent floating point errors (PR #900)
+
 0.7.4 (2026-02-02)
 ------------------
 
--- a/pyfar/dsp/dsp.py
+++ b/pyfar/dsp/dsp.py
@@ -2592,6 +2592,9 @@ def normalize(signal, reference_method='max', domain='auto',
         raise TypeError(("Input data has to be of type 'Signal', 'TimeData' "
                          "or 'FrequencyData'."))
 
+    # Store signal domain for later.
+    signal_domain = signal.domain
+
     if domain not in ('time', 'freq', 'auto'):
         raise ValueError("domain must be 'time', 'freq' or 'auto' but is"
                          f" '{domain}'.")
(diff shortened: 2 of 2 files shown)


pyfar/dsp/dsp.py
```diff
@@ -2688,7 +2688,9 @@ def normalize(signal, reference_method='max', domain='auto',
         target = np.sqrt(target)
 
     # apply normalization
-    normalized_signal = signal.copy() * target / reference_norm
+    normalized_signal = pyfar.divide(
```
f-brinkmann, line 2696: can you maybe extend the comment above to explain this? Would make sure we don't change it back in 5 years after we forgot why we did this. Maybe

```
# Apply normalization in the current domain of the signal. This avoids unnecessary conversions between the time and frequency domain which introduce numerical errors.
```

f-brinkmann (commented):
Thanks for looking into this. I suggested to explicitly mention why this is done in the code.

pyfar/dsp/dsp.py
tjueterb, line 2633: This forces `signal.domain` to `domain` (the domain used for calculating `reference_norm`. 

Inputting a frequency signal like `pf.signals.linear_perfect_sweep` and normalizing in the time domain changes the signal domain to time.

This can't be intended, right?

tjueterb: Avoided this by restoring the original signal domain

f-brinkmann (approved):

HISTORY.rst
```diff
@@ -2,6 +2,14 @@
 History
 =======
 
+
+[Unreleased]
+------------
+
+Fixed
+^^^^^
+* Apply normalization in the signal's native domain to prevent floating point errors (PR #900)
```
hoyer-a: Maybe would be good to reference `pyfar.dsp.normalize` here

hoyer-a (approved):
Thanks for fixing

HISTORY.rst
tjueterb: Done

HISTORY.rst
```diff
@@ -2,6 +2,14 @@
 History
 =======
 
+
+[Unreleased]
+------------
+
+Fixed
+^^^^^
+* Make `pyfar.dsp.normalize` use the signal's native domain to avoid floating point errors (PR #900)
```
mberz:
```suggestion
* Make `pyfar.dsp.normalize` use the signal's current domain to avoid floating point errors (PR #900)
```
Native may mean something else :)

mberz (approved):
Looks good.
I'm going to apply the change in wording right away and merge.

PR #837, opened 2025-09-01
_PyfarMultichannel class

h-chmeruk:
As proposed in the [Issue #835](https://github.com/pyfar/pyfar/issues/835), here is a draft implementation of the _NumPyBase class. This abstract class should define the NumPy-like functions that must then be implemented by the _Audio, Coordinates and other classes.

Edit: From the fourth commit , the class _NumPyBase is now called _PyfarMultichannel.

PR #904, opened 2026-02-12
Refactor pf.Orientations class to pf.Rotation

### LINKED ISSUE #600: Move show functions to plot module

### LINKED ISSUE #639: revise Orientations class
[pyfar.Orientations](https://pyfar.readthedocs.io/en/stable/classes/pyfar.orientations.html) is derived form [scipy.spatial.transform.Rotation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html) 
our aim was to also add the view/up notation to the class (and plots, e.g. ``show``)
there are a few points which we can discuss to enhance the class behaviour 
- execute examples in the doc, e.g. show plots in Sphinx 
- we implemented ``from_view_up`` as construcor and ``as_view_up_right`` to get the view up and right, we should stick to one convention (do we need right?)
- we allow empty Orientations, where scipy does not. so calling pf.Oreintations() will not create an empty orientation, it will use ``quat=np.array([0., 0., 0., 1.])``, this might lead to confusing behaviour, since the user could expect an empty class.
    - for this reason we overwrote the init function, which we should not, if we really need. I can cause problems with new scipy versions.
- our class forces ``normalize=True``, where scipy does not, is there a reason for that?
- ``__setitem__`` is not working 

```
from pyfar import Orientations
views = [[1, 0, 0], [2, 0, 0]]
ups = [[0, 1, 0], [0, -2, 0]]
orientations = Orientations.from_view_up(views, ups)
positions = [[0, 0.5, 0], [0, -0.5, 0]]
orientations.show(positions)
from scipy.spatial.transform import Rotation
rot_x45 = Rotation.from_euler('x', 45, degrees=True)
orientations[1] = orientations[1] * rot_x45
orientations.show(positions)
```
grafik
grafik

one of the arrows in the second plot should be rotated.

hoyer-a:
### Note

I cherry picked #848 to fix the tests

---

Pyfar `Rotation` class to replace `Orientations` taking the inheritance problem from #886 into account and including #639 and #600 (remove show())

Idea:
- No inheritance from scipy
- Instantiate objects only via `from_...` classmethods wrapping scipy `Rotation.from_...`
- Store scipy Rotation in class

Changes to previous `Orientations`:
- remove `as_view_up_right` in favor of `as_view_up`
- add `cshape, csize` property
- Objects can only be instantiated via `from_...` classmethod. `Rotation()` throws an error
- `setitem` is disabled and throws an error
- Remove possibility of using degrees to be consistent with `Coordinates` 
- [x] remove `show() `method
- [x] fix footnotes

### DIFF
--- a/pyfar/classes/rotation.py
+++ b/pyfar/classes/rotation.py
@@ -29,17 +29,54 @@ class Rotation():
         num_orientations = \
             1 if quat.ndim == 1 else quat.shape[0]
 
-        _repr = f"Pyfar Rotations object with {num_orientations} rotations."
-        return _repr
+        repr = f"Pyfar Rotations object with {num_orientations} rotations."
+        return repr
 
-    # Factory Methods
+    def __getitem__(self, idx):
+        self._rot = self._rot[idx]
+        return self
+
+    def __iter__(self):
+        raise NotImplementedError
+
+    def __setitem__(self):
+        raise NotImplementedError('Setting an item is disabled for pyfar'
+        'Rotations. If you want to modify the Rotation, use an array'
+        'representation like `as_quat()` or `as_matrix()` and create a new '
+        'object.')
+
+    def __mul__(self, other):
+        """"""
+        if isinstance(other, Rotation):
+            other = scRotation.from_quat(other.as_quat())
+        if isinstance(other, scRotation):
+            self._rot *= other
+            return self
+
+    def __rmul__(self, other):
+        """"""
+        if isinstance(other, Rotation):
+            other = scRotation.from_quat(other.as_quat())
+        if isinstance(other, scRotation):
+            self._rot *= other
+            return self
+
+    def __pow__(self, n):
+        """Compose orientation with itself n times."""
+        self._rot = self._rot.__pow__(n)
+        return self
+
+    def __eq__(self, other):
+        """Check for equality of two objects."""
+        return np.array_equal(self.as_quat(), other.as_quat())
+
+    # from-... methods
     @classmethod
     def from_davenport(cls, axes, order, angles, degrees=False):
         """"""
         instance = cls.__new__(cls)
         instance._rot = scRotation.from_davenport(
             axes, order, angles, degrees=degrees)
-
         return instance
 
     @classmethod
@@ -48,7 +85,6 @@ class Rotation():
         instance = cls.__new__(cls)
         instance._rot = scRotation.from_euler(
             seq, angles, degrees=degrees)
-
         return instance
 
     @classmethod
@@ -56,7 +92,12 @@ class Rotation():
         """"""
         instance = cls.__new__(cls)
         instance._rot = scRotation.from_matrix(matrix)
+        return instance
 
+    @classmethod
+    def from_mrp(cls, mrp):
+        instance = cls.__new__(cls)
+        instance._rot = scRotation.from_mrp(mrp)
         return instance
 
     @classmethod
@@ -65,6 +106,107 @@ class Rotation():
         instance = cls.__new__(cls)
         instance._rot = scRotation.from_quat(quat)
         return instance
+
+    @classmethod
+    def from_rotvec(cls, rotvec, degrees=False):
+        """"""
+        instance = cls.__new__(cls)
+        instance._rot = scRotation.from_rotvec(rotvec, degrees)
+        return instance
+
+    @classmethod
+    def from_view_up(cls, views, ups):
+        """Initialize Orientations from a view an up vector.
+
+        Orientations are internally stored as quaternions for better spherical
+        linear interpolation (SLERP) and spherical harmonics operations.
+        More intuitionally, they can be expressed as view and up vectors
+        which cannot be collinear. In this case, they are restricted to be
+        perpendicular to minimize rounding errors.
+
+        Parameters
+        ----------
+        views : array_like, shape (N, 3) or (3,), Coordinates
+            A single vector or a stack of vectors, giving the look-direction of
+            an object in three-dimensional space, e.g. from a listener, or the
+            acoustic axis of a loudspeaker, or the direction of a main lobe.
+            Views can also be passed as a Coordinates object.
+
+        ups : array_like, shape (N, 3) or (3,), Coordinates
+            A single vector or a stack of vectors, giving the up-direction of
+            an object, which is usually the up-direction in world-space. Views
+            can also be passed as a Coordinates object.
+
+        Returns
+        -------
+        orientations : Orientations
+            Object containing the orientations represented by quaternions.
+        """
+        instance = cls.__new__(cls)
+
+        # init views and up
+        try:
+            views = np.atleast_2d(views).astype(np.float64)
+            ups = np.atleast_2d(ups).astype(np.float64)
+        except VisibleDeprecationWarning as exc:
+            raise ValueError(
+                "Expected `views` and `ups` to have shape (N, 3)") from exc
+
+        # check views and ups
+        if (views.ndim > 2 or views.shape[-1] != 3 or
+                ups.ndim > 2 or ups.shape[-1] != 3):
+            raise ValueError(f"Expected `views` and `ups` to have shape (N, 3)"
+                             f" or (3,), got {views.shape}")
+        if views.shape == ups.shape:
+            pass
+        elif views.shape[0] > 1 and ups.shape[0] == 1:
+            ups = np.repeat(ups, views.shape[0], axis=0)
+        elif ups.shape[0] > 1 and views.shape[0] == 1:
+            views = np.repeat(views, ups.shape[0], axis=0)
+        else:
+            raise ValueError("Expected 1:1, 1:N or N:1 `views` and `ups` "
+                             f"not M:N, got {views.shape} and {ups.shape}")
+
+        if not (np.all(np.linalg.norm(views, axis=1)) and
+                np.all(np.linalg.norm(ups, axis=1))):
+            raise ValueError("View and Up Vectors must have a length.")
+        if not np.allclose(0, np.einsum('ij,kj->k', views, ups)):
+            raise ValueError("View and Up vectors must be perpendicular.")
+
+        # Assuming that the direction of the cross product is defined
+        # by the right-hand rule
+        rights = np.cross(views, ups)
+
+        # In a standard Cartesian right-handed coordinate system,
+        # these vectors are defined as [x, y, z] = [view, left, up], where
+        # left is the same vector as -rights
+        rotation_matrix = np.asarray([views, -rights, ups])
+        rotation_matrix = np.swapaxes(rotation_matrix, 0, 1)
+
+        return instance.from_matrix(rotation_matrix)
+
+    # other class methods
+    @classmethod
+    def align_vectors(a, b, weights=None, return_sensitivity=False):
+        raise NotImplementedError
+
+    @classmethod
+    def concatenate(cls, rotation):
+        raise NotImplementedError
+
+    @classmethod
+    def identity(cls, num=None, *, shape=None):
+        raise NotImplementedError
+
+    @classmethod
+    def random(cls, num=None, rng=None, *, shape=None):
+        raise NotImplementedError
+
+    # private class methods
+    @classmethod
+    def _decode(cls, obj_dict):
+        """Decode object based on its respective `_encode` counterpart."""
+        return cls.from_quat(obj_dict['quat'])
 
     @classmethod
     def _from_scipy_rotation(cls, sc_rotation):
@@ -90,6 +232,26 @@ class Rotation():
         """"""
         return self._rot.as_matrix()
 
+    def as_view_up_right(self):
+        raise NotImplementedError
+
+    def copy(self):
+        """Return a deep copy of the Orientations object."""
+        return self.from_quat(self.as_quat())
+
+    def inv(self):
+        raise NotImplementedError
+
     def mean(self, weights=None, axis=None):
         """"""
         return self._from_scipy_rotation(self._rot.mean(weights, axis))
+
+    def reduce(self, left=None, right=None, return_indices=False):
+        raise NotImplementedError
+
+    def show(self, positions=None,
+             show_views=True, show_ups=True, show_rights=True, **kwargs):
+        raise NotImplementedError
+
+    # private instance methods
+
--- a/pyfar/classes/rotation.py
+++ b/pyfar/classes/rotation.py
@@ -95,7 +95,15 @@ class Rotation():
         'object.')
 
     def __mul__(self, other):
-        """"""
+        """
+        Multiply Rotation object with another Rotation or a
+        scipy.spatial.transform.Rotation.
+
+        Parameters
+        ----------
+        other : Rotation or scipy.spatial.transform.Rotation
+            The object to multiply with.
+        """
         if isinstance(other, Rotation):
             other = scRotation.from_quat(other.as_quat())
         if isinstance(other, scRotation):
@@ -104,7 +112,7 @@ class Rotation():
 
     def __rmul__(self, other):
         """
-        Multiply Rotation object with another Rotation or a
+        Right multiplication of Rotation object with another Rotation or a
         scipy.spatial.transform.Rotation.
 
         Parameters
@@ -579,8 +587,8 @@ class Rotation():
                 optimal matrix algorithm", Journal of Astronautical Sciences,
                 Vol. 41, No.2, 1993, pp. 261-280.
         .. [#] Bar-Itzhack, Itzhack Y., Daniel Hershkowitz, and Leiba Rodman,
-                "Pointing in Real Euclidean Space", Journal of Guidance,
-                Control, and Dynamics, Vol. 20, No. 5, 1997, pp. 916-922.
+               "Pointing in Real Euclidean Space", Journal of Guidance,
+               Control, and Dynamics, Vol. 20, No. 5, 1997, pp. 916-922.
         """
         result = scRotation.align_vectors(a, b, weights, return_sensitivity)
 
@@ -598,7 +606,7 @@ class Rotation():
 
         Parameters
         ----------
-        orientations : sequence of Orientations objects
+        rotations : sequence of Orientations objects
             The orientations to concatenate. If a single Orientations object is
             passed in, a copy is returned.
 
@@ -731,10 +739,10 @@ class Rotation():
         Parameters
         ----------
         axes : array_like, shape (..., [1 or 2 or 3], 3) or (..., 3)
-            Axis of rotation, if one dimensional. If N dimensional, describes the
-            sequence of axes for rotations, where each axes[..., i, :] is the ith
-            axis. If more than one axis is given, then the second axis must be
-            orthogonal to both the first and third axes.
+            Axis of rotation, if one dimensional. If N dimensional, describes
+            the sequence of axes for rotations, where each axes[..., i, :] is
+            the ith axis. If more than one axis is given, then the second axis
+            must be orthogonal to both the first and third axes.
         order : string
             If it belongs to the set {'e', 'extrinsic'}, the sequence will be
             extrinsic. If it belongs to the set {'i', 'intrinsic'}, sequence
@@ -760,11 +768,11 @@ class Rotation():
         References
         ----------
         .. [#] Shuster, Malcolm & Markley, Landis. (2003). Generalization of
-               the Euler Angles. Journal of the Astronautical Sciences. 51. 123-132.
-               10.1007/BF03546304.
+               the Euler Angles. Journal of the Astronautical Sciences. 51.
+               123-132. 10.1007/BF03546304.
         .. [#] Bernardes E, Viollet S (2022) Quaternion to Euler angles
-               conversion: A direct, general and computationally efficient method.
-               PLoS ONE 17(11): e0276302. 10.1371/journal.pone.0276302
+               conversion: A direct, general and computationally efficient
+               method. PLoS ONE 17(11): e0276302. 10.1371/journal.pone.0276302
         .. [#] https://en.wikipedia.org/wiki/Gimbal_lock#In_applied_mathematics
         """
 
@@ -828,7 +836,8 @@ class Rotation():
         .. [#] https://en.wikipedia.org/wiki/Gimbal_lock#In_applied_mathematics
         """
 
-        return self._rot.as_euler(seq, degrees, suppress_warnings)
+        return self._rot.as_euler(seq, degrees,
+                                  suppress_warnings=suppress_warnings)
 
     def as_matrix(self):
         """
@@ -932,7 +941,7 @@ class Rotation():
         ----------
         .. [#] https://en.wikipedia.org/wiki/Quaternions_and_spatial_rotation
         """
-        return self._rot.as_quat(canonical, scalar_first)
+        return self._rot.as_quat(canonical, scalar_first=scalar_first)
 
     def as_rotvec(self, degrees=False):
         """
--- a/pyfar/classes/rotation.py
+++ b/pyfar/classes/rotation.py
@@ -19,13 +19,13 @@ class Rotation():
     Rotation in the three-dimensional space.
 
     This class is largely based on
-    :py:class:`scipy:scipy.spatial.transform.Rotation` and wraps all
+    :py:class:`scipy.spatial.transform.Rotation` and wraps all
     functionality that scipy's Rotation class provides.
     In addition the pyfar Rotation class adds the creation from perpendicular
     view and up vectors through :py:func:`~from_view_up`, and the
     representation as view / up in :py:func:`~as_view_up`.
 
-    A rotation can be visualized with the triple of view, up and right
+    A rotation can be represented with the triple of view, up and right
     vectors and it is tied to the object's local coordinate system.
     Alternatively the object's rotation can be illustrated with help of the
     right hand: Thumb (view), forefinger (up) and middle finger (right).
@@ -90,11 +90,11 @@ class Rotation():
     def __mul__(self, other):
         """
         Multiply Rotation object with another Rotation or
-        :py:class:`scipy:scipy.spatial.transform.Rotation`.
+        :py:class:`scipy.spatial.transform.Rotation`.
 
         Parameters
         ----------
-        other : Rotation or :py:class:`scipy:scipy.spatial.transform.Rotation`
+        other : Rotation or :py:class:`scipy.spatial.transform.Rotation`
             The object to multiply with.
         """
         if isinstance(other, Rotation):
@@ -105,11 +105,11 @@ class Rotation():
     def __rmul__(self, other):
         """
         Right multiplication of Rotation object with another Rotation or a
-        :py:class:`scipy:scipy.spatial.transform.Rotation`.
+        :py:class:`scipy.spatial.transform.Rotation`.
 
         Parameters
         ----------
-        other : Rotation or :py:class:`scipy:scipy.spatial.transform.Rotation`.
+        other : Rotation or :py:class:`scipy.spatial.transform.Rotation`.
             The object to multiply with.
         """
         if isinstance(other, Rotation):
@@ -148,7 +148,7 @@ class Rotation():
         """
         Initialize from Davenport angles.
 
-        Wraps :py:meth:`scipy:scipy.spatial.transform.Rotation.from_davenport`.
+        Wraps :py:meth:`scipy.spatial.transform.Rotation.from_davenport`.
 
         Rotations in 3-D can be represented by a sequence of 3
         rotations around a sequence of axes.
@@ -209,7 +209,7 @@ class Rotation():
         """
         Initialize from Euler angles.
 
-        Wraps :py:meth:`scipy:scipy.spatial.transform.Rotation.from_euler`.
+        Wraps :py:meth:`scipy.spatial.transform.Rotation.from_euler`.
 
         Rotations in 3-D can be represented by a sequence of 3
         rotations around a sequence of axes. In theory, any three axes spanning
@@ -251,7 +251,7 @@ class Rotation():
         """
         Initialize from rotation matrix.
 
-        Wraps :py:meth:`scipy:scipy.spatial.transform.Rotation.from_matrix`.
+        Wraps :py:meth:`scipy.spatial.transform.Rotation.from_matrix`.
 
         Rotations in 3 dimensions can be represented with 3 x 3 orthogonal
         matrices [#]_. If the input is not orthogonal, an approximation is
@@ -294,7 +294,7 @@ class Rotation():
         """
         Initialize from Modified Rodrigues Parameters (MRPs).
 
-        Wraps :py:meth:`scipy:scipy.spatial.transform.Rotation.from_mrp`.
+        Wraps :py:meth:`scipy.spatial.transform.Rotation.from_mrp`.
 
         MRPs are a 3 dimensional vector co-directional to the axis of rotation
         and whose magnitude is equal to ``tan(theta / 4)``, where ``theta`` is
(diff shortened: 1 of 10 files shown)


pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1187 @@
...
+        except VisibleDeprecationWarning as exc:
+            raise ValueError(
+                "Expected `views` and `ups` to have shape (N, 3)") from exc
+
+        # check views and ups
+        if (views.ndim > 2 or views.shape[-1] != 3 or
+                ups.ndim > 2 or ups.shape[-1] != 3):
+            raise ValueError(f"Expected `views` and `ups` to have shape (N, 3)"
+                             f" or (3,), got {views.shape}")
+        if views.shape == ups.shape:
```
ahms5: How about using [np.broadcast_shapes](https://numpy.org/doc/stable/reference/generated/numpy.broadcast_shapes.html) to get the broadcasted shape, check the shape is correct and then broadcast both arrays via [numpy.broadcast_to](https://numpy.org/doc/stable/reference/generated/numpy.broadcast_to.html#numpy.broadcast_to). 
See coordinated class as example

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1187 @@
...
+else:
+    from numpy.exceptions import VisibleDeprecationWarning
+
+# this warning needs to be caught and appears if numpy array are generated
+# from nested lists containing lists of unequal lengths,
+#  e.g., [[1, 0, 0], [1, 0]]
+warnings.filterwarnings("error", category=VisibleDeprecationWarning)
+
+
+class Rotation():
```
ahms5, line 17:
```suggestion
class Rotations():
```

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1187 @@
...
+    def __setitem__(self, *args):
+        """
+        Assign rotation(s) at given index(es) from object.
+
+        This is disabled for `Rotation`.
+        """
+        raise NotImplementedError('Setting an item is disabled for pyfar '
+        'Rotations. If you want to modify the Rotation, use an array '
+        'representation like `as_quat()` or `as_matrix()` and create a new '
+        'object.')
```
ahms5:
```suggestion
        raise NotImplementedError(
            'Setting an item is disabled for pyfar '
            'Rotations. If you want to modify the Rotation, use an array '
            'representation like `as_quat()` or `as_matrix()` and create a new '
            'object.')
```

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1187 @@
...
+        rot = scRotation.from_euler(
+            seq, angles, degrees=degrees)
+        return cls._from_scipy_rotation(rot)
+
+    @classmethod
+    def from_matrix(cls, matrix):
+        """
+        Initialize from rotation matrix.
+
+        Wraps :py:meth:`scipy:scipy.spatial.transform.Rotation.from_matrix`.
```
ahms5:
```suggestion
        Wraps :py:math:`scipy:scipy.spatial.transform.Rotation.from_matrix`.
```

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1187 @@
...
+        Rotations in 3 dimensions can be represented with 3 x 3 orthogonal
+        matrices [#]_. If the input is not orthogonal, an approximation is
+        created by orthogonalizing the input matrix using the method described
+        in [#]_, and then converting the orthogonal rotation matrices to
+        quaternions using the algorithm described in [#]_. Matrices must be
+        right-handed.
+
+        Parameters
+        ----------
+        matrix : array_like, shape (..., 3, 3)
```
ahms5, line 262: apparently scipy allows mutlidimensional rotations, why we dont allow it?

e.g. n_rotations, instead of cshape/csize
than we could lateron inherit this class from [pyfarMultiChannel](https://github.com/pyfar/pyfar/pull/837/changes).

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1187 @@
...
+
+        Rotations are internally stored as quaternions for better spherical
+        linear interpolation (SLERP) and spherical harmonics operations.
+        More intuitionally, they can be expressed as view and up vectors
+        which cannot be collinear. In this case, they are restricted to be
+        perpendicular to minimize rounding errors.
+
+        Parameters
+        ----------
+        views : array_like, shape (N, 3) or (3,), Coordinates
```
ahms5:
```suggestion
        views : array_like, shape (..., 3) or (3,), Coordinates
```

I would like to also allow multidimensional Ritations, similar to what we do with Coordinates.

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1187 @@
...
+
+        # Assuming that the direction of the cross product is defined
+        # by the right-hand rule
+        rights = np.cross(views, ups)
+
+        # In a standard Cartesian right-handed coordinate system,
+        # these vectors are defined as [x, y, z] = [view, left, up], where
+        # left is the same vector as -rights
+        rotation_matrix = np.asarray([views, -rights, ups])
+        rotation_matrix = np.swapaxes(rotation_matrix, 0, 1)
```
ahms5:
```suggestion
        rotation_matrix = np.stack([views, -rights, ups], axis=-1)
```
shouldnt this work?

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1187 @@
...
+        if np.asarray(rotations).shape[0] == 1:
+            return rotations[0].copy()
+
+        rotations = [rotation._rot for rotation in rotations]
+
+        rot = scRotation.concatenate(rotations)
+        return cls._from_scipy_rotation(rot)
+
+    @classmethod
+    def identity(cls, num=None, *, shape=None):
```
ahms5: what is the meaning of the ```*```

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1187 @@
...
+        """
+        Get identity rotation(s).
+
+        Wraps :py:meth:`scipy:scipy.spatial.transform.Rotation.identity`.
```
ahms5:
```suggestion
        Wraps :py:math:`scipy:scipy.spatial.transform.Rotation.identity`.
```

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1187 @@
...
+        """
+        rot  = scRotation.identity(num, shape=shape)
+        return cls._from_scipy_rotation(rot)
+
+    @classmethod
+    def random(cls, num=None, rng=None, *, shape=None):
+        """
+        Generate rotations that are uniformly distributed on a sphere.
+
+        Wraps :py:meth:`scipy:scipy.spatial.transform.Rotation.random`.
```
ahms5:
```suggestion
        Wraps :py:math:`scipy:scipy.spatial.transform.Rotation.random`.
```

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1187 @@
...
+
+        Rotation are internally stored as quaternions for better spherical
+        linear interpolation (SLERP) and spherical harmonics operations.
+        More intuitionally, they can be expressed as view and and up of vectors
+        which cannot be collinear. In this case are restricted to be
+        perpendicular to minimize rounding errors.
+
+        Returns
+        -------
+        vector_triple: ndarray, shape (N, 3), normalized vectors
```
ahms5:
```suggestion
        vector_triple: ndarray, shape (..., 3), normalized vectors
```

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1187 @@
...
+        """
+        if return_indices:
+            rot, left_idx, right_idx = \
+                self._rot.reduce(left, right, return_indices)
+            return self._from_scipy_rotation(rot), left_idx, right_idx
+
+        rot = self._rot.reduce(left, right, return_indices)
+        return self._from_scipy_rotation(rot)
+
+    def show(self, positions=None,
```
ahms5: remove

tests/test_rotation.py
```diff
@@ -0,0 +1,306 @@
...
+    match = \
+        "Rotation objects must be created using one of the `from_...` methods"
+    with pytest.raises(RuntimeError, match=match):
+        Rotation()
+
+def test_class_properties(rotation):
+    """Test Rotation.n_rotations parameter."""
+    assert rotation.n_rotations == 3
+
+def test_rotation_from_view_up():
```
ahms5: this tests sounds like parametrization.

tests/test_rotation.py
```diff
@@ -0,0 +1,306 @@
...
+    """
+    views = np.atleast_2d(views).astype(np.float64)
+    views /= np.linalg.norm(views, axis=1)[:, np.newaxis]
+    ups = np.atleast_2d(ups).astype(np.float64)
+    ups /= np.linalg.norm(ups, axis=1)[:, np.newaxis]
+
+    views_, ups_ = rotation.as_view_up()
+
+    np.testing.assert_allclose(views_, views, atol=1e-15)
+    np.testing.assert_allclose(ups_, ups, atol=1e-15)
```
ahms5, line 128: can we maybe simplify this test, so that it become easier to understand?
I would suggest to define view and ups for this case without conftests

tests/test_rotation.py
```diff
@@ -0,0 +1,306 @@
...
+        [0, 1, 0],
+        [0, -1, 0],
+        [0, 0, 1],
+        [0, 0, -1]])
+
+    for v1 in range(len(vec)):
+        for v2 in range(len(vec)):
+            if np.all(np.abs(vec[v1]) == np.abs(vec[v2])):
+                continue
+            print(f"Testing combination ({v1}, {v2})")
```
ahms5:
```suggestion
@pytest.parametrize('v1', [
      [1, 0, 0],
        [-1, 0, 0],
        [0, 1, 0],
        [0, -1, 0],
        [0, 0, 1],
        [0, 0, -1]]
])
@pytest.parametrize('v2', [
      [1, 0, 0],
        [-1, 0, 0],
        [0, 1, 0],
        [0, -1, 0],
        [0, 0, 1],
        [0, 0, -1]]
])
def test_from_view_as_view_roundtrip(v1, v2):
```

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1187 @@
...
+        Parameters
+        ----------
+        other : Rotation or :py:class:`scipy:scipy.spatial.transform.Rotation`
+            The object to multiply with.
+        """
+        if isinstance(other, Rotation):
+            other = scRotation.from_quat(other.as_quat())
+        if isinstance(other, scRotation):
+            self._rot *= other
+            return self
```
ahms5:
```suggestion
        if isinstance(other, Rotation):
            other = scRotation.from_quat(other.as_quat())
        self._rot *= other
        return self
```
how about this? let scipy handle errors :D
if other is not an object, nothing would happen in the current implentation

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1187 @@
...
+    Rotate by 45 degree in x-direction:
+
+    >>> rot_x45 = Rotation.from_euler('x', 45, degrees=True)
+    >>> rotations = rotations * rot_x45
+    >>> rotations.show(positions)
+
+    To create `Rotation` objects use ``from_...`` methods.
+    """
+
+    def __init__(self):
```
ahms5, line 49: we could also add an as_scipy_rotation() method. This would be cinstistance with the from_scipy_rotation()

ahms5 (changes requested):
Thank you. I would vote for mutliChannel option, as we also support it for out data object and coodinates.
 I think there are some old tests, I would take the change to refactor the test to make them allign with our guidelines.

pyfar/classes/rotation.py
hoyer-a: good point, i will change from_view_up to multidim and add cshape / csize as properties

pyfar/classes/rotation.py
hoyer-a: All parameters after `*` can only be passed as keywords. It is used in the scipy method, we don't have to keep it.
But it seems like a useful mechainsm to ensure correct use here.

pyfar/classes/rotation.py
hoyer-a: Was this agreed on? I actually don't know and just followed https://github.com/pyfar/pyfar/issues/886#issuecomment-3777920749.
Would most certainly help to avoid confusion of scipy.Rotation and pyfar.Rotation.

pyfar/classes/rotation.py
hoyer-a: Do we want to 'expose' the scipy Rotations in our class or keep them under the hood? 
If we add `as_scipy_rotation()` i would also make `_from_scipy_rotation()` public. Right now it's just there to save some lines of code internally.

pyfar/classes/rotation.py
hoyer-a: Ah yes, good catch.

pyfar/classes/rotation.py
f-brinkmann: The class in spharpy is called `RotationSH`, so `Rotation` might be more consistent here.

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1187 @@
...
+
+# this warning needs to be caught and appears if numpy array are generated
+# from nested lists containing lists of unequal lengths,
+#  e.g., [[1, 0, 0], [1, 0]]
+warnings.filterwarnings("error", category=VisibleDeprecationWarning)
+
+
+class Rotation():
+    """
+    This class for Rotation in the three-dimensional space,
```
f-brinkmann: Should we start with a one-line, e.g., 'Rotation in three dimensions.'?

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1187 @@
...
+    is largely based on :py:class:`scipy:scipy.spatial.transform.Rotation` and
+    wraps all functionality that
+    scipy's Rotation class provides.
+    In addition the pyfar Rotation class adds the creation from perpendicular
+    view and up vectors through :py:func:`~from_view_up`, the representation
+    as view / up in :py:func:`~as_view_up` and a convenient plot
+    function :py:func:`~show`.
```
f-brinkmann: Remove the mention of the plot?
```suggestion
    as view / up in :py:func:`~as_view_up`.
```

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1187 @@
...
+    """
+
+    def __init__(self):
+        raise RuntimeError("Rotation objects must be created using one of "
+                           "the `from_...` methods")
+
+    def __repr__(self):
+        """String representation of Rotation object."""
+        repr_string = \
+              f"Pyfar Rotations object with {self.n_rotations} rotations."
```
f-brinkmann: Maybe even shorter and if becoming multi-dimensional as suggeted by ahms5, it should probably use cshape
```suggestion
              f"pyfar.Rotations with {self.cshape} rotations."
```
Other opionons on this?

pyfar/classes/rotation.py
f-brinkmann: Marked as unresolved - maybe local changes not commited yet?

pyfar/classes/rotation.py
f-brinkmann: No string opinon here - would be fine with both

f-brinkmann (changes requested):

tests/test_rotation.py
hoyer-a: I added some comments. Actually i think we can remove this test, it tests if the returned vectors are normalized, which since we use scipy's Rotation is always the case. And i don't think we need to test if scipy normalizes correctly.
The `from_view_up` / `as_view_up` behaviour is tested in `test_from_view_as_view_roundtrip`.

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1133 @@
...
+    """
+
+    def __init__(self):
+        raise RuntimeError("Rotation objects must be created using one of "
+                           "the `from_...` methods")
+
+    def __repr__(self):
+        """String representation of Rotation object."""
+        repr_string = \
+              f"Pyfar.Rotation with {self.cshape} rotations."
```
f-brinkmann:
```suggestion
              f"pyfar.Rotation with {self.cshape} rotations."
```

f-brinkmann (changes requested):
Only had minor comments left. For some reason the tests are failing locally for me, I will check later.

pyfar/classes/rotation.py
hoyer-a: `py:math:` is not the right sphinx directive for methods. Forgot to comment

pyfar/classes/rotation.py
hoyer-a: see above

pyfar/classes/rotation.py
hoyer-a: see above

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1133 @@
...
+        return np.prod(self.cshape)
+
+
+    # from-... methods
+    @classmethod
+    def from_davenport(cls, axes, order, angles, degrees=False):
+        """
+        Initialize from Davenport angles.
+
+        Wraps :py:meth:`scipy:scipy.spatial.transform.Rotation.from_davenport`.
```
ahms5:
```suggestion
        Wraps :py:func:`scipy:scipy.spatial.transform.Rotation.from_davenport`.
```

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1133 @@
...
+        rot = scRotation.from_davenport(
+            axes, order, angles, degrees=degrees)
+        return cls._from_scipy_rotation(rot)
+
+    @classmethod
+    def from_euler(cls, seq, angles, degrees=False):
+        """
+        Initialize from Euler angles.
+
+        Wraps :py:meth:`scipy:scipy.spatial.transform.Rotation.from_euler`.
```
ahms5:
```suggestion
        Wraps :py:meth:`scipy.spatial.transform.Rotation.from_euler`.
```

pyfar/classes/rotation.py
ahms5:
```suggestion
        Wraps :py:func:`scipy:scipy.spatial.transform.Rotation.from_matrix`.
```
upps

pyfar/classes/rotation.py
ahms5: Good. I think all other classes except Coodinates is singlar.

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1133 @@
...
+    vectors and it is tied to the object's local coordinate system.
+    Alternatively the object's rotation can be illustrated with help of the
+    right hand: Thumb (view), forefinger (up) and middle finger (right).
+
+    Examples
+    --------
+    >>> from pyfar import Rotation
+    >>> views = [[1, 0, 0], [2, 0, 0]]
+    >>> ups = [[0, 1, 0], [0, -2, 0]]
+    >>> rotations = Rotation.from_view_up(views, ups)
```
ahms5:
```suggestion
    >>> import pyfar as pf
    >>> views = [[1, 0, 0], [2, 0, 0]]
    >>> ups = [[0, 1, 0], [0, -2, 0]]
    >>> rotations = pf.Rotation.from_view_up(views, ups)
```
maybe this makes it easier to see that is pyfar Rotaion

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1133 @@
...
+
+    Visualize rotations at certain positions:
+
+    >>> positions = [[0, 0.5, 0], [0, -0.5, 0]]
+    >>> rotations.show(positions)
```
ahms5:
```suggestion
```

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1133 @@
...
+
+    Rotate by 45 degree in x-direction:
+
+    >>> rot_x45 = Rotation.from_euler('x', 45, degrees=True)
```
ahms5:
```suggestion
    >>> rot_x45 = pf.Rotation.from_euler('x', 45, degrees=True)
```

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1133 @@
...
+    >>> rotations = rotations * rot_x45
+    >>> rotations.show(positions)
```
ahms5:
```suggestion
```

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1133 @@
...
+        angles : float or array_like, shape (..., [1 or 2 or 3])
+            Angles specified in radians (`degrees` is False) or degrees
+            (`degrees` is True).
+            Each angle i in the last dimension of `angles` turns around the
+            corresponding axis axis[..., i, :]. The resulting rotation has the
+            shape np.broadcast_shapes(np.atleast_2d(axes).shape[:-2],
+            np.atleast_1d(angles).shape[:-1]) Dimensionless angles are thus
+            only valid for a single axis.
+
+        degrees : bool, optional
```
ahms5: I know that scipy allows this degrees, but the coordinates class we decised to always use radiance. Maybe we should also force radiance in the Rotaion class to stay constistance

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1133 @@
...
+            rotations cannot be mixed in one function call.
+        angles : float or array_like, shape (...,  [1 or 2 or 3])
+            Euler angles specified in radians (`degrees` is False) or degrees
+            (`degrees` is True).
+            Each character in `seq` defines one axis around which `angles`
+            turns. The resulting rotation has the shape
+            np.atleast_1d(angles).shape[:-1]. Dimensionless angles are thus
+            only valid for single character `seq`.
+
+        degrees : bool, optional
```
ahms5: same here

pyfar/classes/rotation.py
ahms5: I would vote for removing this. It does not mean that we need to follow scipy in every case. We should still follow our own notations.

I also feel that giving num and shape is not a very elegant way. maybe we just have cshape as input to keep it easy and similar to other pyfar behaviour.
same comment for the other num shape cases

ahms5 (changes requested):
Thanks, I would maybe make the class less scipy and more pyfar, see comments.

pyfar/classes/rotation.py
hoyer-a: Thanks for catching

pyfar/classes/rotation.py
f-brinkmann: I think `meth` is correct here because it it links the method of the Rotation object: https://www.sphinx-doc.org/en/master/usage/domains/python.html

hoyer-a: This doesn't create a link to scipy docs. Do you want it to be shorter to just display 'from_davenport' instead of the whole path?

tests/test_rotation.py
```diff
@@ -0,0 +1,317 @@
...
+        [0, 1, 0],
+        [0, -1, 0],
+        [0, 0, 1],
+        [0, 0, -1],
+    ],
+)
+def test_from_view_as_view_roundtrip(v1, v2):
+    """Test from_view_up / as_view_up."""
+    if np.all(np.abs(v1) == np.abs(v2)):
+        pytest.skip()
```
f-brinkmann: This would do the same and because it is intended I would prefer it over skip, which somehow suggests something went wrong. Can you maybe also add a comment why these are not done?
```suggestion
        return
```

f-brinkmann (changes requested):
Only one comment from me left after getting the tests running. I agree with comments from ahms5. But thats not a strong opinion

hoyer-a:
Removed _degrees_ from `from-...` methods.
There are some optional degree returns in the `as_...` methods, should i remove those too?

f-brinkmann (approved):
I'm fine with having the option to return angeles in degrees in the `as_` functions.

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1114 @@
...
+    Rotation in the three-dimensional space.
+
+    This class is largely based on
+    :py:class:`scipy:scipy.spatial.transform.Rotation` and wraps all
+    functionality that scipy's Rotation class provides.
+    In addition the pyfar Rotation class adds the creation from perpendicular
+    view and up vectors through :py:func:`~from_view_up`, and the
+    representation as view / up in :py:func:`~as_view_up`.
+
+    A rotation can be visualized with the triple of view, up and right
```
ahms5:
```suggestion
    A rotation can be represented with the triple of view, up and right
```

pyfar/classes/rotation.py
ahms5:
```suggestion
        Wraps :py:meth:`scipy.spatial.transform.Rotation.from_davenport`.
```
i found it

tests/test_rotation.py
```diff
@@ -0,0 +1,319 @@
...
+    # any of views and ups has zero-length
+    views = [[1, 0, 0], [0, 0, 1]]
+    ups = [[0, 1, 0], [0, 0, 0]]
+    match = 'View and Up Vectors must have a length.'
+    with pytest.raises(ValueError, match=match):
+        Rotation.from_view_up(views, ups)
+    # views' and ups' shape must be (..., 3) or (3,)
+    views = [0, 1]
+    ups = [0, 1]
+    match = 'Expected `views` and `ups` to have shape'
```
ahms5, line 101: isnt broadcastable sufficient?

ahms5 (changes requested):
nice, Thanks, I would remove all dgrees as it would be inconstiant and we dont have degrees in coordinates. instead we have a method to convert spherical into degree. here the user can just use numpy.rad2deg directly, if required.

tests/test_rotation.py
hoyer-a: Views and ups could be broadcastable but still have `shape[-1] != 3`.

I removed the 'malformed list' case above since it tested a numpy warning and nothing pyfar related.

hoyer-a:
> nice, Thanks, I would remove all dgrees as it would be inconstiant and we dont have degrees in coordinates. instead we have a method to convert spherical into degree. here the user can just use numpy.rad2deg directly, if required.

Done, removed degrees

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1110 @@
...
+
+# this warning needs to be caught and appears if numpy array are generated
+# from nested lists containing lists of unequal lengths,
+#  e.g., [[1, 0, 0], [1, 0]]
+warnings.filterwarnings("error", category=VisibleDeprecationWarning)
+
+
+class Rotation():
+    """
+    Rotation in the three-dimensional space.
```
mberz:
```suggestion
    Rotation in three-dimensional space.
```

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1110 @@
...
+    :py:class:`scipy.spatial.transform.Rotation` and wraps all
+    functionality that scipy's Rotation class provides.
+    In addition the pyfar Rotation class adds the creation from perpendicular
+    view and up vectors through :py:func:`~from_view_up`, and the
+    representation as view / up in :py:func:`~as_view_up`.
+
+    A rotation can be represented with the triple of view, up and right
+    vectors and it is tied to the object's local coordinate system.
+    Alternatively the object's rotation can be illustrated with help of the
+    right hand: Thumb (view), forefinger (up) and middle finger (right).
```
mberz:
```suggestion
    right hand: Thumb (view), index-finger (up) and middle finger (right).
```
I think that's a more common name

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1110 @@
...
+
+    def __init__(self):
+        raise RuntimeError("Rotation objects must be created using one of "
+                           "the `from_...` methods")
+
+    def __repr__(self):
+        """String representation of Rotation object."""
+        repr_string = \
+              f"pyfar.Rotation with {self.cshape} rotations."
+        return repr_string
```
mberz:
```suggestion
        return f"pyfar.Rotation with {self.cshape} rotations."
```
Does that fit a single line?

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1110 @@
+"""This module contains the Rotation class."""
+from scipy.spatial.transform import Rotation as scRotation
```
mberz:
```suggestion
from scipy.spatial.transform import Rotation as ScipyRotation
```
more readable in the code

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1110 @@
...
+        if quat.ndim==1:
+            return (1,)
+        return quat.shape[:-1]
+
+    @property
+    def csize(self):
+        """Number of rotations."""
+        return np.prod(self.cshape)
+
+
```
mberz:
```suggestion
```

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1110 @@
...
+        rot = scRotation.from_rotvec(rotvec, degrees=False)
+        return cls._from_scipy_rotation(rot)
+
+    @classmethod
+    def from_view_up(cls, views, ups):
+        """
+        Initialize Rotation from a view an up vector.
+
+        Rotations are internally stored as quaternions for better spherical
+        linear interpolation (SLERP) and spherical harmonics operations.
```
mberz:
```suggestion
        linear interpolation (SLERP) .
```
The SH application is not really implemented in here, isn't it?

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1110 @@
...
+        .. [#] https://en.wikipedia.org/wiki/Axis%E2%80%93angle_representation#Rotation_vector
+        """
+
+        return self._rot.as_rotvec(degrees=False)
+
+    def as_view_up(self):
+        """Get Rotation as a view, up, and right vector.
+
+        Rotation are internally stored as quaternions for better spherical
+        linear interpolation (SLERP) and spherical harmonics operations.
```
mberz:
```suggestion
        linear interpolation (SLERP).
```

pyfar/classes/rotation.py
```diff
@@ -0,0 +1,1110 @@
...
+        More intuitionally, they can be expressed as view and and up of vectors
+        which cannot be collinear. In this case are restricted to be
+        perpendicular to minimize rounding errors.
+
+        Returns
+        -------
+        vector_triple: ndarray, shape (..., 3), normalized vectors
+            - views, see :py:func:`Rotation.from_view_up`
+            - ups, see :py:func:`Rotation.from_view_up`
```
mberz: This docstring does not correspond to the returned variables

mberz (changes requested):
Found a couple of small things, mostly docstring related things.

pyfar/classes/rotation.py
hoyer-a: That whole docstring seems outdated.
We are no longer storing rotations as quaternions. I replaced it with a more general description of the view and up vectors.

Same in `as_view_up()`

mberz: I feel that the reference to spherical harmonics still does not apply here.

pyfar/classes/rotation.py
mberz: same as comment above (or below, not sure anymore)

mberz (commented):
Comments on spherical harmonics left.

pyfar/classes/rotation.py
hoyer-a: Apparently forgot to push the changes

mberz (approved):
Nice, thanks!

PR #908, opened 2026-02-20
[Maint]  update CircleCI python versions

mberz:

- Update the python version in all jobs executed on a single python version to 3.14
- Fix additional failing ruff rule checks (zip is required to use explicit)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -173,7 +173,7 @@ workflows:
           matrix:
             parameters:
               version:
-                - "3.12"
+                - "3.14"
           requires:
             - build_and_test
 
@@ -181,7 +181,7 @@ workflows:
           matrix:
             parameters:
               version:
-                - "3.12"
+                - "3.14"
           requires:
             - build_and_test
 
@@ -197,7 +197,7 @@ workflows:
           matrix:
             parameters:
               version:
-                - "3.12"
+                - "3.14"
           requires:
             - build_and_test
 
@@ -226,7 +226,7 @@ workflows:
           matrix:
             parameters:
               version:
-                - "3.12"
+                - "3.14"
           requires:
             - build_and_test
           filters:
@@ -240,7 +240,7 @@ workflows:
           matrix:
             parameters:
               version:
-                - "3.12"
+                - "3.14"
           requires:
             - build_and_test
           filters:
@@ -268,7 +268,7 @@ workflows:
           matrix:
             parameters:
               version:
-                - "3.12"
+                - "3.14"
           requires:
             - build_and_test
             - ruff
--- a/pyfar/dsp/filter/gammatone.py
+++ b/pyfar/dsp/filter/gammatone.py
@@ -457,7 +457,7 @@ class GammatoneBands():
 
         # apply phase shift, delay, and gain
         for bb, (phase_factor, delay, gain) in enumerate(
-            zip(self._phase_factors, self._delays, self._gains, strict=False),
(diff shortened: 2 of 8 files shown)


.circleci/config.yml
f-brinkmann, line 1: I wish circle had a 'latest' flag

f-brinkmann (approved):
Never notived that zip by default does not raise an error if iterables are not of equal length. Thank you ruff and mberz!

.circleci/config.yml
ahms5: copier will solve this issue :)

pyfar/dsp/filter/gammatone.py
```diff
@@ -456,9 +456,9 @@ def reconstruct(self, real, imag):
         time = real.time.copy() + 1j * imag.time.copy()
 
         # apply phase shift, delay, and gain
-        for bb, (phase_factor, delay, gain) in enumerate(zip(
-                self._phase_factors, self._delays, self._gains)):
-
+        for bb, (phase_factor, delay, gain) in enumerate(
+            zip(self._phase_factors, self._delays, self._gains, strict=False),
```
ahms5:
```suggestion
            zip(self._phase_factors, self._delays, self._gains, strict=True),
```

why not True, if the lengths of the arrays are not equal, sth went wrong. Strict=True will return an ValueError for this case.
this comment also refers to the other cases

ahms5 (changes requested):
thank you, I'm just wondering about the choice of strict=False, maybe you can crosscheck.

pyfar/dsp/filter/gammatone.py
mberz: Haha, I assume my brain was already toast at that point. Thanks for noticing. strict=True is actually what I wanted, but not what I actually did 🙃

tests/test_filter.py
```diff
@@ -357,7 +357,7 @@ def test_reconstructing_fractional_octave_bands_filter_slopes():
     # test different filter slopes against reference
     x = pf.signals.impulse(2**10)
 
-    for overlap, slope in zip([1, 1, 0], [0, 3, 0]):
+    for overlap, slope in zip([1, 1, 0], [0, 3, 0], strict=False):
```
hoyer-a: Any particular reason why `strict=False` here? Maybe stick to `strict=True` in such cases even though overlap and slope are defined explicitly in `zip()`.

tests/test_utils.py
```diff
@@ -87,7 +89,9 @@ def test_broadcast_cdims(cdim, reference):
     cdims = [len(signal.cshape) for signal in signals]
 
     broadcasted = pf.utils.broadcast_cdims(signals, cdim)
-    for signal, broadcast, cdim in zip(signals, broadcasted, cdims):
+    for signal, broadcast, cdim in zip(
```
hoyer-a, line 92: Also here

tests/test_utils.py
```diff
@@ -115,7 +119,7 @@ def test_concatenate_channels(data_second, domains):
                pf.Signal(data_second, sr, fft_norm='rms'))
 
     # force signal domain
-    for signal, domain in zip(signals, domains):
+    for signal, domain in zip(signals, domains, strict=False):
```
hoyer-a: And here

hoyer-a (commented):
Was this missed or is it intentional?

mberz:
> Was this missed or is it intentional?

Don't think so, thanks for catching those.

ahms5 (approved):

hoyer-a (approved):

PR #910, opened 2026-02-25
[BUG] cross fade function

### LINKED ISSUE #909: [BUG] `pf.dsp._cross_fade` does not sum to 1
The following should return an array of ones, but it does not. Fixing the cross-fade is easy and should be done for pyfar 0.8. where this is used in the new `RegularizedSpectrumInversion` class

```python
from pyfar.dsp.dsp import _cross_fade

first = np.ones(100)
second = np.ones(100)

faded = _cross_fade(first, second, [25, 75])

plt.plot(faded)
plt.show()
```

Image

f-brinkmann:

Closes #909 

- make sure cross-fade sums to one
- improve comments and name of test
- add test for fixed behaviour


mberz:
See https://github.com/pyfar/pyfar/issues/909#issuecomment-3967642802

mberz (changes requested):
Same comment as above ;)

PR #911, opened 2026-02-26
[Maint] deprecate regularized inversion

### LINKED ISSUE #895: Deprecate `pyfar.dsp.regularized_specrtrum_inversion`
We still need to add deprecation warnings and tests for `pyfar.dsp.regularized_spectrum_inversion` which will be deprecated in pyfar 0.10.0. Tests should be added to `tests/test_deprecations`.

I started as suggested with the first step suggested below in #911. The second step is a little tricky because `pyfar.dsp.deconvolve` passes key-value arguments to the deprecated function. I see two options of which I prefer the first:

1. We deprecate the key-value arguments in `deconvolve`:
  - if key-value arguments are passed, we call the old function `pyfar.dsp.regularized_spectrum_inversion` and raise a deprecation warning stating that key-value arguments will be deprecated in 0.10.0
  - if no key-value arguments are passed we use the new class `dsp.RegularizedSpectrumInversion` with default parameters. This is fine for deconvolution with sweeps. If users want more control, they can still manually reguarize and deconvolve.
 
2. We check the names of the key-value args and decide which function to use
  - First step as above
  - use the new `dsp.RegularizedSpectrumInversion` with key-value arguments.

f-brinkmann:

Part of #895

- deprecate `regularized_spectrum_inversion`
- add deprecation tests

### DIFF
--- a/pyfar/dsp/dsp.py
+++ b/pyfar/dsp/dsp.py
@@ -4,6 +4,7 @@ import numpy as np
 from scipy import signal as sgn
 import pyfar
 from pyfar.dsp import fft
+from pyfar.classes.warnings import PyfarDeprecationWarning
 import warnings
 import scipy.fft as sfft
 from typing import Literal
@@ -950,6 +951,11 @@ def regularized_spectrum_inversion(
             numerical aspects of linear inversion. Philadelphia: SIAM, 1998.
 
     """
+
+    warnings.warn(("'regularized_spectrum_inversion' will be deprecated in "
+                   "pyfar 0.10.0 in favor 'RegularizedSpectrumInversion'"),
+                   PyfarDeprecationWarning, stacklevel=2)
+
     if not isinstance(signal, pyfar.Signal):
         raise ValueError("The input signal needs to be of type pyfar.Signal.")
 
--- a/pyfar/dsp/dsp.py
+++ b/pyfar/dsp/dsp.py
@@ -953,7 +953,7 @@ def regularized_spectrum_inversion(
     """
 
     warnings.warn(("'regularized_spectrum_inversion' will be deprecated in "
-                   "pyfar 0.10.0 in favor 'RegularizedSpectrumInversion'"),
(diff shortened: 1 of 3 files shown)


pyfar/dsp/dsp.py
```diff
@@ -950,6 +951,11 @@ def regularized_spectrum_inversion(
             numerical aspects of linear inversion. Philadelphia: SIAM, 1998.
 
     """
+
+    warnings.warn(("'regularized_spectrum_inversion' will be deprecated in "
+                   "pyfar 0.10.0 in favor 'RegularizedSpectrumInversion'"),
```
hoyer-a:
```suggestion
                   "pyfar 0.10.0 in favor of 'RegularizedSpectrumInversion'"),
```

tests/test_deprecations.py
```diff
@@ -311,3 +311,18 @@ def test_deprecations_fractional_octave_frequencies():
     if version.parse(pf.__version__) >= version.parse('0.10.0'):
         with pytest.raises(AttributeError):
             pfilt.fractional_octave_frequencies()
+
+
+def test_deprecations_regularized_spectrum_inversion():
+    message = re.escape(
+        ("'regularized_spectrum_inversion' will be deprecated in "
+         "pyfar 0.10.0 in favor 'RegularizedSpectrumInversion'"))
```
hoyer-a:
```suggestion
         "pyfar 0.10.0 in favor of 'RegularizedSpectrumInversion'"))
```

hoyer-a (changes requested):

mberz:
We're starting to enter dangerous territory, since this is part of an example notebook and the gallery does not have versioning in place https://github.com/pyfar/gallery/issues/101

hoyer-a (approved):
Ruff is failing, it's the `zip()` thing from https://github.com/pyfar/pyfar/pull/908, so probably ok not to fix it here.

ahms5 (approved):

PR #918, opened 2026-03-10
deprecate Orientations

### LINKED ISSUE #917: MAINT: Deprecate Orientation class
In #904 we've introduced the `Rotation` class as a replacement for the `Orientation` class. We still need to deprecate the `Orientation` class

hoyer-a:
Deprecate Orientations in favor of Rotation

Closes #917

### DIFF
--- a/pyfar/classes/orientations.py
+++ b/pyfar/classes/orientations.py
@@ -2,6 +2,7 @@
 from scipy.spatial.transform import Rotation
 import numpy as np
 import warnings
+from pyfar.classes.warnings import PyfarDeprecationWarning
 
 import pyfar as pf
 if np.__version__ < '2.0.0':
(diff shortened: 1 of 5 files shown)


pyfar/classes/orientations.py
```diff
@@ -64,6 +65,9 @@ def __init__(self, quat=None, normalize=True, copy=True, **kwargs):
         if quat is None:
             quat = np.array([0., 0., 0., 1.])
         super().__init__(quat, copy=copy, normalize=normalize, **kwargs)
+        warnings.warn(("`Orientations` will be deprecated in pyfar v0.10.0 in "
+                       "favor of `Rotation`."),
+            PyfarDeprecationWarning, stacklevel=2)
```
mberz, line 67: Doesn't it make sense to move this to the first line of the constructor so that it will always be shown, even before a scipy error may occur?

hoyer-a: True

f-brinkmann (changes requested):
Thx for starting this. Can you please ignore the deprecation warnings in `test_orientations.py`? Otherwise the tests get cluttered with intended warnings and important warnings might be missed.

hoyer-a:
> Thx for starting this. Can you please ignore the deprecation warnings in `test_orientations.py`? Otherwise the tests get cluttered with intended warnings and important warnings might be missed.

Done

mberz (approved):

f-brinkmann (approved):

PR #923, opened 2026-04-02
FIX: Avoid checking len on zero length numpy arrays

### LINKED ISSUE #922: Unneccessary `TypeError` for `sampling_rate` argument
Hi guys,
I broke it (again)

MWE:
```python
#!/usr/bin/env python3

import numpy as np
from pyfar import Signal

fs = np.array(44100)
sig = Signal(np.zeros(10), sampling_rate=fs)
```

explodes with

```bash
TypeError: len() of unsized object
```

artpelling:

Closes #922

### DIFF
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -965,12 +965,13 @@ class Signal(FrequencyData, TimeData):
         # unpack iterable
         if hasattr(value, '__iter__'):
             value = np.atleast_1d(np.asarray(value)) # fix #922
-            assert len(value) != 0
-            if len(value) != 1:
+            if len(value) == 0:
+                raise ValueError("Sampling rate cannot be empty!")
+            elif len(value) != 1:
                 raise ValueError("Multirate signals are not supported.")
             value = value[0]
-            elif not isinstance(value, (int, float)):
-                raise ValueError("Sampling rate needs to be number.")
+            if not isinstance(value, (int, float)):
+                raise ValueError("Sampling rate needs to be a number.")
         self._sampling_rate = value
 
     @property
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -772,7 +772,7 @@ class Signal(FrequencyData, TimeData):
         """
         # unpack iterable
         if hasattr(sampling_rate, '__iter__'):
-            sampling_rate = np.atleast_1d(np.asarray(sampling_rate))  # fix #922
+            sampling_rate = np.atleast_1d(np.asarray(sampling_rate)) # fix #922
             assert len(sampling_rate) != 0
             if len(sampling_rate) != 1:
                 raise ValueError("Multirate signals are not supported.")
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -770,8 +770,9 @@ class Signal(FrequencyData, TimeData):
         """
         Create audio Signal with time or frequency data and sampling rate.
         """
-        # unpack array
+        # unpack iterable
         if hasattr(sampling_rate, '__iter__'):
+            sampling_rate = np.atleast_1d(np.asarray(sampling_rate))  # fix #922
             assert len(sampling_rate) != 0
             if len(sampling_rate) != 1:
                 raise ValueError("Multirate signals are not supported.")
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -964,7 +964,7 @@ class Signal(FrequencyData, TimeData):
     def sampling_rate(self, value):
         # unpack iterable
         if hasattr(value, '__iter__'):
-            value = np.atleast_1d(np.asarray(value)) # fix #922
+            value = np.atleast_1d(np.asarray(value)).flatten()  # fix #922
             if len(value) == 0:
                 raise ValueError("Sampling rate cannot be empty!")
             elif len(value) != 1:
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -970,7 +970,7 @@ class Signal(FrequencyData, TimeData):
             elif len(value) != 1:
                 raise ValueError("Multirate signals are not supported.")
             value = value[0]
-        if not np.issubdtype((int, float), type(value)):
+        if not isinstance(value, (int, float, np.integer, np.floating)):
             raise ValueError("Sampling rate needs to be a number.")
         self._sampling_rate = value
 
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -969,6 +969,8 @@ class Signal(FrequencyData, TimeData):
             if len(value) != 1:
                 raise ValueError("Multirate signals are not supported.")
             value = value[0]
+            elif not isinstance(value, (int, float)):
+                raise ValueError("Sampling rate needs to be number.")
         self._sampling_rate = value
 
     @property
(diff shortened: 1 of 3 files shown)


pyfar/classes/audio.py
```diff
@@ -770,8 +770,9 @@ def __init__(
         """
         Create audio Signal with time or frequency data and sampling rate.
         """
-        # unpack array
+        # unpack iterable
         if hasattr(sampling_rate, '__iter__'):
+            sampling_rate = np.atleast_1d(np.asarray(sampling_rate))  # fix #922
```
h-chmeruk: This line is too long, which causes a ruff error

h-chmeruk (changes requested):
Thank you for finding and fixing this bug!
Could you also please add some new tests or adjust the existing ones to verify that the proposed changes actually work and to make sure that your implementation covers most cases?

pyfar/classes/audio.py
```diff
@@ -770,8 +770,9 @@ def __init__(
         """
         Create audio Signal with time or frequency data and sampling rate.
         """
-        # unpack array
+        # unpack iterable
         if hasattr(sampling_rate, '__iter__'):
+            sampling_rate = np.atleast_1d(np.asarray(sampling_rate)) # fix #922
             assert len(sampling_rate) != 0
             if len(sampling_rate) != 1:
                 raise ValueError("Multirate signals are not supported.")
```
f-brinkmann: Would this also work. Would be a little simpler
```suggestion
            sampling_rate = np.squeeze(np.asarray(sampling_rate)) # fix #922
            if sampling_rate.size != 1:
                raise ValueError("Multirate signals are not supported.")
```

f-brinkmann (changes requested):
@artpelling thanks for finding. Revisiting the code, I think we would probably implement it in a different way now.

1. Move the check the the setter `Signal.sampling_rate`
2. Call the setter in `Signal.__init__`

I would suggest to do this here as well.

pyfar/classes/audio.py
artpelling: No :^)

pyfar/classes/audio.py
```diff
@@ -969,6 +962,13 @@ def sampling_rate(self):
 
     @sampling_rate.setter
     def sampling_rate(self, value):
+        # unpack iterable
+        if hasattr(value, '__iter__'):
+            value = np.atleast_1d(np.asarray(value)) # fix #922
+            assert len(value) != 0
+            if len(value) != 1:
+                raise ValueError("Multirate signals are not supported.")
+            value = value[0]
```
ahms5, line 972:
```suggestion
            value = value[0]
            elif not isinstance(value, (int, float)):
                raise ValueError("Sampling rate needs to be number.")
```

if I see it correctly, the current implementation allows non-number assignment of ``sampling_rate``.

tests/test_audio_signal.py
```diff
@@ -288,13 +288,25 @@ def test_getter_sampling_rate():
...
+def test_sampling_rate_parsing(fs):
+    Signal([0], fs)
+    Signal([0], np.array(fs))
+
+
+def test_sampling_rate_multirate():
+    match="Multirate signals are not supported."
+    with pytest.raises(ValueError, match=match):
+        Signal(1, [1,2])
+
```
ahms5, line 308:
```suggestion
def test sampling_rate_no_a_number():
    match="Sampling rate needs to be number."
    with pytest.raises(ValueError, match=match):
        Signal(1, 'string')
```

ahms5 (changes requested):
thanks, looks fine, I just found one point.

sikersten (dismissed):
Thank you for fixing!

tests/test_audio_signal.py
```diff
@@ -288,13 +288,31 @@ def test_getter_sampling_rate():
...
+
+def test_sampling_rate_is_a_number():
+    match="Sampling rate needs to be number."
```
hoyer-a:
```suggestion
    match="Sampling rate needs to be a number."
```

pyfar/classes/audio.py
```diff
@@ -969,6 +962,16 @@ def sampling_rate(self):
 
     @sampling_rate.setter
     def sampling_rate(self, value):
+        # unpack iterable
+        if hasattr(value, '__iter__'):
+            value = np.atleast_1d(np.asarray(value)) # fix #922
```
hoyer-a: I think we need to flatten here, otherwise for something like `fs=[[1]]` we still get an array when doing `value[0]`. This should be one reason why test_sampling_rate_parsing fails...

```suggestion
            value = np.atleast_1d(np.asarray(value)).flatten()  # fix #922
```

pyfar/classes/audio.py
```diff
@@ -969,6 +962,16 @@ def sampling_rate(self):
...
+            if len(value) == 0:
+                raise ValueError("Sampling rate cannot be empty!")
+            elif len(value) != 1:
+                raise ValueError("Multirate signals are not supported.")
+            value = value[0]
+            if not isinstance(value, (int, float)):
```
hoyer-a: ... the other reason probably is, that after using `np.atleast_1d` the return value is a numpy data type like np.int64, which is not covered by (int, float)

hoyer-a (changes requested):
Thanks, i have some suggestions to get the tests running and make it more robust

pyfar/classes/audio.py
artpelling: good catch! i thought this was caught by the `len != 1` check but thats obviously not the case

artpelling:
I applied the fixes and fixed the "is a number check". The indentation was wrong and I used `np.issubdtype` for the check because I think that now achieves exactly what we want.

pyfar/classes/audio.py
```diff
@@ -969,6 +962,16 @@ def sampling_rate(self):
...
     def sampling_rate(self, value):
+        # unpack iterable
+        if hasattr(value, '__iter__'):
+            value = np.atleast_1d(np.asarray(value)).flatten()  # fix #922
+            if len(value) == 0:
+                raise ValueError("Sampling rate cannot be empty!")
+            elif len(value) != 1:
+                raise ValueError("Multirate signals are not supported.")
+            value = value[0]
+        if not np.issubdtype((int, float), type(value)):
```
hoyer-a: We could also use the ABC from numbers for this check

```suggestion
        if not isinstance(value, numbers.Real):
```

hoyer-a (changes requested):
Thanks! 
Tests are still failing because of the number check. Not sure what's the best way to do it, but this worked locally.

pyfar/classes/audio.py
artpelling: hmm, I've checked tests before pushing but only the tests in this file. Looks like the other tests are failing. This is a little more complex than expected. I will first look into what's going on. It seems like we are missing a test case in the tests for `sampling_rate`.

artpelling:
Ok now it is only the docs that are failing.

hoyer-a:
> Ok now it is only the docs that are failing.

This is a known issue with the current sphinx version, this should fix the issue https://github.com/sphinx-doc/sphinx/pull/14391

hoyer-a (dismissed):
Thanks

mberz:
I had to change the target branch @hoyer-a & @sikersten please check if your reviews are still recent.

hoyer-a (approved):

mberz (approved):

PR #924, opened 2026-04-06
use depedency groups instead of extras

artpelling:
I would suggest to use [`dependency-groups`](https://packaging.python.org/en/latest/specifications/dependency-groups/) instead of `project.optional-dependencies` for all development dependencies.

The change is pretty minor but is the recommended way of handling dev dependencies. The main difference is that `dependency-groups` are not included when packaging, making the package smaller whereas `optional-dependencies` are meant for package `extras` with optional features (I don't think any pyfar package has those, yet).

The other (minor) benefit is that it makes the dev workflow a bit cleaner, especially with `uv`, where `dev` is activated by default. The linting command is then `uv run ruff check` instead of `uv run --with ".[dev]" ruff check`. Also, it removes the problem with the `[` in some shells. The pip command becomes `pip install --group dev . `.

- change `pyproject.toml`
- update installation instructions in `CONTRIBUTING.rst`
- update CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,7 @@ jobs:
           command: sudo apt-get update && sudo apt-get install -y libsndfile1
       - run:
           name: install dependencies
-          command: pip install ".[tests]"
+          command: pip install --group tests .
       - run:
           name: Run tests
           command: pytest
@@ -59,7 +59,7 @@ jobs:
       - checkout
       - run:
           name: install dependencies
-          command: pip install ".[tests]"
+          command: pip install --group tests .
       - run:
           name: Ruff
           command: ruff check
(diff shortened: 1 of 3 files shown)


ahms5 (commented):
Sounds like a better way to handle it. Looks like matplotlib uses it, while scipy doesnt. I have two thoughts about it:
- How could we than install the test requirements from the wheel, this will be introduced soon: https://github.com/pyfar/spharpy/pull/310/changes
- This structural change would need to be applied to each repo, as it is not really urgant, I would recomment to introduce it in the [template](https://github.com/pyfar/template-pyfar-package) repro, in this way it will be applied to each package automatically, if we dicide to change it.

PR #929, opened 2026-04-17
constain sphinx to versions 8.x.x

mberz:

Not a reported issue

- Constrain Sphinx to versions 8.x.x

This constrain can hopefully relaxed after https://github.com/sphinx-doc/sphinx/pull/14391 is merged

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,7 +55,7 @@ tests = [
     "coverage",
 ]
 docs = [
-    "sphinx",
+    "sphinx>=8.0.0,<9.0.0",
     "autodocsumm>=0.2.14",
     "pydata-sphinx-theme",
     "sphinx_mdinclude",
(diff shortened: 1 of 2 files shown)


ahms5 (approved):

hoyer-a (approved):

pyproject.toml
```diff
@@ -55,7 +55,7 @@ tests = [
     "coverage",
 ]
 docs = [
-    "sphinx",
+    "sphinx>=8.0.0,<9.0.0",
```
ahms5, line 58:
```suggestion
    "sphinx!=9.0,!=9.1.0",
```

how about excluding the known versions with the warning explicitly instead of restricting to 8.x

ahms5 (changes requested):
I would prehibt just the known versions which are not working. If its fixed in the next version we dont need change the dependencies again.

pyproject.toml
sikersten: +1 for @ahms5's solution

mberz: They are not known, since the fix is not released yet.

ahms5: Known for now. Hopefully it will be fixed in the next version.

f-brinkmann: But if it does not get fixed in the specific version, we have to touch it again. I'm in favour of the current solution.

f-brinkmann (approved):

ahms5 (approved):

mberz:
We decided to merge this (required for release workflow to successfully pass.

mberz:
TODO: Make sure this get's merged back into develop since tests fail there as well

pyproject.toml
mberz: We decided against excluding only specific versions, since they are not released yet.

PR #931, opened 2026-04-28
[Docs] Add sphinx version switcher

hoyer-a:

Closes #https://github.com/pyfar/gallery/issues/159

- Add sphinx version switcher to top navbar
- Auto create versions.json at runtime from available pypi versions + latest tag

- [ ] remove custom css

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -107,6 +107,7 @@ intersphinx_mapping = {
 
 html_theme = 'pydata_sphinx_theme'
 html_static_path = ['_static']
+html_extra_path = ['resources']
 html_css_files = ['css/custom.css']
 html_js_files = ["custom.js"]
 html_logo = 'resources/logos/pyfar_logos_fixed_size_pyfar.png'
@@ -130,10 +131,10 @@ rtd_version = os.environ.get("READTHEDOCS_VERSION", switcher_version)
 rtd_url = os.environ.get("READTHEDOCS_CANONICAL_URL")
 
 if rtd_url:
-    json_url = f"{rtd_url}/_static/versions.json"
+    json_url = f"{rtd_url}/resources/versions.json"
 else:
     json_url = \
-        f"https://{project}.readthedocs.io/en/{rtd_version}/_static/versions.json"
+        f"https://{project}.readthedocs.io/en/{rtd_version}/resources/versions.json"
 
 html_theme_options = {
     "navbar_start": ["navbar-logo"],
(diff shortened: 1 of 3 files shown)


hoyer-a:
With all versions included, the list exceeds the display.
I think we can also make it scrollable by adding something like

```
.bd-header .dropdown-menu {
    max-height: 300px;
    overflow-y: auto;
}
```
to custom.css on gallery.

docs/conf.py
```diff
@@ -204,3 +225,43 @@ def download_files_from_gallery(link, folders_in):
...
+# Add stable entry
+output.append({
+    "name": f"{versions[-1]} (stable)",
+    "version": str(versions[-1]),
+    "url": f"{base_url}/stable/",
+    "preferred": True
+})
+
+# Add latest entry
+output.append({
```
ahms5, line 258: the question is, if we really want latest and/or develop?

docs/conf.py
```diff
@@ -204,3 +225,43 @@ def download_files_from_gallery(link, folders_in):
...
+    "name": "latest",
+    "version": "develop",
+    "url": f"{base_url}/latest/",
+})
+
+ouput = output.reverse()
+
+# Save to JSON file
+with open("_static/versions.json", "w") as f:
```
ahms5:
```suggestion
with open("resources/versions.json", "w") as f:
```

this is not really static

docs/conf.py
```diff
@@ -204,3 +225,43 @@ def download_files_from_gallery(link, folders_in):
 shortcuts_path = os.path.join("resources", "plot_shortcuts.rst")
 with open(shortcuts_path, "w") as f_id:
     f_id.writelines(shortcuts)
+
+
+# -- get versions for version switcher and write to docs/_static/versions.json
```
ahms5:
```suggestion
# -- get versions for version switcher and write to docs/resources/versions.json
```

ahms5 (changes requested):
looks nice thank you. I have some smaller comments below and It would be cool to make the drop down scroleable, in my change I cannot select before v0.4.1 as the list is to long for my window.

ahms5 (changes requested):
thanks, nice, would it be possible to make the dropdown scrolable, as I cannot select each version because the list is too long.

hoyer-a:
Switcher is now scrollable. I pushed _custom.css_ for preview and will remove it & open a pull on gallery once approved.

docs/conf.py
hoyer-a: Honestly, i don't even know what the latest tag is pointing to. Is it develop?

h-chmeruk:
Thanks for implementing this!
I was thinking whether we should perhaps add the changes from the `docs/conf.py` file to the copier template in the `template-pyfar-package` repository? Once you have finished this PR and we have merged it, we can create a PR in `template-pyfar-package`

hoyer-a:
Yes, that would be great!

mberz:
I've removed the milestone since this seems to require some work (and does not classify as a typical bugfix I guess).

PR #933, opened 2026-05-07
exclude broken sphinx version

ahms5:

replaces #929

- Sphinx version 9.0.x is working well, only 9.1.0 (currently latest) is not.
- hopefully it will be fixed in the next version.
- instead of restricting to <9 as suggested in #929 , i would just exclude the version which is not working. If they fix it in the next version, we would not need to revert this change. 

similary we could do that in the other packages too.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -5,7 +5,7 @@ History
 Fixed
 ^^^^^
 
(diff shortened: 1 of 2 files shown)


ahms5:
sphinx 9.0 is not working

PR #884, opened 2026-01-09
[ENH] Add convenience function to attach fraction octave tickers to an axis

mberz:
### Changes proposed in this pull request:

- Convenience function to attach fraction octave tickers to an axis
- Could be made more modular or similar functions could be added for other kinds of often used tickers

PR #937, opened 2026-05-21
[ENH] Ticker module: Remove custom LogFrequencyLocator

sikersten:
Part of adding the ticker module in #877.

The LogFrequencyLocator is a simple wrapper of matplotlib's [LogLocator](https://matplotlib.org/stable/api/ticker_api.html#matplotlib.ticker.LogLocator), just called with the parameter _subs_ differing from matplotlib's default. We had a similar situation with the EngFormatter in #935, where we decided against a custom wrapper. In this PR, I suggest to remove the LogFrequencyLocator and simply call the matplotlib LogLocator in the plot functions for consistency as well.

Instead of unnecessarily wrapping matplotlib's formatters and locators , I suggest adding convenience functions and example notebooks as part of #877 (similar to #884). For example, a frequency axis convenience function would apply matplotlib's EngFormatter and LogLocator with the pyfar specific options.

### DIFF
--- a/pyfar/plot/ticker.py
+++ b/pyfar/plot/ticker.py
@@ -48,61 +48,6 @@ class FractionalOctaveLocator(FixedLocator):
         super().__init__(ticks)
 
 
-class LogFrequencyLocator(LogLocator):
-    """
-    Log-locator particularly suited for frequency axes.
-
-    This locator is a wrapper of :class:`matplotlib.ticker.LogLocator`
-    with default subdivisions optimized for frequency axes.
-    The locator is used per default for frequency axes in pyfar plots.
-
-
-    Parameters
-    ----------
-    base : float, default: 10.0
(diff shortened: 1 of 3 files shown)


ahms5 (approved):

ahms5:
please have a look at ruff, otherwise approved.

pyfar/plot/ticker.py
f-brinkmann: Not used anymore (see ruff)
```suggestion
```

f-brinkmann (changes requested):
Please fix ruff. I think not adding a history entry is ok here and assume it will be done as part of #877

f-brinkmann (approved):
Failing tests are already fixed on main. Since this goes into a separate dev branch, merging is fine for me. Tests can be fixed as soon as patch release 0.8.1 is done.

PR #941, opened 2026-06-17
FIX: Improve sub-sample delay estimation

### LINKED ISSUE #940: [BUG] `find_impulse_response_delay` fails due to small search region
I noticed that `find_impulse_repsonse_delay` fails to extract the ITD for the FABIAN dataset for certain source directions. I investigated and this is probably due to a too narrow search window around the correlation peak.

Minimal example:
```python
import pyfar as pf

ir = pf.Signal([2, 0, -2, 0, -1], 44100)
pf.dsp.find_impulse_response_delay(ir)
```

fails with 

```bash
  File "/home/pelling/projects/pyfar/mwe.py", line 4, in <module>
    pf.dsp.find_impulse_response_delay(ir)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^
  File "/home/pelling/projects/pyfar/pyfar/dsp/dsp.py", line 1465, in find_impulse_response_delay
    search_region_poly = np.polyfit(
        search_region_range[mask]-argmax, search_region[mask], N)
  File "/home/pelling/projects/pyfar/.venv/lib/python3.13/site-packages/numpy/lib/_polynomial_impl.py", line 645, in polyfit
    raise TypeError("expected non-empty vector for x")
TypeError: expected non-empty vector for x
```
on `develop`

artpelling:
The old implementation tried to fit a polynomial around the peak of the Hilbert transform correlation. The "interval around the peak" was somewhat hardcoded and could imply that no zero-crossing is included in the interval. This version selects the search interval such that the zero-crossing closest to the peak is included and then interpolates. This approach is also what is described in the Tanim paper.

Closes #940

- fixes the search area to always include closest zero crossing to the correlation peak
- matches cited reference more closely
- uses new NumPy polynomial interface (old one was deprecated)
- fixes typos


pyfar/dsp/dsp.py
```diff
@@ -1345,6 +1346,47 @@ def time_shift(
     return shifted
 
 
+def _estimate_zero_crossing(lags, values, argmax, order):
+    """Estimate the nearest zero crossing around ``argmax``."""
+    if values[argmax] == 0:
+        return lags[argmax]
+
+    for offset in range(values.size - 1):
+        for left, right in ((argmax - offset - 1, argmax - offset),
```
f-brinkmann: To make it easier to maintain maybe comment
```suggestion
        # iteratively increase window length for searching zero corssings
        for left, right in ((argmax - offset - 1, argmax - offset),
```

pyfar/dsp/dsp.py
```diff
@@ -1345,6 +1346,47 @@ def time_shift(
...
+            if y_2 == 0:
+                return lags[right]
+            if np.sign(y_1) == np.sign(y_2):
+                continue
+
+            x_1, x_2 = lags[left], lags[right]
+
+            # handle linear case explicitly for speed
+            if order == 1:
+                return x_1 - y_1 * (x_2 - x_1) / (y_2 - y_1)
```
f-brinkmann: Isn't this only correct if the search window width ``right-left==1``? Otherwise it would not be a polynomial fit. Maybe remove the if-case?

f-brinkmann (changes requested):
Thanks for noting and fixing. I had one qestion on the if case. In addition, I would suggest to add a brief test using the previously failing code example you posted in the issue.

pyfar/dsp/dsp.py
```diff
@@ -1368,7 +1410,8 @@ def find_impulse_response_delay(impulse_response, N=1):
     impulse_response : Signal
         The impulse response.
     N : int, optional
-        The order of the polynom used for root finding, by default 1.
+        The order of the polynomial used for root finding, by default 1
+        (original Tanim et al. method).
```
h-chmeruk:
```suggestion
        (original Tamim et al. method).
```
Is that a typo, and should it actually be `Tamim` here?

h-chmeruk (changes requested):
Thanks for reporting the bug and fixing it! I would like to join @f-brinkmann in suggesting that the new function could be tested with an additional test.  :)

pyfar/dsp/dsp.py
artpelling: A typo. But in my head rather than my fingers :D

pyfar/dsp/dsp.py
artpelling: Hmm I don't understand your comment regarding `right-left`.

 `order` is the interpolation order, I spelled it out explicitly to avoid going through the whole NumPy interpolation machine for the simple case. It should also work correctly, if we remove the if clause though.

artpelling:
I added a small tests that fails on `main` but passes here.

pyfar/dsp/dsp.py
f-brinkmann: If for example left is 0 and right is 2, `x_1 - y_1 * (x_2 - x_1) / (y_2 - y_1)` is a linear interpolation between `y[0]` and `y[2]`, which differs from a polyfit that would use `y[0]`, `y[1]`, and `y[2]`. I was not sure if this was intended.

Looking at this again, I wonder if we need the loop. Would it work with vectorization if detecting the `lag` that is closest to `argmax` and has a positive gradient, and then perform the polyfit around this position?

However, I did not look into this in detail, so I don't have a good understanding of how much the above effects the results and runtime.

mberz: How much speedup does this yield and how much time does it consume? If it's only marginal, I'd vote for getting rid of the if statement.

pyfar/dsp/dsp.py
```diff
@@ -1345,6 +1346,48 @@ def time_shift(
     return shifted
 
 
+def _estimate_zero_crossing(lags, values, argmax, order):
```
mberz, line 1349: Please provide a docstring for the function. We generally require this for private functions as well.

tests/test_dsp.py
```diff
@@ -894,6 +894,15 @@ def test_impulse_response_delay_sinc(sign):
     npt.assert_allclose(start_samples, delay_samples, atol=1e-3, rtol=1e-4)
 
 
+def test_impulse_response_delay_issue_940():
+    """Test a signal that previously failed the delay search."""
+    ir = pf.Signal([2, 0, -2, 0, -1], 44100)
+
+    start_sample = pf.dsp.find_impulse_response_delay(ir)
+
+    npt.assert_allclose(start_sample, [1], atol=1e-6)
```
mberz: Why would you say the start sample is 1 and not 0?
The signal you seems rather arbitrary to me here.

mberz (changes requested):
Thanks for your fix, I'm a bit uncertain about the choice of the test signal you provide. Please elaborate on this or choose a more representative one.

pyfar/dsp/dsp.py
artpelling: for the FABIAN dataset I get 9 seconds with the direct implementation vs 16 seconds with NumPy. I thought it is nice to pick up free speed with two lines of code but it makes it uglier for sure.

artpelling: > If for example left is 0 and right is 2, `x_1 - y_1 * (x_2 - x_1) / (y_2 - y_1)` is a linear interpolation between `y[0]` and `y[2]`, which differs from a polyfit that would use `y[0]`, `y[1]`, and `y[2]`. I was not sure if this was intended.

I am convinced that the explicit formula is equivalent to the `NumPy` with `order==1`. If you look at `lags[start:start+degree+1]` which enters `Polynomial.fit`, it is just two values, not three.

pyfar/dsp/dsp.py
```diff
@@ -1345,6 +1346,61 @@ def time_shift(
...
+    -------
+    root : float or None
+        Estimated zero-crossing lag, or ``None`` if no positive-gradient zero
+        crossing is found.
+    """
+    if values[argmax] == 0:
+        return lags[argmax]
+
+    left = np.flatnonzero(
+        (values[:-1] * values[1:] < 0) & (values[1:] > values[:-1]))
```
f-brinkmann, line 1376:
```suggestion
    # integer indices after which a zero crossing with positive gradient occurs
    left = np.flatnonzero(
        (values[:-1] * values[1:] < 0) & (values[1:] > values[:-1]))
```

pyfar/dsp/dsp.py
```diff
@@ -1345,6 +1346,61 @@ def time_shift(
...
+    y_1, y_2 = values[left], values[right]
+    linear_roots = x_1 - y_1 * (x_2 - x_1) / (y_2 - y_1)
+    idx = np.argmin(np.abs(linear_roots - lags[argmax]))
+
+    if order == 1:
+        return linear_roots[idx]
+
+    degree = min(order, values.size - 1)
+    start = min(max(0, left[idx] - (degree - 1) // 2),
+                left[idx] + 1, values.size - (degree + 1))
```
f-brinkmann: isn't `left[idx] + 1` always larger than `max(0, left[idx] - (degree - 1) // 2)` and could be omitted?
```suggestion
    # set starting point for higher order fit and make sure that
    # the fitting region is cenetered around the expected root and
    # the end point does not exceed array size
    start = min(max(0, left[idx] - (degree - 1) // 2),
                values.size - (degree + 1))
```

pyfar/dsp/dsp.py
```diff
@@ -1345,6 +1346,61 @@ def time_shift(
...
+    roots = Polynomial.fit(
+        lags[start:start + degree + 1],
+        values[start:start + degree + 1],
+        degree,
+    ).convert().roots()
+    roots = np.real(roots[np.isreal(roots)])
+    roots = roots[(roots >= x_1[idx]) & (roots <= x_2[idx])]
```
f-brinkmann, line 1403:
```suggestion
    # root must within the samples at which the zero crossing
    # with positive gradient occurs
    roots = roots[(roots >= x_1[idx]) & (roots <= x_2[idx])]
```

tests/test_dsp.py
f-brinkmann: I assume the focus was on finding a minimal example where the detection failed before. We could make this even more clear in the docstring, e.g., `"""Test a signal that failed the delay search before pull request #941."""`

f-brinkmann (commented):
Thanks. The code looks fine to me now. I suggested adding comments to make things easier to understand and maintain

tests/test_dsp.py
mberz: My problem with the example is that it is quite arbitrary and the start is not known. Under the assumption of periodicity, the start may even be at the very end, where it goes from -1 to 2. 
So from the example it's not clear if this just now prevents an error, or if it's returning results that are physically meaningful.

artpelling: I agree fully, I had trouble finding _any_ MWE that reproduces the issue, i.e. fails on `develop` but passes here.

I will see if chatty can find a better example that is physically meaningfull. If not, I will investigate the channels of the FABIAN dataset that failed and maybe can extract an MWE from those.

artpelling:
@mberz @f-brinkmann I've updated the test now and hardcoded an excerpt from the FABIAN dataset into the test. This search window used to emit a `RankWarning` on `develop` because only 1 value was included in the search area. The test fails on `develop`, passes on this branch, and is as realistic as it gets :D

f-brinkmann (dismissed):

mberz:
Changed target branch, since this is a bugfix.

mberz:
When testing this PR a bit further, I realized that the implementation fails for 0.75 sample shifts (and other values as well). I checked the against the previous version and realized it's the same issue there as well.

@artpelling Did you maybe encounter similar behavior?

In any case, it might be a good idea to fix this in a separate PR, as it is an unrelated issue.

h-chmeruk:
> When testing this PR a bit further, I realized that the implementation fails for 0.75 sample shifts (and other values as well). I checked the against the previous version and realized it's the same issue there as well.

@mberz, could you please provide an example of what you tried and where exactly the implementation failed? I have tested the `_estimate_zero_crossing()` function with many different logs and values, and it has produced correct results so far. So I suspect the problem lies in the `find_impulse_response_delay()` function itself?
Maybe you could open an issue when you have some time :)

mberz:
I'll do that after @artpelling had the chance to reply here

mberz:
Changed my mind, new issue in #946

artpelling:
> I'll do that after @artpelling had the chance to reply here

replied in #946

pyfar/dsp/dsp.py
```diff
@@ -1345,6 +1346,67 @@ def time_shift(
...
+    roots = Polynomial.fit(
+        lags[start:start + degree + 1],
+        values[start:start + degree + 1],
+        degree,
+    ).convert().roots()
+    roots = np.real(roots[np.isreal(roots)])
+    # root must within the samples at which the zero crossing
+    # with positive gradient occurs
+    roots = roots[(roots >= x_1[idx]) & (roots <= x_2[idx])]
+    if roots.size:
```
mberz, line 1405: Can you please make this if statement more explicit? Which condition is checked here?

pyfar/dsp/dsp.py
```diff
@@ -1442,38 +1503,29 @@ def find_impulse_response_delay(impulse_response, N=1):
...
+                        lags, search_region, argmax, N)
+
+                if root is not None:
+                    start_sample[idx] = root
                 else:
                     start_sample[idx] = np.nan
-                    warnings.warn('Starting sample not found for channel '
-                                  f'{ch}', stacklevel=2)
+                    warnings.warn(
+                        f'Starting sample not found for channel {ch}',
```
mberz:
```suggestion
                        f'Delay sample not found for channel {ch}',
```

mberz (changes requested):
Thanks. I only have two small nitpicks left. Otherwise good to go

mberz (approved):

PR #942, opened 2026-06-24
[BUG] fix deprecated Matplotlib subplot handing

f-brinkmann:

Test failed after a Matplotlib update. The update now raises an error, if `plt.subplots` gets an existing figure to create the subplots. This is fixed using `fig.subplots` instead.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,13 @@
 History
 =======
 
+unreleased
+----------
+
+Fixed
+^^^^^
+- Fixed failing tests due to a Matplotlib update (PR #942)
+
 
 0.8.0 (2026-03-16)
 ------------------
--- a/tests/test_plot.py
+++ b/tests/test_plot.py
@@ -419,8 +419,7 @@ def test_2d_colorbar_options(function, colorbar, handsome_signal_2d):
             function(signal, colorbar=False)
     elif colorbar == "axes":
         # test plotting colorbar to specified axis
-        fig.clear()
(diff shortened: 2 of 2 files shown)


tests/test_plot.py
```diff
@@ -419,8 +419,7 @@ def test_2d_colorbar_options(function, colorbar, handsome_signal_2d):
             function(signal, colorbar=False)
     elif colorbar == "axes":
         # test plotting colorbar to specified axis
-        fig.clear()
```
mberz, line 422: Why has the clear been removed?

f-brinkmann: Its not strictly required. A fresh figure without any axes is spawned in `create_figure`. So I assumed the code would be cleaner without it.

HISTORY.rst
```diff
@@ -2,6 +2,13 @@
 History
 =======
 
+unreleased
+----------
+
+Fixed
+^^^^^
+- Fixed failing tests due to a Matplotlib update (PR #942)
```
hoyer-a: maybe a bit more specific? e.g. something like
_due to updated figure creation / management in Matplotlib 3.11_

hoyer-a (approved):
Thanks, approved! (no strong opinion on the history entry)

HISTORY.rst
f-brinkmann: I was lazy and thought linking the PR number will give the detailed answer, if desired 🙈

mberz:
Please give the PR a descriptive name before we can merge.

PR #945, opened 2026-07-01
[BUG] multidimensional, normalized correlation

pingelit:
### Proposed changes

When calculating a normalised correlation with multidimensional signals, a broadcasting error was raised. The cause was that the normalisation factor was "missing" the last (time) dimension. As NumPy broadcasting is evaluated from the right, the arrays could not be broadcast. 

This PR fixes this issue by appending a new dimension to the normalisation factor. Furthermore, the nd-correlation test is extended to test normalisation as well.

### DIFF
--- a/pyfar/dsp/dsp.py
+++ b/pyfar/dsp/dsp.py
@@ -3018,7 +3018,9 @@ def correlate(signal_1, signal_2, mode='full', normalize=False):
 
     # apply normalization
     if normalize:
-        correlation /= normalization
+        # add an axis to normalization to allow broadcasting
+        # for multi-channel signals
+        correlation /= normalization[..., None]
 
     # compute lags, i.e., times that the second signal was shifted
     # with respect to the first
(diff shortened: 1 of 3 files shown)


f-brinkmann (approved):
Thx for finding and fixing. Sometimes I wonder if I will ever stop making simple mistakes like this :)

pyfar/dsp/dsp.py
```diff
@@ -3018,7 +3018,9 @@ def correlate(signal_1, signal_2, mode='full', normalize=False):
 
     # apply normalization
     if normalize:
-        correlation /= normalization
+        # add an axis to normalization to allow broadcasting
+        # for multi-channel signals
+        correlation /= normalization[..., None]
```
mberz:
```suggestion
        correlation /= normalization[..., np.newaxis]
```
This makes it more explicit what is happening here.

mberz (commented):
Thanks for the bugfix. Looks good to me.
I only have a small nitpick to make it more explicit, but otherwise approved.

mberz (approved):

PR #947, opened 2026-07-06
Add rounding options for find_nearest_time() and find_nearest_frequency() methods

### LINKED ISSUE #867: [ENH] Add rounding options for find_nearest_time/freq for Signal like objects
Currently, these methods simply return the closest time or frequency.
For more fine grained control it may be useful to add functionality to round to the next larger or lower values.

For example rounding to the closest time which is smaller than the specified value.

This could be implemented using an additional input to the method.

h-chmeruk:

Closes #867

- Add a new parameter called `method` to the `find_nearest_time()` and `find_nearest_frequency()` methods. This parameter can take one of the following three values: {`nearest`, `floor`, `ceil`}. 
- Expand the docstrings. 
- Add tests to the files `test_audio_frequency_data.py` and `test_audio_data_time.py`.
- The function’s previous functionality is preserved in `nearest` mode.
- If the specified value (list of values) is outside the range of `self.time` or `self.frequencies`, the first or last index is returned. 

### Open Question
1. How should the functions (`find_nearest_time()` and `find_nearest_frequency()`) behave if `self.times` or `self.frequencies` is empty? Currently, both functions simply return -1 for the `floor` and `ceil` modes, since there are no indices at all when the `times` or `frequencies` arrays are empty. For the `nearest` mode, numpy raises a `ValueError`.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,12 +2,12 @@
 History
 =======
 
-0.9.0 (2026-XX-XX)
-------------------
+unreleased
+----------
 
 Added
 ^^^^^
-- The methods `find_nearest_frequency()` and `find_nearest_time()` were extended with a new parameter `method`." (PR #947)
+- The methods `find_nearest_frequency()` and `find_nearest_time()` were extended with a new parameter `method`. (PR #947)
 
 0.8.0 (2026-03-16)
 ------------------
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,12 @@
 History
 =======
 
+0.9.0 (2026-XX-XX)
+------------------
+
+Added
+^^^^^
+- The methods `find_nearest_frequency()` and `find_nearest_time()` were extended with a new parameter `method`." (PR #947)
 
 0.8.0 (2026-03-16)
 ------------------
--- a/pyfar/classes/audio.py
+++ b/pyfar/classes/audio.py
@@ -33,6 +33,7 @@ from copy import deepcopy
 import warnings
 import deepdiff
 import numpy as np
+from tomlkit import value
 import pyfar.dsp.fft as fft
 from typing import Callable
 
@@ -436,26 +437,12 @@ class TimeData(_Audio):
             Query times outside the range of self.times are mapped to the first
             or last valid index.
         """
-        times = np.atleast_1d(value)
-        indices = np.zeros_like(times).astype(int)
+        if len(self.times) == 0:
+            raise ValueError("The TimeData object must not be empty.")
 
-        if method == "nearest":
-            for idx, time in enumerate(times):
-                indices[idx] = np.argmin(np.abs(self.times - time))
-        elif method == "floor":
-            # find the insertion positions and select the previous indices
-            indices = np.searchsorted(self.times, times, side="right") -1
-            # clip to ensure the index is not out of range of time indices
-            indices = np.clip(indices, 0, len(self.times) - 1)
-        elif method == "ceil":
-            # find the insertion positions
-            indices = np.searchsorted(self.times, times, side="left")
-            # clip to ensure the index is not out of range of time indices
-            indices = np.clip(indices, 0, len(self.times) - 1)
-        else:
-            raise ValueError("Invalid value for the parameter 'method'."
-                    " Has to be 'nearest', 'floor', or 'ceil'.")
-        return np.squeeze(indices)
+        result = _find_nearest_helper(self.times, value, method)
+
+        return result
 
     def _assert_matching_meta_data(self, other):
         """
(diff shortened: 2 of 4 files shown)


pyfar/classes/audio.py
```diff
@@ -409,24 +409,51 @@ def times(self):
...
 
         Parameters
         ----------
         value : float, array-like
             The times for which the indices are to be returned
+        method : {'nearest', 'floor', 'ceil'}, optional
+            Specifies how query times that do not exactly match a sample
+            are handled:
+
+            - ``'nearest'``: Return the index of the nearest time.
```
hoyer-a: Maybe add that this is the default.

pyfar/classes/audio.py
```diff
@@ -601,24 +628,51 @@ def n_bins(self):
...
 
         Parameters
         ----------
         value : float, array-like
             The frequencies for which the indices are to be returned
+        method : {'nearest', 'floor', 'ceil'}, optional
+            Specifies how query frequencies that do not exactly match a bin
+            are handled:
+
+            - ``'nearest'`` Return the index of the nearest frequency.
```
hoyer-a: also mention that this is the default here please

hoyer-a (commented):
Thanks, already looks good to me.

Regarding your question, do we actually want to allow empty Signals, TimeData and FrequencyData objects? If so, i would vote for raising an error when using `find_nearest...`. 

Otherwise, let's move the topic to a new issue.

h-chmeruk:
@hoyer-a thank you for the review! Currently, it is possible to create empty `FrequencyData` and `TimeData` objects, but I have no idea if this was intended or if it is relevant to certain other functions...

jf-som:
I like this feature, and using `np.searchsorted` looks really neat to me!

It looks to me like the two functions are basically the same, right? In that case, I'm always a fan of avoiding redundant code. I think it should be possible here to do something like this, i.e. passing  `value` and `self.times`/`self.frequencies` to a helper function:
```py
    def _find_nearest_helper(value, signal_values, method):
        target_values= np.atleast_1d(value)
        indices = np.zeros_like(target_values).astype(int)

        if method == "nearest":
            for idx, target in enumerate(target_values):
                indices[idx] = np.argmin(np.abs(signal_values - target))
        elif method == "floor":
            # find the insertion positions and select the previous indices
            indices = np.searchsorted(signal_values, target_values, side="right") -1
            # clip to ensure the index is not out of range of available indices
            indices = np.clip(indices, 0, len(signal_values) - 1)
        elif method == "ceil":
            # find the insertion positions
            indices = np.searchsorted(signal_values, target_values, side="left")
            # clip to ensure the index is not out of range of available indices
            indices = np.clip(indices, 0, len(signal_values) - 1)
        else:
            raise ValueError("Invalid value for the parameter 'method'."
                    " Has to be 'nearest', 'floor', or 'ceil'.")
        return np.squeeze(indices)
```
(I copied and edited this on GitHub, didn't tests/ruff)

The actual functions could then both simply forward to this, like
```py
_find_nearest_helper(value, self.time, method)
_find_nearest_helper(value, self.frequencies, method)
```

Also regarding the empty times/frequencies arrays, I too think this function should raise Errors, or alternatively empty pyfar Audio objects should be prevented in the first place.
I suggest adding a check to the helper function either way since it can't hurt, and then moving the empty audio object discussion to its own issue.

Also feel free to request me for review if needed! I'm not part of the pyfar org yet so it doesn't automatically request me alongside the group.

h-chmeruk:
@jf-som, thank you very much for the review! I liked your suggestion a lot :) 
And thanks for mentioning it: I wasn't aware that you weren't part of the pyfar/pyfar team...

pyfar/classes/audio.py
```diff
@@ -1902,3 +1930,27 @@ def _match_fft_norm(fft_norm_1, fft_norm_2, division=False):
...
+
+        values = np.atleast_1d(desired_value)
+        indices = np.zeros_like(values).astype(int)
+
+        if method == "nearest":
+            for idx, value in enumerate(values):
+                indices[idx] = np.argmin(np.abs(self_values - value))
+        elif method == "floor":
+            # find the insertion positions and select the previous indices
+            indices = np.searchsorted(self_values, values, side="right") -1
```
jf-som:
```suggestion
            indices = np.searchsorted(self_values, values, side="right") - 1
```
missing whitespace for consistent formatting

jf-som (approved):
Only one minor formatting hint, otherwise this looks great!

HISTORY.rst
```diff
@@ -2,6 +2,12 @@
 History
 =======
 
+0.9.0 (2026-XX-XX)
```
mberz:
```suggestion
unreleased
```

Please make this unreleased to avoid merge conflicts

pyfar/classes/audio.py
```diff
@@ -409,25 +409,39 @@ def times(self):
...
+            or last valid index.
         """
-        times = np.atleast_1d(value)
-        indices = np.zeros_like(times).astype(int)
-        for idx, time in enumerate(times):
-            indices[idx] = np.argmin(np.abs(self.times - time))
-        return np.squeeze(indices)
+        if len(self.times) == 0:
+            raise ValueError("The TimeData object must not be empty.")
+
```
mberz: The check can also be moved into the helper.

pyfar/classes/audio.py
```diff
@@ -601,25 +615,39 @@ def n_bins(self):
...
+            Query frequencies outside the range of self.frequencies are mapped
+            to the first or last valid index.
         """
-        freqs = np.atleast_1d(value)
-        indices = np.zeros_like(freqs).astype(int)
-        for idx, freq in enumerate(freqs):
-            indices[idx] = np.argmin(np.abs(self.frequencies - freq))
-        return np.squeeze(indices)
+        if len(self.frequencies) == 0:
+            raise ValueError("The FrequencyData object must not be empty.")
```
mberz: This check can also be moved into the helper.

mberz (commented):
I only have two small comments.

hoyer-a (approved):

pyfar/classes/audio.py
h-chmeruk: Thank you for the review!
My only reason for NOT moving it to the helper function was that, within the helper function, we can't determine whether it's a TimeData or a FrequencyData object. Therefore, the message for the ValueError within the helper function would be a bit more general

PR #948, opened 2026-07-06
Level/core

jf-som:
This is the first part of #949, the umbrella Draft PR for the level module.

This PR introduces the shared helper functions that I intend to use for all the standard level functions, and implements the most simple out of all the level functions using these helpers, which is L_Eq.

Since thorough documentation will be very similar/redundant between these level functions, it should be done in a separate PR later, when more/all of the standard level functions are merged into the `level_module` branch. (see #949 for elaboration)

### DIFF
--- a/pyfar/level/_utils.py
+++ b/pyfar/level/_utils.py
@@ -7,8 +7,8 @@ import pyfar as pf
 
 
 def _check_signal_type(signal):
-    """Throws if not a pyfar.Signal, otherwise returns the signal with
-    type hinting.
+    """Raises a TypeError if not a pyfar.Signal, otherwise returns
+    the signal with type hinting.
     """
     if not isinstance(signal, pf.Signal):
         raise TypeError("'signal' parameter must be a pyfar.Signal")
@@ -18,7 +18,7 @@ def _check_signal_type(signal):
 
 def _apply_frequency_weighting(signal, frequency_weighting):
     """Applies frequency weighting to the signal
-    or throws if the weighting name is invalid.
+    or raises a ValueError if the weighting name is invalid.
     """
     if frequency_weighting == "Z":
         return signal
--- a/pyfar/level/_utils.py
+++ b/pyfar/level/_utils.py
@@ -1,0 +1,54 @@
+"""Reusable helper functions for the individual processing steps shared
+between many of the standardized level functions.
+"""
+
+import numpy as np
+import pyfar as pf
+
+
+def _check_signal_type(signal):
+    """Throws if not a pyfar.Signal, otherwise returns the signal with
+    type hinting.
+    """
+    if not isinstance(signal, pf.Signal):
+        raise TypeError("'signal' parameter must be a pyfar.Signal")
+    signal: pf.Signal = signal
+    return signal
+
+
+def _apply_frequency_weighting(signal, frequency_weighting):
+    """Applies frequency weighting to the signal
+    or throws if the weighting name is invalid.
+    """
+    if frequency_weighting == "Z":
+        return signal
+    elif frequency_weighting in ["A", "C"]:
+        return pf.dsp.filter.frequency_weighting_filter(
+            signal, frequency_weighting)
+    else:
+        raise ValueError("Frequency weighting must be 'A', 'C', or 'Z'")
+
+
+def _apply_multi_band(signal, num_octave_band_fractions: int | None):
+    """Applies fractional octave band filtering to the signal or throws if the
+    number of bands is invalid.
+    """
+    error_message = "Number of octave band fractions must be " \
+                    "a positive integer or None"
+    if num_octave_band_fractions is None:
+        return signal
+    elif not isinstance(num_octave_band_fractions, int):
+        raise TypeError(error_message)
+    elif num_octave_band_fractions <= 0:
+        raise ValueError(error_message)
+    else:
+        return pf.dsp.filter.fractional_octave_bands(
+            signal, num_octave_band_fractions)
+
+
+def _energies_to_levels(energies: np.ndarray, reference_pressure: float):
+    """Converts energies to levels in dB relative to the reference pressure.
+    """
+    if reference_pressure == 0:
+        raise ValueError("Reference pressure must not be zero.")
+    return 10 * np.log10(energies / reference_pressure**2)
--- a/pyfar/level/_utils.py
+++ b/pyfar/level/_utils.py
@@ -30,20 +30,13 @@ def _apply_frequency_weighting(signal, frequency_weighting):
 
 
 def _apply_multi_band(signal, num_octave_band_fractions: int | None):
-    """Applies fractional octave band filtering to the signal or throws if the
-    number of bands is invalid.
+    """Applies fractional octave band filtering to the signal or does
+    nothing if `num_octave_band_fractions` is None.
     """
-    error_message = "Number of octave band fractions must be " \
-                    "a positive integer or None"
     if num_octave_band_fractions is None:
         return signal
-    elif not isinstance(num_octave_band_fractions, int):
-        raise TypeError(error_message)
-    elif num_octave_band_fractions <= 0:
-        raise ValueError(error_message)
-    else:
-        return pf.dsp.filter.fractional_octave_bands(
-            signal, num_octave_band_fractions)
+    return pf.dsp.filter.fractional_octave_bands(
+        signal, num_octave_band_fractions)
 
 
 def _energies_to_levels(energies: np.ndarray, reference_pressure: float):
--- a/pyfar/level/standard_levels.py
+++ b/pyfar/level/standard_levels.py
@@ -2,6 +2,7 @@
 
 from typing import Literal
 import numpy as np
+import pyfar as pf
 
 from ._utils import (
     _check_signal_type,
@@ -11,11 +12,12 @@ from ._utils import (
 )
 
 
-def equivalent_continuous_level(signal,
-                                frequency_weighting: Literal["A", "C", "Z"],
-                                num_octave_band_fractions: int | None = None,
-                                reference_pressure: float = 20e-6,
-                                ):
+def equivalent_continuous_level(
+        signal,
+        frequency_weighting: Literal["A", "C", "Z"],
+        num_octave_band_fractions: int | None = None,
+        reference_pressure: float = pf.constants.reference_sound_pressure,
+):
     r"""Calculate the frequency-weighted equivalent continuous sound pressure
     level (Leq).
 
--- a/pyfar/level/standard_levels.py
+++ b/pyfar/level/standard_levels.py
@@ -16,18 +16,19 @@ def equivalent_continuous_level(signal,
                                 num_octave_band_fractions: int | None = None,
                                 reference_pressure: float = 20e-6,
                                 ):
-    r"""Calculates the frequency-weighted equivalent continuous sound pressure
-    level (Leq) for each channel of a signal.
+    r"""Calculate the frequency-weighted equivalent continuous sound pressure
+    level (Leq).
 
-    The levels are calculated according to IEC 61672-1 [#]_.
+    The levels are calculated per channel and according to IEC 61672-1 [#]_.
     For instance, the A-weighted equivalent continuous level is calculated as:
 
     .. math::
         L_\text{Aeq} = 10 \log_{10} \left[ \frac{(1/N) \sum_{n=0}^{N-1}
         p_{\text{A}}^2[n]} {p_0^2} \right] \text{ dB}
 
-    where :math:`N` is the number of samples in the signal and
-    :math:`p_0` is the `reference_pressure`.
+    where :math:`N` is the number of samples in the signal,
+    :math:`p_\mathrm{A}` the A-weighted sound pressure at index :math:`n`,
+    and :math:`p_0` is the `reference_pressure`.
 
     Parameters
     ----------
@@ -39,24 +40,22 @@ def equivalent_continuous_level(signal,
         frequency weighting filter is applied before level
         calculation. If ``"Z"``, no frequency weighting is applied.
 
-        .. note::
-            The frequency weighting is applied using
-            `pyfar.dsp.filter.frequency_weighting_filter` with its
-            (standard-compliant) default parameters. If you need more control,
-            you can set this parameter to ``"Z"`` and apply the frequency
-            weighting filter yourself before calling this function.
+        The frequency weighting is applied using
+        :py:func:`~pyfar.dsp.filter.frequency_weighting_filter` with its
+        (standard-compliant) default parameters. If you need more control,
+        you can set this parameter to ``"Z"`` and apply the frequency
+        weighting filter yourself before calling this function.
 
     num_octave_band_fractions: int or ``None``
         Can be used to calculate the level in octave (``1``), third-octave
         (``3``), or other positive integer fractional octave bands.
         If ``None``, levels are calculated for the full-band signal.
 
-        .. note::
-            The fraction octave band filtering is applied using
-            `pyfar.dsp.filter.fractional_octave_bands` with its
-            (standard-compliant) default parameters. If you need more control,
-            you can set this parameter to ``None`` and apply the filter bank
-            yourself before calling this function.
+        The fraction octave band filtering is applied using
+        :py:func:`~pyfar.dsp.filter.fractional_octave_bands` with its
+        (standard-compliant) default parameters. If you need more control,
+        you can set this parameter to ``None`` and apply the filter bank
+        yourself before calling this function.
 
     reference_pressure: float
         The reference pressure to calculate levels relative to. The default
@@ -72,9 +71,9 @@ def equivalent_continuous_level(signal,
         The calculated levels of each channel in dB relative to the
         `reference_pressure`.
 
-        .. note::
-            If `num_octave_band_fractions` is not ``None``, the returned array
-            will have an additional first dimension for the individual bands.
+        `levels` has shape ``signal.cshape`` if `num_octave_band_fractions`
+        is ``None`` and ``(n_bands, signal.cshape)`` otherwise, where
+        `n_bands` denotes the number of (fractional) octave bands.
 
     References
     ----------
--- a/pyfar/level/standard_levels.py
+++ b/pyfar/level/standard_levels.py
@@ -1,0 +1,90 @@
+"""Level calculation functions according to IEC 61672-1."""
+
+from typing import Literal
+import numpy as np
+
+from ._utils import (
+    _check_signal_type,
+    _apply_frequency_weighting,
+    _apply_multi_band,
+    _energies_to_levels,
+)
+
+
+def equivalent_continuous_level(signal,
+                                frequency_weighting: Literal["A", "C", "Z"],
+                                num_octave_band_fractions: int | None = None,
+                                reference_pressure: float = 20e-6,
+                                ):
+    r"""Calculates the frequency-weighted equivalent continuous sound pressure
+    level (Leq) for each channel of a signal.
+
+    The levels are calculated according to IEC 61672-1 [#]_.
+    For instance, the A-weighted equivalent continuous level is calculated as:
+
+    .. math::
+        L_\text{Aeq} = 10 \log_{10} \left[ \frac{(1/N) \sum_{n=0}^{N-1}
+        p_{\text{A}}^2[n]} {p_0^2} \right] \text{ dB}
+
+    where :math:`N` is the number of samples in the signal and
+    :math:`p_0` is the `reference_pressure`.
+
+    Parameters
+    ----------
+    signal: Signal
+        The signal object to calculate the levels of
+
+    frequency_weighting: ``"A"``, ``"C"``, or ``"Z"``
+        The frequency weighting type. If ``"A"`` or ``"C"``, the corresponding
+        frequency weighting filter is applied before level
+        calculation. If ``"Z"``, no frequency weighting is applied.
+
+        .. note::
+            The frequency weighting is applied using
+            `pyfar.dsp.filter.frequency_weighting_filter` with its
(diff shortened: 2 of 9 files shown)


pyfar/level/standard_levels.py
```diff
@@ -0,0 +1,90 @@
...
+)
+
+
+def equivalent_continuous_level(signal,
+                                frequency_weighting: Literal["A", "C", "Z"],
+                                num_octave_band_fractions: int | None = None,
+                                reference_pressure: float = 20e-6,
+                                ):
+    r"""Calculates the frequency-weighted equivalent continuous sound pressure
+    level (Leq) for each channel of a signal.
```
f-brinkmann: We usualy start with imperative (?) and keep it short:

```suggestion
    r"""Calculate the frequency-weighted equivalent continuous sound pressure
    level (Leq).
```

pyfar/level/standard_levels.py
```diff
@@ -0,0 +1,90 @@
...
+
+    The levels are calculated according to IEC 61672-1 [#]_.
+    For instance, the A-weighted equivalent continuous level is calculated as:
+
+    .. math::
+        L_\text{Aeq} = 10 \log_{10} \left[ \frac{(1/N) \sum_{n=0}^{N-1}
+        p_{\text{A}}^2[n]} {p_0^2} \right] \text{ dB}
+
+    where :math:`N` is the number of samples in the signal and
```
f-brinkmann:
```suggestion
    where :math:`N` is the number of samples in the signal, :math:`p_\mathrm{A}` the A-weighted sound pressure, and
```

pyfar/level/standard_levels.py
```diff
@@ -0,0 +1,90 @@
...
+        The signal object to calculate the levels of
+
+    frequency_weighting: ``"A"``, ``"C"``, or ``"Z"``
+        The frequency weighting type. If ``"A"`` or ``"C"``, the corresponding
+        frequency weighting filter is applied before level
+        calculation. If ``"Z"``, no frequency weighting is applied.
+
+        .. note::
+            The frequency weighting is applied using
+            `pyfar.dsp.filter.frequency_weighting_filter` with its
```
f-brinkmann: This should link the function and omit the module name

```suggestion
            :py:func:`~pyfar.dsp.filter.frequency_weighting_filter` with its
```

pyfar/level/standard_levels.py
```diff
@@ -0,0 +1,90 @@
...
+            weighting filter yourself before calling this function.
+
+    num_octave_band_fractions: int or ``None``
+        Can be used to calculate the level in octave (``1``), third-octave
+        (``3``), or other positive integer fractional octave bands.
+        If ``None``, levels are calculated for the full-band signal.
+
+        .. note::
+            The fraction octave band filtering is applied using
+            `pyfar.dsp.filter.fractional_octave_bands` with its
```
f-brinkmann:
```suggestion
            :py:func:`~pyfar.dsp.filter.fractional_octave_bands` with its
```

pyfar/level/standard_levels.py
```diff
@@ -0,0 +1,90 @@
...
+    _apply_frequency_weighting,
+    _apply_multi_band,
+    _energies_to_levels,
+)
+
+
+def equivalent_continuous_level(signal,
+                                frequency_weighting: Literal["A", "C", "Z"],
+                                num_octave_band_fractions: int | None = None,
+                                reference_pressure: float = 20e-6,
```
f-brinkmann: I would suggest to use

```suggestion
                                reference_pressure: float = pyfar.constants.reference_sound_pressure,
```

and adapt the docstring below to state this, but still mention that this is 20e-5 Pa.

pyfar/level/standard_levels.py
```diff
@@ -0,0 +1,90 @@
...
+    ----------
+    signal: Signal
+        The signal object to calculate the levels of
+
+    frequency_weighting: ``"A"``, ``"C"``, or ``"Z"``
+        The frequency weighting type. If ``"A"`` or ``"C"``, the corresponding
+        frequency weighting filter is applied before level
+        calculation. If ``"Z"``, no frequency weighting is applied.
+
+        .. note::
```
f-brinkmann: Personally, I would not use note for this and the other cases below. Using it too often, makes it easier to overlook, when attention is required. Writing everything in 'plain' text would be fine for me.

pyfar/level/standard_levels.py
```diff
@@ -0,0 +1,90 @@
...
+
+    Returns
+    -------
+    levels: NDArray
+        The calculated levels of each channel in dB relative to the
+        `reference_pressure`.
+
+        .. note::
+            If `num_octave_band_fractions` is not ``None``, the returned array
+            will have an additional first dimension for the individual bands.
```
f-brinkmann: I would be more explicit, e.g.


```suggestion


`levels` has shape ``signal.cshape``  if `num_octave_band_fractions` is ``None`` and ``(n_bands, signal.cshape)`` otherwise, where `n_bands` denotes the number of (fractional) octave bands.
```

pyfar/level/_utils.py
```diff
@@ -0,0 +1,54 @@
...
+    """Throws if not a pyfar.Signal, otherwise returns the signal with
+    type hinting.
+    """
+    if not isinstance(signal, pf.Signal):
+        raise TypeError("'signal' parameter must be a pyfar.Signal")
+    signal: pf.Signal = signal
+    return signal
+
+
+def _apply_frequency_weighting(signal, frequency_weighting):
```
f-brinkmann, line 19: I would suggest to remove helper function and move the check for `frequency_weigting in ['A', 'C', 'Z']` to `pf.dsp.filter.frequency_weighting_filter`. We currently do not support `'Z'` there I think but it might be cleaner than this solution.

pyfar/level/_utils.py
```diff
@@ -0,0 +1,54 @@
...
+    if frequency_weighting == "Z":
+        return signal
+    elif frequency_weighting in ["A", "C"]:
+        return pf.dsp.filter.frequency_weighting_filter(
+            signal, frequency_weighting)
+    else:
+        raise ValueError("Frequency weighting must be 'A', 'C', or 'Z'")
+
+
+def _apply_multi_band(signal, num_octave_band_fractions: int | None):
```
f-brinkmann, line 32: As above I would remove this because I think the check for `num_octave_band_fractions` should be done in `pf.dsp.filter.fractional_octave_bands`, which also states that it must be an integer.

f-brinkmann (changes requested):
Thx for starting this. The general functionality looks good to me. Only had structural and docstring comments.

pyfar/level/_utils.py
jf-som: Should the frequency weighting filter function then have the option `"Z"` too even if it does nothing (for signal input) or returns a unit filter / dirac (for sampling rate input)? That would be strange in my opinion.

pyfar/level/_utils.py
jf-som: True, the positive integer checks are redundant. I would still rather keep the helper method though so I don't have to write identical if/else branches for the `None` case in every level function.

pyfar/level/standard_levels.py
jf-som: I like this! Implemented.

pyfar/level/standard_levels.py
jf-som: Fair, I made it regular text. Although I plan to move most of these hints to the module docstring at a later time, when all the functions with identical hints are implemented. So let's not be too strict with the parameters for now since they will be reworked later anyway

pyfar/level/standard_levels.py
jf-som: I always forget.... thanks and done

pyfar/level/standard_levels.py
jf-som: yup, done

pyfar/level/standard_levels.py
jf-som: Oh yeah, good catch. Done.

pyfar/level/standard_levels.py
jf-som: Have adapted the comment so channels get mentioned later, and made it imparative

pyfar/level/standard_levels.py
jf-som: Oh, I didn't realize that was already there as a constant. Changed it as requested

pyfar/level/_utils.py
f-brinkmann: True, did not think about that. Resolving this.

pyfar/level/_utils.py
jf-som: I removed the redundant checks now but kept the function itself for the None check.

f-brinkmann (approved):

pyfar/level/_utils.py
```diff
@@ -0,0 +1,47 @@
...
+between many of the standardized level functions.
+"""
+
+import numpy as np
+import pyfar as pf
+
+
+def _check_signal_type(signal):
+    """Throws if not a pyfar.Signal, otherwise returns the signal with
+    type hinting.
```
h-chmeruk: Maybe it's just my personal preference, but I would change it to `Throws/Raises a TypeError` to make the docstring sound more complete

pyfar/level/_utils.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    if not isinstance(signal, pf.Signal):
+        raise TypeError("'signal' parameter must be a pyfar.Signal")
+    signal: pf.Signal = signal
+    return signal
+
+
+def _apply_frequency_weighting(signal, frequency_weighting):
+    """Applies frequency weighting to the signal
+    or throws if the weighting name is invalid.
```
h-chmeruk: Same here: I would change it to `throws/raises a TypeError` to make the docstring sound more complete

tests/test_level_standard_levels.py
```diff
@@ -0,0 +1,13 @@
...
+
+# TODO thorough parameter testing can be added more effectively once
+# the more there are more functions using the same helper functions.
+
+def test_level_equivalent_continuous_level_reference_pressure():
+    s = pf.signals.sine(1000, 22050)
+    levels = pf.level.equivalent_continuous_level(
+        s, "Z", reference_pressure=2e-5)
+    # 94 dB is 1 Pa, -3.01 dB is the crest factor of sine singals
+    assert np.isclose(levels, 94 - 3.01, atol=0.1)
```
h-chmeruk: Regarding this test: Maybe I missed something, but why don't we also test the function with different weightings? I could imagine doing something like this:

```
@pytest.mark.parametrize(("weighting", "expected"), [("Z", 90.99),
                                ("C", 90.6), ("A", 71.83)])
def test_level_equivalent_continuous_level_reference_pressure(weighting, expected):
    s = pf.signals.sine(100, 22050)
    levels = pf.level.equivalent_continuous_level(
        s, weighting, reference_pressure=2e-5)
    # 94 dB is 1 Pa, -3.01 dB is the crest factor of sine singals
    assert np.isclose(levels, expected, atol=0.1)
```

Or, of course, with other frequencies of the sine signal.

h-chmeruk (commented):
Thank you for implementing this and breaking the code down into smaller PRs. This makes it much easier to review.

I really like the implementation overall, and I appreciate the detailed docstrings, so I just have some suggestions and a question regarding the test functions.

I understand that as more functions are added, more tests can be expanded and added as well, but as I see it, you could already add tests in this PR to verify that errors are raised correctly — for example, in the helper function `_check_signal_type() ` when incorrect values are passed as a signal to the `equivalent_continuous_level()` function. Or did I overlook something?

pyfar/level/_utils.py
jf-som: Good point, did it.

pyfar/level/_utils.py
jf-som: I now mention that it raises a ValueErrror

jf-som:
Thanks @h-chmeruk for the review. I can see that my reasons for leaving the tests for later weren't very clear. So I decided to implement the tests I was planning now rather than later after all, even if they might look like a bit much for now with just a single function being tested.
My idea is that by using the pytest parametrize decorator on the level functions being tested, we can avoid hundreds of lines of near-identical tests for the different level functions later. So starting with the next level function that will be added, tests can be added and reviewed much more easily and quickly than in this PR.

tests/test_level_common_parameters.py
```diff
@@ -0,0 +1,158 @@
...
+    lambda s: pf.level.equivalent_continuous_level(s, "Z"),
+    # other level functions go here once implemented
+])
+def test_level_common_signal_parameter(function):
+    # should not raise an error
+    function(pf.signals.sine(1000, 22050))
+
+    # all these should raise errors
+    with pytest.raises(TypeError):
+        function("not a signal")
```
h-chmeruk: I would recommend using parameterization for the `signal` as well. However, that would mean that every function you add as a `function` parameter would be tested with the same input signals. 

```suggestion
@pytest.mark.parametrize("function", [
    lambda s: pf.level.equivalent_continuous_level(s, "Z"),
    # other level functions go here once implemented
])
@pytest.mark.parametrize(("signal"), ["not a signal", 123, None, np.array([1, 2, 3]),
                                    pf.TimeData(np.array([1, 2, 3]), [1, 2, 3]),
                                    pf.FrequencyData(np.array([1, 2, 3]), [1, 2, 3])])
def test_level_common_signal_parameter(function, signal):
    # should not raise an error
    function(pf.signals.sine(1000, 22050))

    # all these should raise errors
    with pytest.raises(TypeError):
```

Or, if future functions in the `function` parameter might fail for some data types but not for others in the list of signals above, you could assign a separate `signal_list` to each `function`. However, this is less practical, since pytest would then interpret this as one test per function and would not distinguish between different `signal` inputs. 

```suggestion
@pytest.mark.parametrize(("function", "signal_list"), [
    (lambda s: pf.level.equivalent_continuous_level(s, "Z"),
     ("not a signal", 123, None, np.array([1, 2, 3]),
      pf.TimeData(np.array([1, 2, 3]), [1, 2, 3]),
      pf.FrequencyData(np.array([1, 2, 3]), [1, 2, 3]))),
    # other level functions go here once implemented
])
def test_level_common_signal_parameter(function, signal_list):
    # should not raise an error
    function(pf.signals.sine(1000, 22050))

    # all these should raise errors
    for signal in signal_list:
        with pytest.raises(TypeError):
            function(signal)
```

tests/test_level_common_parameters.py
```diff
@@ -0,0 +1,158 @@
...
+### signal parameter tests ###
+
+
+@pytest.mark.parametrize("function", [
+    lambda s: pf.level.equivalent_continuous_level(s, "Z"),
+    # other level functions go here once implemented
+])
+def test_level_common_signal_parameter(function):
+    # should not raise an error
+    function(pf.signals.sine(1000, 22050))
```
h-chmeruk: Do I understand correctly that all future functions will use the helper function `_check_signal_type()` for input validation of the `signal` parameter? If so, then it isn't necessary to test every `function` with the same parameters. It would be sufficient to test the helper function `_check_signal_type()` with all of these input signals

tests/test_level_common_parameters.py
```diff
@@ -0,0 +1,158 @@
...
+    """Test that an invalid number of octave band fractions raises an error."""
+    s = pf.signals.sine(1000, 22050)
+    with pytest.raises(TypeError, match="integer"):
+        function(s, "3")
+    with pytest.raises(TypeError, match="integer"):
+        function(s, 2.5)
+    with pytest.raises(ValueError, match="positive"):
+        function(s, -1)
+    with pytest.raises(ValueError, match="positive"):
+        function(s, 0)
```
h-chmeruk: The number of bands can also be parameterized

```suggestion
@pytest.mark.parametrize("function", FUNCTION_WRAPPERS_BAND_FRACTIONS)
@pytest.mark.parametrize("num_bands", ["3", 2.5, -1, 0])
def test_level_common_num_octave_band_fractions_errors(function, num_bands):
    """Test that an invalid number of octave band fractions raises an error."""
    s = pf.signals.sine(1000, 22050)
    with pytest.raises(TypeError, match="integer"):
        function(s, num_bands)
```

tests/test_level_common_parameters.py
```diff
@@ -0,0 +1,158 @@
...
+    band_filtered = function(signal, num_fractions)
+
+    # new first dimension with the same other dimensions
+    assert band_filtered.shape == (num_bands, *full_band.shape)
+
+
+@pytest.mark.parametrize("function", FUNCTION_WRAPPERS_BAND_FRACTIONS)
+def test_level_common_num_octave_band_fractions_errors(function):
+    """Test that an invalid number of octave band fractions raises an error."""
+    s = pf.signals.sine(1000, 22050)
```
h-chmeruk, line 122: As far as I can see, all the checks and the logic for fractional octave band filtering are performed in the function `pf.dsp.filter.fractional_octave_bands()`, which is used in the helper function `_apply_multi_band()`. Therefore, I would suggest testing the helper function directly as well, rather than the higher-level functions that use it

tests/test_level_common_parameters.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+@pytest.mark.parametrize("function", FUNCTION_WRAPPERS_FREQ_WEIGHTING)
+def test_level_common_freq_weighting_errors(function):
+    """Test that an invalid frequency weighting raises an error."""
+    s = pf.signals.sine(1000, 22050)
+    with pytest.raises(ValueError, match="Frequency weighting"):
+        function(s, "X")
+    with pytest.raises(ValueError, match="Frequency weighting"):
+        function(s, None)
```
h-chmeruk: Same here: If the checks for frequency weighting in the future functions are to be performed using the helper function `_apply_frequency_weighting()`, there is no need to test this ten times in different functions. You could test the helper function itself

h-chmeruk (commented):
Thank you very much for adding the tests! The coverage is already very good, and the comments are very helpful.

I have just one major suggestion: Most of the test functions check the functionality of the utility functions, which, in my opinion, could or rather should be done directly in a separate file for testing the `level/_utils` functions. The new file could, for example, be named “test_level__utils.py.” In the `test_level_common_parameters.py` file, I would only include test functions to verify that the utility functions were actually called correctly, as well as test functions for the cases and checks that are not performed via helper functions. 

The test to verify whether the helper functions were called can be implemented using Mocking, for example like this: 

```

@pytest.mark.parametrize("function", [
   pf.level.equivalent_continuous_level,
    # other level functions go here once implemented
])
@pytest.mark.parametrize("helper_function", [
    "pyfar.level.standard_levels._check_signal_type",
    "pyfar.level.standard_levels._apply_frequency_weighting",
    "pyfar.level.standard_levels._apply_multi_band",
    "pyfar.level.standard_levels._energies_to_levels",
])

def test_call_helper_functions(function, helper_function):

    with patch(f"{helper_function}") as mock_validation:
        function(pf.signals.sine(1000, 22050), "Z")
    mock_validation.assert_called()
```

Or maybe I'm wrong, and it's not an efficient approach or isn't good enough in terms of maintainability. So I'd be happy to hear what others think. :)

tests/test_level_common_parameters.py
jf-som: I moved the different signal types into a parametrize decorator now.

tests/test_level_common_parameters.py
jf-som: I moved it to a parametrize decorator, although the error type and match string had to be added to the decorator too.

tests/test_level_standard_levels.py
jf-som: Forgot to reply here the other day. Now that we have tests, this no longer applies.

jf-som:
Thanks again @h-chmeruk . I did think about testing the helper functions instead, although I didn't know before today that python had reflection such as the mock function you mentioned, so thanks for sharing the idea!

If fully testing every level function becomes too slow, we could change it to tests only for the helper functions and use mock to ensure the level functions actually use these helpers as you suggest. But personally, I'm in favor of testing the level functions as "black boxes" (as I do now). The mock way only checks whether the helpers were used, not whether they were used correctly, and it also limits potential changes to the functions later too.

This might be more of a testing philosophy topic though haha. I prefer my black box approach but could be convinced to use your proposal.
Maybe this is a discussion we can save for later (that is, a different PR that is also part of #949), when more level functions are implemented and we see how long the tests actually take? Perhaps this won't be an issue at all.

tests/test_level_common_parameters.py
jf-som: please see my comment below

tests/test_level_common_parameters.py
jf-som: please see my comment below

tests/test_level_common_parameters.py
jf-som: please see my comment below

h-chmeruk:
@jf-som, you are welcome! And it's totally fine if you want to stick with your current implementation. Everyone has their own preferences when it comes to implementation :) 
I don't think it was wrong or not good enough, I just thought we might be able to make it even more efficient. But of course, you know your code better and are much better able to evaluate what makes sense and what doesn't!

Personally, I think we can leave it as it is, and only if the duration of the tests becomes unusually long, we could perhaps discuss the suggestion again.

f-brinkmann:
@jf-som the missing entry in history prevents the other test from starting. Due to this, I would suggest to already start a history entry for the feature here. It could be modified or extended in the upcoming related PRs.

@h-chmeruk have all your comments been addressed? If yes you could approve already - or after the above has been addressed.

jf-som:
I added the history entry now. If there is nothing major or Leq-specific to discuss left, I'd be happy to merge this PR soon, since all the other parts of the level submodule depend on this PR.

h-chmeruk (approved):

PR #951, opened 2026-07-15
Level/exposure

jf-som:
Part of #949. This PR adds the `exposure_level()` function to the level module.

Potential points of discussion:
- Do you like my approach for the `duration` parameter? From what I understand, the exposure level is mostly used to calculate the daily sound expose, i.e. over an 8h work day. But it's unlikely pyfar users will have 8h long continuous audio recordings (which would be several GB) available. Therefore I added the parameter to let users extrapolate the average sound energy of the provided signal over a longer period of time, which should be more useful. Let me know what you think or different ideas to deal with this.
- I used `reference_pressure`, but the norm technically defines a reference *power*, though they define it as (2e-5)^2 (the reference pressure, squared). To keep the function consistent with the other level functions, I prefer using the pressure as parameter.

Update after weekly meeting: 
We decided to keep the duration parameter this way, and also keep the `reference_pressure` in favor of consistency, but with clear explanations regarding how the reference pressure relates to the reference energy (E0, which is p0^2 *T0) in the docstring.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -9,6 +9,7 @@ Added
 ^^^^^
 - A new submodule `pyfar.level` with various functions to compute and work with levels (PR #949), consisting of:
     - `level.equivalent_continuous_level` for the Leq (PR #948)
+    - `level.exposure_level` for the L_E (PR #951)
 
 Fixed
 ^^^^^
--- a/pyfar/level/standard_levels.py
+++ b/pyfar/level/standard_levels.py
@@ -163,6 +163,8 @@ def exposure_level(
         signal, frequency_weighting, None, reference_pressure)
 
     duration = signal.signal_length if duration is None else duration
+    if not isinstance(duration, (int, float, np.number)):
+        raise TypeError("Duration must be a number.")
     if duration <= 0:
         raise ValueError("Duration must be a positive number.")
     duration_term = 10 * np.log10(duration)
--- a/pyfar/level/standard_levels.py
+++ b/pyfar/level/standard_levels.py
@@ -133,7 +133,7 @@ def exposure_level(
         set to the length of the signal in seconds to calculate the exposure
         of just the signal. You can specify a different duration to extrapolate
         the exposure level to a longer (or shorter) time period assuming the
-        same average sound pressure remains constant. 
+        same average sound pressure remains constant.
         The duration must be a positive number.
 
     reference_pressure: float
--- a/pyfar/level/standard_levels.py
+++ b/pyfar/level/standard_levels.py
@@ -89,3 +89,21 @@ def equivalent_continuous_level(
     mean_energy_per_channel = np.mean(signal.time**2, axis=-1)
     levels = _energies_to_levels(mean_energy_per_channel, reference_pressure)
     return levels
+
+
+def exposure_level(
+        signal,
+        frequency_weighting: Literal["A", "C", "Z"],
+        duration: float | None = None,
+        reference_pressure: float = pf.constants.reference_sound_pressure,
+):
+    signal = _check_signal_type(signal)
+    eq_level = equivalent_continuous_level(
+        signal, frequency_weighting, None, reference_pressure)
+
+    duration = signal.signal_length if duration is None else duration
+    if duration <= 0:
+        raise ValueError("Duration must be a positive number.")
+    duration_term = 10 * np.log10(duration)
+
+    return eq_level + duration_term
--- a/pyfar/level/standard_levels.py
+++ b/pyfar/level/standard_levels.py
@@ -115,7 +115,7 @@ def exposure_level(
     Parameters
     ----------
     signal: Signal
-        The signal object to calculate the levels of
+        The signal object to calculate the levels of.
 
     frequency_weighting: ``"A"``, ``"C"``, or ``"Z"``
         The frequency weighting type. If ``"A"`` or ``"C"``, the corresponding
--- a/pyfar/level/standard_levels.py
+++ b/pyfar/level/standard_levels.py
@@ -112,6 +112,16 @@ def exposure_level(
     :math:`p_0` is the `reference_pressure`,
     and :math:`T` is the duration of the signal or the `duration` specified.
 
+    .. note::
+        The standard defines the sound exposure level relative to
+        a reference *energy* :math:`E_0 = p_0^2 T_0
+        = 20\,\mu\text{Pa}^2 \cdot 1 \text{ s}
+        = 400 \cdot 10^{-12} \text{ J}`.
+        However, this function uses a reference *pressure* instead to remain
+        consistent with the other level functions.
+        You can obtain the `reference_pressure` by taking the square root of of
+        the desired reference energy.
+
     Parameters
     ----------
     signal: Signal
--- a/pyfar/level/standard_levels.py
+++ b/pyfar/level/standard_levels.py
@@ -97,6 +97,67 @@ def exposure_level(
         duration: float | None = None,
         reference_pressure: float = pf.constants.reference_sound_pressure,
 ):
+    r"""Calculate the frequency-weighted sound exposure level.
+
+    The levels are calculated per channel and according to IEC 61672-1 [#]_.
+    For instance, the A-weighted sound exposure level is calculated as:
+
+    .. math::
+        L_{\text{A}E,T} = 10 \log_{10} \left[ \frac{(1/N) \sum_{n=0}^{N-1}
+        p_{\text{A}}^2[n] T} {p_0^2} \right] \text{ dB}
+        = L_\text{Aeq} + 10 \log_{10}(T) \text{ dB}
+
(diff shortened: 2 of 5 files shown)


pyfar/level/standard_levels.py
```diff
@@ -89,3 +89,82 @@ def equivalent_continuous_level(
...
+
+    .. math::
+        L_{\text{A}E,T} = 10 \log_{10} \left[ \frac{(1/N) \sum_{n=0}^{N-1}
+        p_{\text{A}}^2[n] T} {p_0^2} \right] \text{ dB}
+        = L_\text{Aeq} + 10 \log_{10}(T) \text{ dB}
+
+    where :math:`N` is the number of samples in the signal,
+    :math:`p_\mathrm{A}` the A-weighted sound pressure at index :math:`n`,
+    :math:`p_0` is the `reference_pressure`,
+    and :math:`T` is the duration of the signal or the `duration` specified.
```
h-chmeruk: I'm not quite sure about this, but if the formula is the same for every weighting type, why don't we use more general variable names in the formula so that no one gets confused and thinks it might work differently with a weighting other than A?

```suggestion
    The levels are calculated per channel and according to IEC 61672-1 [#]_.
    A weighted sound exposure level is calculated as:

    .. math::
        L_{\text{w}E,T} = 10 \log_{10} \left[ \frac{(1/N) \sum_{n=0}^{N-1}
        p_{\text{w}}^2[n] T} {p_0^2} \right] \text{ dB}
        = L_\text{w,eq} + 10 \log_{10}(T) \text{ dB}

    where :math:`N` is the number of samples in the signal,
    :math:`p_\mathrm{w}`is the frequency-weighted sound pressure (A-, C-, or Z-weighted, per frequency_weighting) at index :math:`n`,
    :math:`p_0` is the `reference_pressure`,
    and :math:`T` is the duration of the signal or the `duration` specified.
```

(The suggestion may look completely wrong, I haven't tested it!! But I wanted to show what I mean by more general variable names using an example)

h-chmeruk (commented):
Thank you for implementing it!
I just have a question/suggestion regarding the docstring for the `exposure_level()` function, and I don't think there's a test checking whether the corresponding errors are raised if the `duration` parameter is set to an invalid value. It would be great if you could add such a test as well :)

pyfar/level/standard_levels.py
jf-som: I was thinking the same, but for some reason the norm defines it like this as well, I try to make the level module as close to the norm as possible.
It should be fine though, especially given that even the norm committee seems to think the A-weighted example is enough for all weightings.

jf-som:
Thanks for the review, @h-chmeruk !
I added the tests as you suggested, I indeed forgot about them so far. I also added a type check for the duration that was missing to get more consistent and clear error messages.

Also, as discussed in the last meeting, I added a note explaining why this function uses reference pressure rather than reference energy, and how the two are defined and related.

HISTORY.rst
```diff
@@ -9,6 +9,7 @@ Added
 ^^^^^
 - A new submodule `pyfar.level` with various functions to compute and work with levels (PR #949), consisting of:
     - `level.equivalent_continuous_level` for the Leq (PR #948)
+    - `level.exposure_level` for the L_E (PR #951)
```
h-chmeruk: I'm not very familiar with the abbreviations in this area, so I'm assuming that L_E simply stands for “Level_Exposure”? Maybe you could make the description a little clearer, but that's not necessary, it is just a suggestion of mine :)

h-chmeruk (approved):
Thank you for implementing the changes! I think it looks great. I just had one small suggestion regarding the entry in the HISTORY.rst file, see the comment below

HISTORY.rst
jf-som: Uhh, I think I wanted to make it consistent with the line above. But with the underscore that isn't really the case, and the exposure level is less known than the Leq anyway, so I just wrote it out instead now.

pyfar/level/standard_levels.py
```diff
@@ -89,3 +89,94 @@ def equivalent_continuous_level(
...
+        reference_pressure: float = pf.constants.reference_sound_pressure,
+):
+    r"""Calculate the frequency-weighted sound exposure level.
+
+    The levels are calculated per channel and according to IEC 61672-1 [#]_.
+    For instance, the A-weighted sound exposure level is calculated as:
+
+    .. math::
+        L_{\text{A}E,T} = 10 \log_{10} \left[ \frac{(1/N) \sum_{n=0}^{N-1}
+        p_{\text{A}}^2[n] T} {p_0^2} \right] \text{ dB}
```
f-brinkmann: Maybe make more clear that T is outside the sum by moving it to the front and writing T/N? Mathematically the parentheses around 1/N would not be required and for me are not really required to structure the equation.

pyfar/level/standard_levels.py
```diff
@@ -89,3 +89,94 @@ def equivalent_continuous_level(
...
+    """
+    signal = _check_signal_type(signal)
+    eq_level = equivalent_continuous_level(
+        signal, frequency_weighting, None, reference_pressure)
+
+    duration = signal.signal_length if duration is None else duration
+    if not isinstance(duration, (int, float, np.number)):
+        raise TypeError("Duration must be a number.")
+    if duration <= 0:
+        raise ValueError("Duration must be a positive number.")
```
f-brinkmann, line 179: I would move these checks up to avoid wasting compute on equivalent_continuous_level in case these checks fail.

pyfar/level/standard_levels.py
```diff
@@ -89,3 +89,94 @@ def equivalent_continuous_level(
...
+        (standard-compliant) default parameters. If you need more control,
+        you can set this parameter to ``"Z"`` and apply the frequency
+        weighting filter yourself before calling this function.
+
+    duration: float or ``None``
+        The duration of the signal in seconds. If ``None``, the duration is
+        set to the length of the signal in seconds to calculate the exposure
+        of just the signal. You can specify a different duration to extrapolate
+        the exposure level to a longer (or shorter) time period assuming the
+        same average sound pressure remains constant.
```
f-brinkmann: Maybe a little more explicit

```suggestion
        the exposure level to a longer (or shorter) time period. This extrapolation is
        valid if the equivalent sound pressure level :math:`L_\text{Aeq}` is the same
        for the longer (or shorter) signal.
```

f-brinkmann (changes requested):
Thanks for implementing. I only had small comments.

pyfar/level/standard_levels.py
jf-som: Thanks to your comment I noticed that the denominator missed the T_0 factor entirely so far... oops. Fixed it now.
Regarding the T, I had it there inside the sum to match the denominator (since it shouldn't matter for factors to be inside or outside a sum), but I like your idea of combining T/N, since both are necessary to convert the integral over T from the norm to our discrete sum.
I refactored the formula to your suggestions.

pyfar/level/standard_levels.py
jf-som: True! I now arranged to lines to have all checks first, and then all computations at the end.

pyfar/level/standard_levels.py
jf-som: Adapted something close to your suggestion.

pyfar/level/standard_levels.py
```diff
@@ -89,3 +89,96 @@ def equivalent_continuous_level(
...
+    .. math::
+        L_{\text{A}E,T} = 10 \log_{10} \left[ \frac{T/N \sum_{n=0}^{N-1}
+        p_{\text{A}}^2[n]} {p_0^2 T_0} \right] \text{ dB}
+        = L_\text{Aeq} + 10 \log_{10}(T) \text{ dB}
+
+    where :math:`N` is the number of samples in the signal,
+    :math:`p_\mathrm{A}` the A-weighted sound pressure at index :math:`n`,
+    :math:`p_0` is the `reference_pressure`,
+    :math:`T_0` is the reference duration of 1 second, and
+    :math:`T` is the duration of the signal or the `duration` specified.
```
f-brinkmann: This is confusing but admittadly hard to describe. How about simplifing a bit
```suggestion
    :math:`T` is assumed exposure `duration` to the signal (see below).
```

f-brinkmann (changes requested):
thanks for the changes. I found one more part in the docs, that was a little confusing to me and suggested a possible fix. Otherwise approved.

pyfar/level/standard_levels.py
jf-som: I agree, though your version was also a bit confusing to me. I made it even shorter and simpler and added the "see below" like you suggest.

f-brinkmann (approved):

jf-som:
Since this PR has two approvals and it only merges into another PR branch from me, I'm merging this myself.
Of course for the final merge into develop at the end I will request someone of maintainer status for the merge :)

PR #952, opened 2026-07-15
Level/sliding

jf-som:
Part of #949. 

This PR adds a moving average utility function and uses it to add the `sliding_equivalent_continuous_level()` function according to the standard.

### DIFF
--- a/pyfar/level/_utils.py
+++ b/pyfar/level/_utils.py
@@ -4,6 +4,7 @@ between many of the standardized level functions.
 
 import numpy as np
 import pyfar as pf
+import scipy.ndimage
 
 
 def _check_signal_type(signal):
@@ -45,3 +46,51 @@ def _energies_to_levels(energies: np.ndarray, reference_pressure: float):
     if reference_pressure == 0:
         raise ValueError("Reference pressure must not be zero.")
     return 10 * np.log10(energies / reference_pressure**2)
+
+
+def _moving_average(array: np.ndarray,
+                    window_size: int,
+                    axis: int = -1,
+                    cyclic: bool = False,
+                    center_window: bool = False,
+                    ):
+    """Efficiently calculates the moving average of an array along
+    an axis.
+    If `cyclic` is ``True``, the array is treated as if it were periodic
+    (like a cyclic convolution). If ``False`` (default), edges are
+    zero-padded, leading to fade-like boundary effects.
+    If `center_window` is ``True``, the window is centered on the current
+    sample. If ``False`` (default), the window is causal, meaning it only
+    considers past values, same as the `convolve` functions in `numpy`,
+    `scipy.signal`, and `pyfar.dsp`.
+    """
+    if (not isinstance(window_size, int)):
+        raise TypeError("Window size must be an integer.")
+    if window_size < 1:
+        raise ValueError("Window size must be a positive integer.")
+
+    # If the input are integers, the outputs are too (rounded), which
+    # we never want in floating point audio.
+    if np.issubdtype(array.dtype, np.integer):
+        array = array.astype(np.float64)
+
+    if center_window:
+        origin = 0  # 0 means the window is centered
+    else:
+        # For a causal window, the window needs to be shifted back
+        if window_size % 2 == 1:
+            origin = window_size // 2
+        else:
+            # since scipy.ndimage is designed for images, which normally
+            # use odd window sizes, even sizes are finnicky...
+            origin = window_size // 2 - 1
+
+    mode = "wrap" if cyclic else "constant"
+
+    # This is effectively the same as a convolution with a rectangular window,
+    # but more efficient, especially for long windows.
+    # Unlike convolve, this function has a built-in "wrap" mode, but no
+    # "full" mode
+    result = scipy.ndimage.uniform_filter1d(
+        array, size=window_size, axis=axis, mode=mode, origin=origin)
+    return result
--- a/pyfar/level/_utils.py
+++ b/pyfar/level/_utils.py
@@ -40,11 +40,19 @@ def _apply_multi_band(signal, num_octave_band_fractions: int | None):
         signal, num_octave_band_fractions)
 
 
-def _energies_to_levels(energies: np.ndarray, reference_pressure: float):
+def _energies_to_levels(energies: np.ndarray,
+                        reference_pressure: float,
+                        replace_zeros: bool = False):
     """Converts energies to levels in dB relative to the reference pressure.
+    Raises an error if the reference pressure is zero.
+    If `replace_zeros` is ``True``, zeros in the energies are replaced with
+    the float epsilon to avoid ``-inf`` values and resulting numpy warnings.
     """
     if reference_pressure == 0:
         raise ValueError("Reference pressure must not be zero.")
+    if replace_zeros:
+        energies = np.where(
+            energies == 0, np.finfo(energies.dtype).eps, energies)
     return 10 * np.log10(energies / reference_pressure**2)
 
 
--- a/pyfar/level/standard_levels.py
+++ b/pyfar/level/standard_levels.py
@@ -182,6 +182,19 @@ def sliding_equivalent_continuous_level(
     .. [#] International Electrotechnical Commission,
         "IEC 61672-1:2013 - Electroacoustics - Sound level meters - Part 1:
         Specifications", IEC, 2013.
+
+    Examples
+    --------
+    Obtain the equivalent continuous level in 2-second intervals in dbFS(A).
+
+    >>> import pyfar as pf
+    >>> fs = 48000
+    >>> interval = 2
+    >>> signal = pf.signals.files.guitar()
+    >>> sliding_levels = pf.level.sliding_equivalent_continuous_level(
+    >>>         signal, "A", window_duration=interval, reference_pressure=1)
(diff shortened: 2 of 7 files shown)


tests/test_level__utils.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    signal = pf.Signal(signal_data, 48000)
+    filt = pf.Signal(np.ones(window_size) / window_size, 48000)
+    mode = "cyclic" if cyclic else "cut"
+    expected = pf.dsp.convolve(signal, filt, mode=mode).time
+
+    result = pf.level._utils._moving_average(
+        signal_data, window_size, -1, cyclic, False)
+    result = np.atleast_2d(result)
+    print(result)
```
h-chmeruk:
```suggestion
```
You don't need that in test functions, right?

pyfar/level/_utils.py
```diff
@@ -39,9 +40,65 @@ def _apply_multi_band(signal, num_octave_band_fractions: int | None):
...
+            energies == 0, np.finfo(energies.dtype).eps, energies)
     return 10 * np.log10(energies / reference_pressure**2)
+
+
+def _moving_average(array: np.ndarray,
+                    window_size: int,
+                    axis: int = -1,
+                    cyclic: bool = False,
+                    center_window: bool = False,
+                    ):
```
h-chmeruk, line 64: If I see this correctly, there is no check anywhere to verify whether the arguments for `center_window`, `cyclic`, and `axis` have the correct data type. There is also no such check in the upper-level function `sliding_equivalent_continuous_level`. But maybe you may not need to implement any checks for `axis` at all, since its value is set to the default in this function and is not modified in the `sliding_equivalent_continuous_level` function... 
So it’s up to you, I'm just pointing this out in case it was overlooked

pyfar/level/standard_levels.py
```diff
@@ -89,3 +90,120 @@ def equivalent_continuous_level(
...
+    Obtain the equivalent continuous level in 2-second intervals in dbFS(A).
+
+    >>> import pyfar as pf
+    >>> fs = 48000
+    >>> interval = 2
+    >>> signal = pf.signals.files.guitar()
+    >>> sliding_levels = pf.level.sliding_equivalent_continuous_level(
+    >>>         signal, "A", window_duration=interval, reference_pressure=1)
+    >>> interval_levels = sliding_levels[0][fs*interval::fs*interval]
+    >>> print(interval_levels) # [-32.24998448 -31.61894792 -28.19029974]
```
h-chmeruk: I think the example would be even more helpful if it included a graphical representation. You could plot these two graphs (guitar data and the `interval_levels`) side by side so that people can see how you can easily get some nice dB values from such a chaotic signal. In theory, this should be clear from the description of the function, but I always find plots very helpful because they usually clarify things that were unclear from the text.
```suggestion
    >>> import pyfar as pf
    >>> fs = 48000
    >>> interval = 2
    >>> signal = pf.signals.files.guitar()
    >>> sliding_levels = pf.level.sliding_equivalent_continuous_level(
    >>>         signal, "A", window_duration=interval, reference_pressure=1)
    >>> interval_levels = sliding_levels[0][fs*interval::fs*interval]
    >>> _, axs = plt.subplots(2, 1)
    >>> pf.plot.time(signal, ax=axs[0], unit='s')
    >>> pf.plot.time(interval_levels, ax=axs[1], unit='s')
```
(I haven't tested the suggestion, but I suppose you cannot plot interval_levels like that, because it is not a `Signal` or `TimeData` object, so you would need to separately specify the x-axis for this plot with interval_levels)

h-chmeruk (commented):
Thanks for implementing this! It looks good to me and I haven’t noticed anything that definitely needs to be changed, so all my comments are more like optional suggestions :)

tests/test_level__utils.py
jf-som: Oh, I missed that one, thanks

pyfar/level/standard_levels.py
jf-som: Fair point. I added a plot to this example and also added a new example that also shows the other parameters' effects graphically.

pyfar/level/_utils.py
jf-som: Right, I added checks for the two booleans now, but leave axis as it is, since it's not exposed to the user and it's used the same way as everywhere in numpy.

h-chmeruk (approved):
Thank you for adding and extending the examples! It looks good to me :)

pyfar/level/standard_levels.py
```diff
@@ -89,3 +90,177 @@ def equivalent_continuous_level(
...
+        (standard-compliant) default parameters. If you need more control,
+        you can set this parameter to ``None`` and apply the filter bank
+        yourself before calling this function.
+
+    window_duration: float
+        The duration of the sliding window in seconds, which determines
+        the time over which the energy is averaged at each sample.
+        The default value is ``1``.
+
+    cyclic: bool
```
f-brinkmann, line 237: I assume only `False` is standard conform, right? If yes, this could be mentioned.

pyfar/level/standard_levels.py
```diff
@@ -89,3 +90,177 @@ def equivalent_continuous_level(
...
+        If ``True``, the signal is treated as if it were periodic (like a
+        cyclic convolution). If ``False``, edges are zero-padded,
+        leading to fade-like boundary effects. The default is ``False``.
+
+    center_window: bool
```
f-brinkmann, line 242: Are both possibilities standard conform? If not, this would be nice to mention.

pyfar/level/standard_levels.py
```diff
@@ -89,3 +90,177 @@ def equivalent_continuous_level(
...
+
+    References
+    ----------
+    .. [#] International Electrotechnical Commission,
+        "IEC 61672-1:2013 - Electroacoustics - Sound level meters - Part 1:
+        Specifications", IEC, 2013.
+
+    Examples
+    --------
+    Obtain the equivalent continuous level in 2-second intervals in dbFS(A).
```
f-brinkmann: I would prefer focusing on the sliging level here and not compute and plot `interval_levels` to make the example simpler. What was your idea for including `interval_levels`?

tests/test_level_standard_levels.py
```diff
@@ -16,3 +16,21 @@ def test_level_equivalent_continuous_level_known_value():
...
+def test_level_sliding_equivalent_continuous_level_known_value():
+    # phase prevents the first sample from being exactly zero, which would
+    # cause a division by zero in the level calculation
+    s = pf.signals.sine(1000, 44100, phase=0.01, sampling_rate=44100)
+    levels = pf.level.sliding_equivalent_continuous_level(
+        s, "Z", None, 1, False, False, 2e-5)
+    # 94 dB is 1 Pa, -3.01 dB is the crest factor of sine signals
+    # with 1 second window size, the value at 1 second should be
+    # the same as the equivalent continuous level of the full signal
+    assert np.isclose(levels[0][-1], 94 - 3.01, atol=0.1)
```
f-brinkmann: `atol` is rather high. Could it be reduced by using `20 * np.log10(np.sqrt(2))` instead of `3.01`?

f-brinkmann (changes requested):
Thanks for implementing. Code looks good to me. Only had minor comments on docs and tests.

pyfar/level/standard_levels.py
jf-som: As far as I know, a sliding Leq is not standardized. At least not in DIN EN 61672-1 where all the others are defined, and I don't know where else it would be.
I think this one is just a convenience thing that many digital sound pressure meters offer, because it's convenient. 
So unless I'm missing a standard, "standard-conform" does not really apply here. But I could add a half-sentence that setting it to false would match the behavior of a live measurement using a real (analog) microphone maybe. What do you think?

pyfar/level/standard_levels.py
jf-som: Same as `cyclic`above.

pyfar/level/standard_levels.py
jf-som: My idea was to mimic the 5s-interval level (Takt-Mittelungspegel) that are part of some workplace/neighborhood noise standards (not DIN EN 61672 however).
I agree that this example does too much right now though. I could remove the intervals or move it to its own example. Personally I'm in favor or removing it, since thinking about it again, I don't really see a use-case where users would use this function for some interval level calculation.

tests/test_level_standard_levels.py
jf-som: I now defined exact variables for one pascal and the crest factor at the top of the file to use in these tests, and lowered the tolerance now.

pyfar/level/standard_levels.py
f-brinkmann: Ah, did not think of that. In that case, I'm fine with the current version.

pyfar/level/standard_levels.py
f-brinkmann: +1 for removing the intervals

f-brinkmann (changes requested):
Thanks for the responses. Will approve, after the plot has been updated.

pyfar/level/standard_levels.py
jf-som: Alright, did it :)

f-brinkmann (approved):

PR #954, opened 2026-07-15
Level/time weighting

jf-som:
Part of #949.

This PR adds the function `time_weighted_level()` and its public utility `time_weighted_pressure()`.

### DIFF
--- a/pyfar/level/standard_levels.py
+++ b/pyfar/level/standard_levels.py
@@ -4,6 +4,7 @@ from typing import Literal
 import numpy as np
 import pyfar as pf
 
+from .utils import time_weighted_pressure
 from ._utils import (
     _check_signal_type,
     _apply_frequency_weighting,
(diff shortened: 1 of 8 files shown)


tests/test_level_standard_levels.py
```diff
@@ -6,6 +6,8 @@
 against known values or other tests that are specific to a single function.
 """
 
+import pytest
+
```
h-chmeruk: Is this blank line really needed here?

pyfar/level/utils.py
```diff
@@ -0,0 +1,108 @@
...
+        >>> pf.plot.time(audio, dB=True, label="Audio content", alpha=0.7)
+        >>> pf.plot.time(fast_weighted, dB=True, label="Fast-weighted level")
+        >>> pf.plot.time(slow_weighted, dB=True, label="Slow-weighted level")
+        >>> plt.ylim(-45, 5)
+        >>> plt.legend()
+        >>> plt.show()
+    """
+    if not isinstance(time_weighting, str):
+        raise TypeError("Time weighting must be a string.")
+    weighting = time_weighting.upper()
```
h-chmeruk: I don’t think there is a test that checks whether the "f" and "s" values for the ‘time_weighting’ parameter are accepted. If they should be accepted, I would add a test; if not, I would remove the line `weighting = time_weighting.upper()`

h-chmeruk (commented):
Thank you for implementing these features! I think everything looks good overall, and I only have two minor comments.

tests/test_level_standard_levels.py
jf-som: No, it must have happened by accident when using auto-imports... I removed it.

pyfar/level/utils.py
jf-som: Thanks, good point. Looking at the code again, the way I check for F and S is weird to begin with. Not sure what happened there, but I cleaned this check up and removed the `.upper()` call as requested.

pyfar/level/standard_levels.py
```diff
@@ -89,3 +90,127 @@ def equivalent_continuous_level(
...
+    The returned array has the same shape as the input in the time domain,
+    containing sound pressure levels in dB relative to the
+    `reference_pressure`.
+
+    For instance, the A-weighted, F-time-weighted level is defined as:
+
+    .. math::
+        L_\text{AF}(t) = 10 \log_{10} \left[ \frac{(1 / \tau_{\text{F}})
+        \int_{-\infty}^{t} p_{\text{A}}^2(\xi) e^{-(t-\xi)/\tau_\text{F}} d\xi}
+        {p_0^2}\right] \text{ dB}
```
f-brinkmann, line 456: Please define the symbols used in the formula below.

pyfar/level/standard_levels.py
```diff
@@ -89,3 +90,127 @@ def equivalent_continuous_level(
...
+    .. plot::
+
+        >>> import pyfar as pf
+        >>> import matplotlib.pyplot as plt
+        >>> import numpy as np
+        >>> noise = pf.signals.noise(
+        >>>     24000, rms=0.001, sampling_rate=48000, seed=0)
+        >>> signal = pf.dsp.pad_zeros(noise, 24000, "beginning")
+        >>> signal = pf.dsp.pad_zeros(signal, 3 * 48000, "end")
+        >>> levels = pf.level.time_weighted_level(signal, "A", "F",
```
f-brinkmann: Mabe don't use freq. weighting to make the time weighting the only difference in the shown levels?
```suggestion
        >>> levels = pf.level.time_weighted_level(signal, "Z", "F",
```

pyfar/level/standard_levels.py
```diff
@@ -89,3 +90,127 @@ def equivalent_continuous_level(
...
+        >>> plt.grid()
+        >>> plt.legend()
+        >>> plt.show()
+    """
+    signal = _check_signal_type(signal)
+    signal = _apply_frequency_weighting(signal, frequency_weighting)
+    signal = _apply_multi_band(signal, num_octave_band_fractions)
+    weighted = time_weighted_pressure(signal, time_weighting)
+    energies = weighted.time**2
+    levels = _energies_to_levels(energies, reference_pressure, replace_zeros)
```
f-brinkmann, line 575: pure beauty and elegance

pyfar/level/utils.py
```diff
@@ -0,0 +1,107 @@
...
+import scipy.signal
+import pyfar as pf
+
+from ._utils import _check_signal_type
+
+def time_weighted_pressure(signal, time_weighting: Literal["F", "S"]):
+    r"""
+    Calculates sound pressure with exponential time weighting.
+    This uses the method defined in IEC 61672-1 [#]_, but returns
+    sound pressure values instead of levels.
```
f-brinkmann: A one-liner without the reference would be better. References in the first line do not work well with Sphinx. E.g.

```suggestion
    Calculate the time-weigted sound pressure.
    
    This uses the method defined in IEC 61672-1 [#]_ to compute the time-weighted sound pressure by applying an exponential decaying  time weighting., but returns sound pressure values instead of levels.
```

pyfar/level/utils.py
```diff
@@ -0,0 +1,107 @@
...
+
+    The standard defines the time weighting F as:
+
+    .. math::
+        L_\text{F}(t) = 10 \log_{10} \left[ \frac{(1 / \tau_{\text{F}})
+        \int_{-\infty}^{t} p^2(\xi) e^{-(t-\xi)/\tau_\text{F}} d\xi}
+        {p_0^2}\right] \text{ dB}
```
f-brinkmann, line 23: As above, I would suggest to introduce the symbols

pyfar/level/utils.py
```diff
@@ -0,0 +1,107 @@
...
+
+    This function works on finite, discrete signals and only calculates the
+    weighted pressure:
+
+    .. math::
+        p_\text{F}[n] = \sqrt{ (1/\tau_F) \sum_{0}^{N-1} p^2(n)
+        e^{-(t-n)/\tau_\text{F}} }
```
f-brinkmann: Above and also in the other function it is (time - constant), here it is (constant - time). Which one is correct?

pyfar/level/utils.py
```diff
@@ -0,0 +1,107 @@
...
+        sign of each sample. This function therefore exists mainly as a
+        helper function for other functions in `pyfar.level` as well as
+        for plotting purposes.
+
+    Parameters
+    ----------
+    signal: Signal
+        The signal object to apply the weighting to.
+
+    time_weighting: ``"F"`` or ``"S"``
```
f-brinkmann, line 55: For discussion: Since we are a scientific toolbox, it might be nice to make it possible to pass a custom time weighting in seconds.

pyfar/level/utils.py
```diff
@@ -0,0 +1,107 @@
...
+    # this is effectively the same as looping over all samples and doing
+    # > weighted[i] = exp_func_decay * weighted[i-1] + energies[i]
+    # but it's faster to use an lfilter call to make use of scipy's speed
+    # and channel handling
+    weighted = scipy.signal.lfilter([1, 0], [1, -exp_func_decay], energies)
+
+    # normalize the integral
+    normalized = weighted / time_constant / signal.sampling_rate
+
+    # turn energy to back to pressure
```
f-brinkmann:
```suggestion
    # convert energy back to pressure
```

tests/test_level_standard_levels.py
```diff
@@ -16,3 +17,34 @@ def test_level_equivalent_continuous_level_known_value():
...
+def test_level_time_weighted_level_replace_zeros_false():
+    """Test that setting replace_zeros to False returns -inf
+    and raises a warning from numpy.
+    """
+    s = pf.Signal(np.zeros(1000), sampling_rate=48000)
+    with pytest.warns(RuntimeWarning, match="divide by zero"):
+        levels_no_replace = pf.level.time_weighted_level(
+            s, "Z", "F", replace_zeros=False)
+    assert levels_no_replace[0][100] == -np.inf
```
f-brinkmann: not even
```suggestion
    assert np.all(levels_no_replace == -np.inf)
```

tests/test_level_standard_levels.py
```diff
@@ -16,3 +17,34 @@ def test_level_equivalent_continuous_level_known_value():
...
+
+
+def test_level_time_weighted_level_replace_zeros_true():
+    s = pf.Signal(np.zeros(1000), sampling_rate=48000)
+    levels_replace = pf.level.time_weighted_level(
+        s, "Z", "F", replace_zeros=True)
+
+    # since there are only zeros in the signal, all values must be the same
+    assert np.all(levels_replace[0] == levels_replace[0][0])
+    assert np.isfinite(levels_replace[0][100])
```
f-brinkmann: Can it be more explicit
```suggestion
    np.all(levels_no_replace == x)
```
with x being 10 log10(eps)? Or is there integration involved resulting in different values?

f-brinkmann (changes requested):
Overall structure looks great. Mainly small comments on docs and tests.

tests/test_level_standard_levels.py
jf-som: Right... not sure what I was (not) thinking there XD

tests/test_level_standard_levels.py
jf-som: Good point. I now check against the actually expected value.

pyfar/level/utils.py
jf-som: Oh right, that was still breaking and I always forget about the first line rule. Fixed it in both functions.

pyfar/level/standard_levels.py
jf-som: 😂 👍

pyfar/level/standard_levels.py
jf-som: Oh yeah, how did I miss that... Added them now.

pyfar/level/utils.py
jf-som: yep, done

pyfar/level/utils.py
jf-som: the (t - xi) in the first formulas is correct. The (t - n) just makes no sense here, because the formula is meant to be discrete, but I used the continuous `t` regardless 🤦‍♂️ 
I overhauled the whole formula, so together with the symbol legend it should hopefully make sense now.
Reworking the formula also made me realize the implementation below can be simplified too, so I did that as well.

jf-som: the (t - xi) in the first formulas is correct. The (t - n) just makes no sense here, because the formula is meant to be discrete, but I used the continuous `t` regardless 🤦‍♂️ 
I overhauled the whole formula, so together with the symbol legend it should hopefully make sense now.
Reworking the formula also made me realize the implementation below can be simplified too, so I did that as well.

pyfar/level/utils.py
jf-som: We could do that, yes. I would leave the `_level` function unchanged with F and S, but for this helper we could do something like change the `time_weighting` parameter to `time_constant`, make `0.125` (the F constant) the default value and name the two values in the docstring. So users would have the flexibility while also not really getting extra work to use the standard values.

f-brinkmann (approved):
thx

h-chmeruk (approved):
Looks good, thanks!

jf-som:
As discussed last meeting, we decided to change `time_weighted_pressure()` to allow different time constants by making it a parameter instead of the time weighting literal directly, as suggested by Fabian above.
This also required changing most of the time weighting tests slightly, although nothing changed about what is tested.
The time_weighted_level function behaves just like before, only the helper utility is more flexible now.

pyfar/level/utils.py
jf-som: We decided to implement this during the last meeting. See my comment below (https://github.com/pyfar/pyfar/pull/954#issuecomment-5761296914)

f-brinkmann (approved):
Thx, `time_weighted_pressure` looks good to me now.

PR #955, opened 2026-07-15
Level/peak

jf-som:
Part of #949.

This PR adds the `peak_level()` function and an oversampling helper utility.

Possible points of discussion:
- default value for `oversampling` could also be `None` (for no oversampling). But since the standard describes the peak level as the highest value at any time, I feel like inter-sample peaks should be considered by default.
- Should I add more detail for the oversampling description? Like how it's more relevant for signals that contain a lot of energy at very high frequencies (near Nyquist) and how the maximum possible error shrinks with increased oversampling. All of this is written in the ITU recommendation, which I already linked in the documentation.


pyfar/level/standard_levels.py
```diff
@@ -89,3 +90,77 @@ def equivalent_continuous_level(
...
+def peak_level(
+        signal,
+        frequency_weighting: Literal["A", "C", "Z"],
+        oversampling: float | None = 4,
+        reference_pressure: float = pf.constants.reference_sound_pressure,
+):
+    """Calculate the frequency-weighted peak sound pressure level.
+
+    The levels are calculated per channel and according to IEC 61672-1 [#]_
+    and optionally applies oversampling to find the true (inter-sample) peaks.
```
h-chmeruk:
```suggestion
    The levels are calculated per channel; according to IEC 61672-1 [#]_
    or optionally applies oversampling to find the true (inter-sample) peaks.
```
or 

```suggestion
    The levels are calculated per channel according to IEC 61672-1 [#]_,
   oversampling can be applied optionally to find the true (inter-sample) peaks.
```
Sorry for this comment, but I don't quite understand the logic of the sentence because of those two “and”s in `and according to IEC 61672-1 [#]_  and optionally applies”`. Could you maybe rephrase the sentence somehow?
(My suggestions are more of a question about what was meant by that, rather than actual suggestions, haha)

pyfar/level/standard_levels.py
```diff
@@ -89,3 +90,77 @@ def equivalent_continuous_level(
...
+    You may refer to ITU-R BS.1770-5 [#]_ Annex 2 for further information and
+    recommendations on how oversampling is used for true-peak detection.
+
+    Parameters
+    ----------
+    signal: Signal
+        The signal object to calculate the levels of
```
h-chmeruk:
```suggestion
        The signal object to calculate the levels of.
```

h-chmeruk (commented):
Thank you for implementing this! I really appreciate the comments in the code. I didn't find much to comment about, except for two small suggestions regarding the docs. 

Regarding your two points/questions:
1. I agree that an oversampling factor of 4 should be the standard. It’s better to get the more accurate result by default.
2. I don’t think it is necessary to include a comprehensive scientific explanation of oversampling in the docs, but perhaps you could add a note encouraging users to read the referenced literature if they are unsure how and when to apply oversampling. You could perhaps simply write something like: 

_"Signals with significant energy near the Nyquist frequency can swing higher between samples than at any sampled point. Oversampling approximates this inter-sample ("true") peak more accurately. Please read the referenced recommendation if you are unsure how to apply oversampling or which factors to choose."_

pyfar/level/standard_levels.py
jf-som: Good catch thanks! Though now I notice that this is also done wrong in the other level functions, oops. 
I corrected it in both functions here and in the other level PR branches as well.

pyfar/level/standard_levels.py
jf-som: Oh yeah, I see your point. This sentence was a bit of patchwork of adding more info later. I should have read over it in the end again xD
I split it into multiple sentences now, and also added a short motivation for the oversampling to the paragraph, similar to what you suggested above.

pyfar/level/standard_levels.py
```diff
@@ -89,3 +90,82 @@ def equivalent_continuous_level(
...
+        reference_pressure: float = pf.constants.reference_sound_pressure,
+):
+    """Calculate the frequency-weighted peak sound pressure level.
+
+    The levels are calculated per channel; and according to IEC 61672-1 [#]_.
+
+    Oversampling can be applied optionally to find inter-sample ("true")
+    peaks. This is useful for digital signals containing significant energy
+    at high frequencies (near the Nyquist frequency), which can cause the
+    true (analog) peak between to samples to exceed the digital peak value.
```
h-chmeruk:
```suggestion
    true (analog) peak between two samples to exceed the digital peak value.
```
Is that a typo, and should there be “two” instead of “to”?

jf-som: 🤦‍♂️So much for reading my own sentences after editing them...
Thanks!

h-chmeruk (approved):

pyfar/level/standard_levels.py
```diff
@@ -89,3 +90,82 @@ def equivalent_continuous_level(
...
+        reference_pressure: float = pf.constants.reference_sound_pressure,
+):
+    """Calculate the frequency-weighted peak sound pressure level.
+
+    The levels are calculated per channel; and according to IEC 61672-1 [#]_.
+
+    Oversampling can be applied optionally to find inter-sample ("true")
+    peaks. This is useful for digital signals containing significant energy
+    at high frequencies (near the Nyquist frequency), which can cause the
+    true (analog) peak between two samples to exceed the digital peak value.
```
f-brinkmann: Can also happen rather far away from Nyquist, maybe a little more neutral and shorter as well:

```suggestion
    Applies optional oversampling to find the "true" inter-sample
    peak. The true peak can be between discrete time samples of digital signals, especially for high frequencies. This can cause the
    true analog peak to exceed the digital peak value if oversampling is not applied.
```

pyfar/level/standard_levels.py
```diff
@@ -89,3 +90,82 @@ def equivalent_continuous_level(
...
+    References
+    ----------
+    .. [#] International Electrotechnical Commission,
+        "IEC 61672-1:2013 - Electroacoustics - Sound level meters - Part 1:
+        Specifications", IEC, 2013.
+
+    .. [#] International Telecommunication Union,
+        Recommendation ITU-R BS.1770-5 (11/2023): "Algorithms to measure
+        audio programme loudness and true-peak audio level".
+        https://www.itu.int/dms_pubrec/itu-r/rec/bs/R-REC-BS.1770-5-202311-I!!PDF-E.pdf
```
f-brinkmann: Maybe more elegant to use the title for linking
```suggestion
    .. [#] International Telecommunication Union,
        Recommendation ITU-R BS.1770-5 (11/2023): "(Algorithms to measure
        audio programme loudness and true-peak audio level)[https://www.itu.int/dms_pubrec/itu-r/rec/bs/R-REC-BS.1770-5-202311-I!!PDF-E.pdf]."
```

tests/test_level_common_parameters.py
```diff
@@ -150,3 +153,55 @@ def test_level_common_reference_pressure_errors(function):
...
+
+### oversampling parameter tests ###
+
+FUNCTION_WRAPPERS_OVERSAMPLING = [
+    lambda s, o: pf.level.peak_level(s, "Z", o),
+]
+
+
+@pytest.mark.parametrize("oversampling", [None, 2, 3.5, 8])
+@pytest.mark.parametrize("function", FUNCTION_WRAPPERS_OVERSAMPLING)
```
f-brinkmann, line 173: Are you planning to extend this with other functions in later Pulls? If not I would suggest to explicitly call peak_level, to make the code more readable and even shorter (less lines in total)

tests/test_level_standard_levels.py
```diff
@@ -9,10 +9,20 @@
 import pyfar as pf
 import numpy as np
 
+# 1 Pa in dB SPL
+ONE_PA = 20 * np.log10(1 / pf.constants.reference_sound_pressure)
+SINE_PAPR = 10 * np.log10(2)  # peak-to-average power ratio of sine signals
+
```
f-brinkmann, line 16: 🍺

tests/test_level_standard_levels.py
```diff
@@ -9,10 +9,20 @@
...
-    # 94 dB is 1 Pa, -3.01 dB is the crest factor of sine signals
-    assert np.isclose(levels, 94 - 3.01, atol=0.1)
+    assert np.isclose(levels, ONE_PA - SINE_PAPR, atol=0.001)
+
+
+def test_level_peak_level_known_value():
+    s = pf.signals.impulse(1000)
+    levels, times = pf.level.peak_level(s, "Z", None, 1)
+    assert levels[0] == 0   # peak is 1, so 0 dbFS
+    assert times[0] == 0
```
f-brinkmann: I would extend this to use oversampling (at least `None` and `4`): If you use a delayed pulse `s = pf.signals.impulse(1000, 512)` the maximum should be 1 (or extremely close to it) because you are sampling the maximum of the underlying sinc function. Moreover, it will always appear at exactly at `512/s.sampling_rate` regardless of the oversampling,

f-brinkmann (changes requested):
Looks good - only suggested improvements on docs and tests

pyfar/level/standard_levels.py
jf-som: Thanks. I changed the wording of the second sentence to sound more scientific, but otherwise did as suggested.

pyfar/level/standard_levels.py
jf-som: I like the idea. The syntax is different for Sphinx, but I made it work.

tests/test_level_common_parameters.py
jf-som: Yes, the `maximimum_time_weighted_level()` function will be added later. I agree that it looks convoluted now, but that's just because I already have a branch with all the level functions locally and basically subtracted the other function for this one 😅 It will make more sense and be shorter once I get around to `maximimum_time_weighted_level` in the next PR.

tests/test_level_standard_levels.py
jf-som: Good point, this should be checked. I modified the test similar to your suggestion

f-brinkmann (approved):

PR #963, opened 2026-08-17
[ENH] Option for exclusive intervals for dsp.time_crop

### LINKED ISSUE #870: [ENH] Add exclusive intervals for the time crop function
This was brought forward as an additional feature in #849.

Can be picked up after #849 is finalized.

Related to #751

jf-som:
Adds `exclude_end` parameter to time crop to achieve a behavior in line with Python's `range` and numpy's slicing logic.
Having this parameter at all also makes it more clear that by default, the interval is *inclusive*.

Closes #870


pyfar/dsp/dsp.py
```diff
@@ -591,6 +599,11 @@ def time_crop(signal, interval: Union[list[float], tuple[float, float],
         Values in seconds are rounded to the nearest sample within the
         specified `interval` range. The default is ``'samples'``.
 
+    exclude_end : bool, optional
+        If ``True``, the end of the interval is exclusive, i.e., the sample
+        at the end of the interval is not included in the cropped signal.
+        The default is ``False``.
```
f-brinkmann, line 605: You are not explicitly explaining default behaviour. I would additionally prefer a default of True:

```suggestion
    incude_end : bool, optional
        The default ``True``, include the end of the interval, i.e., the sample
        at the end of the interval is included in the cropped signal. This is not
        the case for ``False``.
```

f-brinkmann (changes requested):
Motivation and code looks good to me. I would suggest a different name, but would wait for more opionons before changing anything.

pyfar/dsp/dsp.py
jf-som: True, it's more intuitive to read it with your version. Mine is kind of like a double negation wording.
I might also add a half sentence that include_end=False matches `range()`/`slice` behavior. 
But as you said, let's wait for other opinions first.

PR #965, opened 2026-08-25
Draft implementation of the `_PyfarArithmetics` class and adjustment of the `_Audio` and `Coordinates` classes

h-chmeruk:
### Introduction:
This PR sketches a possible implementation of the `_PyfarArithmetics` class. It is not (yet) meant to be merged, it only shows what the change would look like and what it would mean if we implement `_PyfarArithmetics` this way.

Coming up with a universal solution for implementing arithmetic operations was not easy without looking deeply into the current implementation in the existing classes. I therefore built this version of `_PyfarArithmetics` by commenting out parts of the code in `_Audio` and `Coordinates` and working out how to provide the same functionality in `_PyfarArithmetics`.

### The proposal:
The functions for arithmetic operations are implemented as instance methods in the abstract class `_PyfarArithmetics`. Each of the functions basically does three things:
```
result = self.copy
a, b, result = _extract_data()
result._data = operation(a, b)
```
Each child class extracts, broadcasts and reshapes its data as it needs to in the private method `_extract_data()`, which returns the data to the instance method that performs the arithmetic operation.

`    def _extract_data(self, other, mode, operation: str):`

- `self` — the class instance itself
- `other` — the other operand
- `mode` — the mode in which the operation should be performed: freq
  (frequency domain), time (time domain), or None (for Coordinates calculation)
- `operation` — which arithmetic operation is to be performed, so that the data can
  be checked and reshaped in the right way
  
### The purpose of this PR:
This PR is a large suggestion of what `_PyfarArithmetics` could look like and how the existing files would need to be adjusted to it. I would like to discuss the design questions here; once we have agreed on a specific implementation, I will move it into the actual `_PyfarArithmetics` pull request (PR #845) (without the changes to the other classes and test functions, of course).

### My personal concerns regarding the proposal:
The main disadvantage is poor backwards compatibility. The module-level functions `pf.add(data[, domain])`, `pf.subtract(data[, domain])`, `pf.multiply(data[, domain])`, `pf.divide(data[, domain])` and `pf.power(data[, domain])` are replaced by the instance methods `add(self, *others, mode=None)`, `subtract(self, *others,
mode=None)`, `multiply(self, *others, mode=None)`, `divide(self, *others,
mode=None)` and `power(self, *others, mode=None)`.

Known issues / reasons the tests fail:

1. `__div__` and `__rdiv__` are not implemented. They are outdated and were replaced by `__truediv__` and `__rtruediv__`in Python 3, so I left them out. This is why `test_coordinates_arithmetic/test__div__()` and `test_coordinates_arithmetic/test__rdiv__()` fail.

2. `test_audio_signal_arithmetic/test_add_arrays()` fails on:
```
   x = np.arange(2 * 3 * 4).reshape((2, 3, 4))
   y = np.arange(2 * 3 * 4).reshape((2, 3, 4))
   z = x.add(y, mode="freq")
```
x and y are plain `numpy.ndarray`, so neither is an instance of a class derived from `_PyfarArithmetics`. Since add(), subtract(), multiply() etc. are now instance methods, ndarray has no add attribute.

3. The `__rpow__` implementation is rather ugly. There is no clean way to swap the operand order of `self.power(other)`, so the exponentiation is implemented directly inside `__rpow__`.

### DIFF
--- a/pyfar/classes/_PyfarArithmetics.py
+++ b/pyfar/classes/_PyfarArithmetics.py
@@ -1,0 +1,77 @@
+"""
+The following documents the abstract class for pyfar classes,
+which defines arithmetic operations.
+"""
+
+from abc import ABC, abstractmethod
+
+class _PyfarArithmetics(ABC):
+    """
+    Internal abstract base class for pyfar classes.
+
+    This class defines the basic arithmetic operations (addition, subtraction,
+    multiplication, division and exponentiation) and their right-hand
+    alternatives, which must be implemented by other classes.
+    """
+
+    @abstractmethod
+    def __add__(self, data):
+        """Addition of two objects (`self + data`)."""
+        pass
+
+    @abstractmethod
+    def __radd__(self, data):
+        """Reflected addition of two objects (`data + self`)."""
+        pass
+
+    @abstractmethod
+    def __sub__(self, data):
+        """Subtraction of two objects (`self - data`)."""
+        pass
+
+    @abstractmethod
+    def __rsub__(self, data):
+        """Reflected subtraction of two objects (`data - self`)."""
+        pass
+
+    @abstractmethod
+    def __mul__(self, data):
+        """Multiplication of two objects (`self * data`)."""
+        pass
+
+    @abstractmethod
+    def __rmul__(self, data):
+        """Reflected multiplication of two objects (`data * self`)."""
+        pass
+
+    @abstractmethod
+    def __truediv__(self, data):
+        """Division of two objects (`self / data`)."""
+        pass
+
+    @abstractmethod
+    def __rtruediv__(self, data):
+        """Reflected division of two objects (`data / self`)."""
+        pass
+
+    @abstractmethod
+    def __pow__(self, data):
+        """Exponentiation.
+        Raise the object to the power of `data` (`self ** data`).
+        """
+        pass
+
+    @abstractmethod
+    def __rpow__(self, data):
+        """
+        Reflected exponentiation.
+        Raise `data` to the power of the object (`data ** self`).
+        """
+        pass
+
+    @abstractmethod
+    def _assert_match_for_arithmetics(self, data):
+        """
+        Check that two objects are compatible for arithmetic
+        operations.
+        """
--- a/pyfar/classes/_PyfarArithmetics.py
+++ b/pyfar/classes/_PyfarArithmetics.py
@@ -3,9 +3,11 @@ The following documents the abstract class for pyfar classes,
 which defines arithmetic operations.
 """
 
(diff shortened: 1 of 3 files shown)


jf-som:
Thanks for taking on this rather large overhaul. I think the new instance methods make much more sense than the old logic and are more in line with what I know from other projects.

Since you say this isn't intended to be merged, that means you don't need detailed reviews at this point in time?
In that case, I just have some comments and replies to the points you raised. Otherwise let me know if you'd like a detailed review.

- If I understand correctly, these changes would solve #813, right? Since with this implementation, add and multiply will just operate in the domain that at least one of the operands has. This is definitely a plus, so perhaps we can link the issue here?
- Is the backwards compatibility really a problem? As I see it, we could still keep the old functions and just have them forward to the new ones or similar, and do the same deprecation warning as usual. We could mark them as deprecated until 1.0, where major changes will come anyway. Or am I missing something?
- that `test_add_arrays` test makes no sense with the new logic. If we decide to keep the old `pf.add` and friends for a while, the test makes sense with the old function. But to be honest I don't think it makes sense for our functions to work on only non-pyfar objects in the first place, so I wouldn't worry about this test.

h-chmeruk:
@jf-som, thanks for your review! I actually forgot to include a more detailed explanation of why this PR exists and what my goal was, hahah. I'll do that shortly.

Regarding your questions/comments:
1. I’ve actually never read the Issue #813 before, but I think a different implementation of the dunder methods might solve the problem.  I’ve noticed that in most use cases, it was expected that all operations on `Signal` objects are performed in the freq domain by default. So I’ve applied this to the dunder methods for `Signal` objects in my implementation as well. But it is easy to remove this again and do the multiplication/division with non-pyfar objects in the time domain, for example like this:
```
    def __mul__(self, data):
        """Return self * data."""
        if isinstance(data, _PyfarArithmetics):
            return self.multiply(data, mode="freq")
        else:
            return self.multiply(data, mode="time")
```

2,3. Yeah, I mean, we could of course keep the old functions, but if we were to remove them completely, we would have different functionality than before, as with the new implementation (with instance methods) we couldn’t simply add two non-pyfar objects. Perhaps "backwards compatibility" is not quite the right term, but if some users have actually been using the `pf.add()`  and friends functions directly (not as dunder methods), they will need to adapt their code. And if, for any reason, they have used these functions for calculations involving only non-pyfar objects, they will no longer have this functionality in the newer versions with the `_PyfarArithmetics` implementation. 

But in general, I agree... I don’t think it’s necessary for our arithmetic functions to work on only non-pyfar objects such as np.ndarray or int. So the question is simply whether all of us, the pyfar developers, agree and think that this is a better implementation.

Issue #516, opened 2023-10-31
Tests are redundant


artpelling:
I think a lot of the tests (at least regarding the `_Audio` related classes) could be simplified by using `mark.parametrize` with different input signals. A lot of tests are identical across `test_audio_signal.py`, `test_audio_time_data.py` and `test_audio_frequency_data.py`. I suspect that one can find more of these instances.

mberz:
I think a lot of this was implemented before we discovered parametrisation is a thing :)

PR #966, opened 2026-09-02
Improve test functions in `test_audio_signal_arithmetic.py` with parameterisation

h-chmeruk:

Part of the solution to Issue  #516
I agree that many of the old test functions are very repetitive and need to be updated. I would create an umbrella (integration) PR and collect all the changes to the various test files there. That is why I would like to merge into the `improve_existing_tests_audio` branch and not into `develop`.

### Open discussion:
I haven’t really looked into the test coverage or whether the tests are useful. Also, I haven’t added any docstring descriptions for the test functions where there weren’t any before.
Please let me know if you think this should be included in this PR. 

#### Merged test functions:
* Merged `test_add_two_signals_time`, `test_add_two_signals_time_complex`, `test_add_two_signal_time_real_and_complex`, `test_add_two_signals_freq`, `test_add_two_signals_freq_complex`, and `test_add_two_signals_freq_real_and_complex` into
  **`test_add_two_signals_time_and_freq`** via `pytest.mark.parametrize`.

* Merged `test_add_signal_and_number`, `test_add_signal_and_complex_number`,
  `test_add_number_and_signal`, `test_add_number_and_complex_signal`, and
  `test_add_number_and_complex_signal_freq` into **`test_add_signal_and_number`**
  via `pytest.mark.parametrize`.
* Merged the checks of `test_add_time_data_and_number`, `test_add_time_data_and_time_data`, `test_add_frequency_data_and_number`, and `test_add_frequency_data_and_frequency_data` into
  **`test_add_time_data_frequency_data`** via `pytest.mark.parametrize`.

* Merged the error-path checks of `test_add_time_data_and_number_wrong_domain`, `test_add_time_data_and_number_wrong_times`, `test_add_frequency_data_and_number` (its inline error check), `test_add_frequency_data_and_number_wrong_domain`, and `test_add_frequency_data_and_number_wrong_frequencies` into
  **`test_add_time_data_frequency_data_errors`** via `pytest.mark.parametrize`.

* Merged `test_add_array_and_signal` and `test_add_signal_and_array` into a single **`test_add_array_and_signal`**, parametrized over array shape and operand order (`swap`).

* Merged `test_multiplication`, `test_complex_multiplication`, `test_complex_real_multiplication`, and `test_complex_real_multiplication_freq` into two parametrized tests: 
  **`test_real_multiplication`** (real × real case) and **`test_complex_multiplication`** (the three combinations involving at least one complex operand). 
These needed to stay separate functions rather than one merged test, because the real × real case returns a frequency-domain array with a different number of bins.

* Merged `test_division` and `test_complex_division` into **`test_division`**
  via `pytest.mark.parametrize`.

* Merged `test_overloaded_operators_signal`, `test_overloaded_operators_time_data`, and `test_overloaded_operators_frequency_data` into
  **`test_overloaded_operators_signal_time_freq_data`**, 
replacing three manual for-loops with a full `pytest.mark.parametrize` cross product.

* Merged `test_matrix_multiplication_TimeData` and `test_matrix_multiplication_FrequencyData` into
  **`test_matrix_multiplication_TimeData_FrequencyData`** via `pytest.mark.parametrize`.

* Merged `test_matrix_multiplication_TimeData_times_array` and `test_matrix_multiplication_FrequencyData_times_array` into
  **`test_matrix_multiplication_TimeData_FrequencyData_times_array`** via `pytest.mark.parametrize`.

#### Test functions split into several:

* Split `test_assert_match_for_arithmetic`  into four parametrized tests: **`test_assert_match_for_arithmetic`,** **`test_assert_match_for_arithmetic_output`** (output-value checks), **`test_assert_match_for_arithmetic_complex_flag`** (complex-flag propagation), and **`test_assert_match_for_arithmetic_complex_flag_errors`** (error handling).

* Split `test_array_broadcasting_errors` (one function checking two unrelated error conditions sequentially) into
**`test_array_broadcasting_dimension_error`** and **`test_array_broadcasting_shape_error`.**

#### Test functions only parametrized (no merge or split):

These kept their names and scope but were converted from manual loops or
sequential repeated checks into `pytest.mark.parametrize`, so failures are
now reported per-case instead of only for the whole function:

* `test_overloaded_operators_array_and_signal`
* `test_get_arithmetic_data_with_signal` (also removed the leftover debug `print` statement)
* `test_matrix_multiplication_operator`

#### Bugs / gaps found and fixed:

* `test_add_two_signals_freq_real_and_complex` built a real `y`  and a complex `x` signal but then called `pf.add((x, x), 'freq')`, so it never actually used the real signal `y`, so the real+complex frequency-domain case was untested. Fixed as part of the merge into `test_add_two_signals_time_and_freq`.

* Added frequency-domain coverage for real × real multiplication (`test_real_multiplication`), which was previously untested: the original `test_multiplication` only checked the time domain.


tests/test_audio_signal_arithmetic.py
```diff
@@ -5,114 +5,36 @@
...
-    # generate test signal
-    x = Signal([1, 0, 0], 44100, is_complex=True)
 
-    # time domain
-    y = pf.add((x, x), 'time')
+@pytest.mark.parametrize(("domain", "x_complex", "y_complex", "desired"), [
+    ('time', False, False, np.atleast_2d([2, 0, 0])),
+    ('time', True, True, np.atleast_2d([2 + 0j, 0, 0])),
+    ('time', True, False, np.atleast_2d([2 + 0j, 0 + 0j, 0 + 0j])),
+    ('freq', False, False, np.atleast_2d([2, 2])),
```
hoyer-a: Technically the returned spectrum is always complex, but assert_allclose doesnt care as long as the values are close.

```suggestion
    ('freq', False, False, np.atleast_2d([2+0j, 2+0j])),
```

tests/test_audio_signal_arithmetic.py
```diff
@@ -5,114 +5,36 @@
...
-    # frequency domain
-    z = pf.add((x, x), 'freq')
-    # check if old signal did not change
-    npt.assert_allclose(x.time, np.atleast_2d([1 + 0j, 0 + 0j, 0 + 0j]),
-                        atol=1e-15)
-    npt.assert_allclose(y.time, np.atleast_2d([1, 0, 0]), atol=1e-15)
+    if domain == 'time':
+        npt.assert_allclose(z.time, desired, atol=1e-15)
+    else:
+        npt.assert_allclose(z.freq, desired, atol=1e-15)
```
f-brinkmann, line 33: The old tests partly also tested `z.domain == domain` and `z.complex == x_complex or y_complex`. Can you please add this? I think it can be added outside the if/else case.

tests/test_audio_signal_arithmetic.py
```diff
@@ -130,212 +52,96 @@ def test_add_three_signals():
...
+@pytest.mark.parametrize(("y", "swap", "domain", "is_complex_x",
+                           "is_complex_y", "desired"), [
+    (1, False, 'time', False, False, np.atleast_2d([2, 1, 1])),
+    (1 + 1j, False, 'time', False, True,
+        np.atleast_2d([2 + 1j, 1 + 1j, 1 + 1j])),
+    (1, True, 'time', False, False, np.atleast_2d([2, 1, 1])),
+    (1, True, 'time', True, False, np.atleast_2d([2 + 0j, 1 + 0j, 1 + 0j])),
+    (1, True, 'freq', True, False, np.atleast_2d([2 + 0j, 2 + 0j, 2 + 0j])),
+])
+def test_add_signal_and_number(y, swap, domain, is_complex_x, is_complex_y,
```
f-brinkmann, line 65: Did we loose `def test_add_time_data_and_number`? If yes, please reintroduce it.

tests/test_audio_signal_arithmetic.py
```diff
@@ -130,212 +52,96 @@ def test_add_three_signals():
...
+        'time', np.atleast_2d([2, 0, 0]), np.atleast_1d([0, .1, .5])),
+    (FrequencyData([1, 0, 0], [0, .1, .5]), 1, 'freq',
+        np.atleast_2d([2, 1, 1]), np.atleast_1d([0, .1, .5])),
+    (FrequencyData([1, 0, 0], [0, .1, .5]),
+        FrequencyData([1, 0, 0], [0, .1, .5]), 'freq',
+        np.atleast_2d([2, 0, 0]), np.atleast_1d([0, .1, .5])),
+
+])
+def test_add_time_data_frequency_data(x, y, domain, desired_data,
+                                       desired_instances):
```
f-brinkmann, line 99: Too much indention, I think.

tests/test_audio_signal_arithmetic.py
```diff
@@ -459,273 +230,124 @@ def test_power():
...
+
+
+@pytest.mark.parametrize(("data", "domain", "match"), [
+    ("s", 'time',
+        "Input argument 'data' must be a tuple."),
+    (("s", ['str', 'ing']), 'time',
+        "Input must be of type Signal, int, float, or complex"),
+    (("s", "s1"), 'time', 'The sampling rates do not match'),
+    (("s", "s2"), 'time', 'The number of samples does not match')])
+def test_assert_match_for_arithmetic_complex_flag_errors(data, domain,match):
```
f-brinkmann, line 321: Maybe rename, this does not seem to involve the complex flag:
```suggestion
def test_assert_match_for_arithmetic_errors(data, domain,match):
```

tests/test_audio_signal_arithmetic.py
```diff
@@ -459,273 +230,124 @@ def test_power():
...
-    # position 1
+SIGNALS = {
+    "s":  lambda: Signal([1, 2, 3, 4], 44100),
+    "s1": lambda: Signal([1, 2, 3, 4], 48000),
+    "s2": lambda: Signal([1, 2, 3], 44100),
+    "s4": lambda: Signal([1, 2, 3, 4], 44100, fft_norm="rms"),
+    "s5": lambda: Signal([1+1j, 2+2j, 3+3j, 4+4j], 48000, is_complex=True),
+    "s6": lambda: FrequencyData(
+        [1+1j, 2+2j, 3+3j, 4+4j], [10, 200, 1000, 20000]),
+}
```
f-brinkmann, line 293: I would suggest to move this to the top of the test file and directly define the signals using more verbose names, e.g.:

`SIGNAL1 = Signal([1, 2, 3, 4], 44100)`

This would avoid the uses `resolved` and simplify tests.

f-brinkmann (changes requested):
Thanks for the effort, its nice to loose so many lines. It was not that simple to see if everything that was tested before is still tested. Could you please additionaly check the test coverage in this branch in comparison to develop to make sure that we do not lose tests?

This can be done with `pytest --cov=. --cov-report=html`

tests/test_audio_signal_arithmetic.py
jf-som: That seems to be already there...?

jf-som (changes requested):
Thanks for starting this chore!

In addition to what Fabian mentioned, I suggest going over the formatting again for consistency. In particular:
- check that there is always 2 blank lines between tests, not 3 or 1
- check that there are no blank lines between the `parametrize` decorator block and the `def test...` line
- check that there is no blank line at the start of a function. There should be either a comment or code
- check that the indentation for parametrize blocks is consistent. Sometimes you have a single tab, sometimes its to the parenthesis above, sometimes completely different. The usual thing I know would be to use single/double tab in new lines, but it should be the same across the file in my opinion.

tests/test_audio_signal_arithmetic.py
h-chmeruk: I have combined the tests `test_add_time_data_and_number`, test_add_time_data_and_time_data, test_add_frequency_data_and_number and test_add_frequency_data_and_frequency_data into the new test function `test_add_time_data_frequency_data`. So, if I’ve done everything correctly, the checks from `test_add_time_data_and_number` should now be included in the new function. 
But please let me know if I have missed something and have not provided a check for certain cases!

tests/test_audio_signal_arithmetic.py
h-chmeruk: Yeah, it’s already there, in lines 36–37! But I’ve also noticed that it’s difficult to review this PR because the diffs look so strange, and there can be a huge red block of removed code within a single function, so it’s not always easy to see where a test function begins and ends...

h-chmeruk:
@f-brinkmann, thank you very much for your review!
I will rebase the branch shortly. I also wanted to compare the coverage between the new and old implementations, but I wasn’t sure how to do that. Could you perhaps briefly explain whether I should also add the HTML with the coverage results to the PR, or how should I make the coverage results visible to everyone?

h-chmeruk:
@jf-som, thank you very much for your review!
These are all very valid points, I will take care of it

h-chmeruk:
Test coverage in this branch is 96%, in the develop branch it is 97%. However, the 1% difference is not due to the file `tests\test_audio_signal_arithmetic.py`, but to the file `tests\test_plot.py`.
In the develop branch, 3 lines/cases were missing in `tests\test_plot.py`, in my new branch, 4 lines/cases are missing.

Here are also the screenshots of the coverage for the file in question. The first image is the refactored version from this branch, the second image is the old version from the develop branch.
image
image

tests/test_audio_signal_arithmetic.py
jf-som: @h-chmeruk That's true. The file is also really long, which doesn't help for the overview. I don't think you have to split this PR into smaller PRs like I do for the level module, that would be overkill here. But as a suggestion, it would help if you gave every reworked test function (or groups of test functions that belong together in a way) its own commit rather than pushing everything in one commit. That way, we can view the changes commit by commit, seeing only how each new function replaces a few old ones :)
I think it's fine for this branch that we already reviewed, but it would probably help us more easily review the other test files, like the one you pushed earlier today. I'll continue this comment there ^^

tests/test_audio_signal_arithmetic.py
```diff
@@ -735,40 +334,38 @@ def test_get_arithmetic_data_with_array():
...
-            ['freq', 'rms'],
-            ['time', 'power'],
-            ['freq', 'power'],
-            ['time', 'psd'],
-            ['freq', 'psd']]
-
+@pytest.mark.parametrize("domain", ["time", "freq"])
+# all possible combinations of `domain`, `signal_type`, and `fft_norm`
+@pytest.mark.parametrize("meta", [
+        ['time', 'none'],
```
jf-som: Small nitpick: Everywhere else you have a single tab indent, here it's two. Not really important, but since this file is getting a rework anyway, might as well unify :)

tests/test_audio_signal_arithmetic.py
```diff
@@ -843,16 +442,15 @@ def test_matrix_multiplication_time_domain():
     npt.assert_allclose(z.time, desired, atol=1e-15)
 
 
-def test_matrix_multiplication_operator():
+@pytest.mark.parametrize(('swap', 'desired'),
+    [(False, np.array([[22, 28], [49, 64]])[..., None] * np.ones((2, 2, 6))),
+     (True, np.array([[9, 12, 15], [19, 26, 33], [29, 40, 51]])[..., None]*
+                                                        np.ones((3, 3, 6)))])
```
jf-som:
```suggestion
     np.ones((3, 3, 6)))])
```
for consistency

tests/test_audio_signal_arithmetic.py
```diff
@@ -843,16 +442,15 @@ def test_matrix_multiplication_time_domain():
     npt.assert_allclose(z.time, desired, atol=1e-15)
 
 
-def test_matrix_multiplication_operator():
+@pytest.mark.parametrize(('swap', 'desired'),
+    [(False, np.array([[22, 28], [49, 64]])[..., None] * np.ones((2, 2, 6))),
+     (True, np.array([[9, 12, 15], [19, 26, 33], [29, 40, 51]])[..., None]*
```
jf-som:
```suggestion
    (True, np.array([[9, 12, 15], [19, 26, 33], [29, 40, 51]])[..., None] *
```

tests/test_audio_signal_arithmetic.py
```diff
@@ -843,16 +442,15 @@ def test_matrix_multiplication_time_domain():
     npt.assert_allclose(z.time, desired, atol=1e-15)
 
 
-def test_matrix_multiplication_operator():
+@pytest.mark.parametrize(('swap', 'desired'),
```
jf-som:
```suggestion
@pytest.mark.parametrize(('swap', 'desired'), [
```

tests/test_audio_signal_arithmetic.py
```diff
@@ -843,16 +442,15 @@ def test_matrix_multiplication_time_domain():
...
+    [(False, np.array([[22, 28], [49, 64]])[..., None] * np.ones((2, 2, 6))),
```
jf-som:
```suggestion
    (False, np.array([[22, 28], [49, 64]])[..., None] * np.ones((2, 2, 6))),
```

jf-som (changes requested):
Thanks for the formatting changes.
I found two formatting inconsistencies that you might have missed, otherwise I think this looks good.
I haven't really checked the test semantics or coverage, but I trust you and Fabian will take care of that :D

